PR #27, opened 2017-05-06
#22 - CSS overflow in event public page.

### LINKED ISSUE #22: Issue with CSS (overflow) in event public page
screen shot 2017-05-01 at 12 36 31 am

**Current behavior:**
Page at http://nextgen.eventyay.com/e/42cf5943 has issues with CSS where both the left (navigation) and right (tweets) segments overflow (no scroll) on a 1440x900p resolution.

**Expected behavior:**
Limit width of center section such as both left (navigation) and right (tweets) section are visible without overflow.

sumedh123:
Initial behaviour : 

initial

After Fix : 

afterfix

Fixes #22 for large and medium screen sizes. However for some small ones, I am working and will soon add commits.


app/styles/partials/footer.scss
```diff
@@ -2,6 +2,9 @@
   display: flex;
   min-height: 100vh;
   flex-direction: column;
+  .main-container {
```
niranjan94: @sumedh123 this isn't the right way to do this. This would affect all the other pages. (Pages which do not have a side-rail and hence don't have an overflow issue). 

Also, we want this fix only for smaller screen sizes.

So, 
- This styles should be applied only to pages that have a left rail
- The styles should be applied depending on the screenwidth. If the screen is wide enough, don't disturb semantic's default width/sizing.

sumedh123: "We want this fix only for smaller screen sizes." But larger screen is having the problem too. Okay I will take care of all the screen sizes.

niranjan94: @sumedh123 Not all screen sizes have this issue. So figure out which all have the issues and fix accordingly. 

And when using media queries to properly support different device width's, ensure you follow Semantic-UI's device widths (https://semantic-ui.com/elements/container.html)

screen shot 2017-05-07 at 12 29 36 am

sumedh123: @niranjan94 For small screen sizes, what is the expected position of rails? Because the middle portion containing information occupies all the space and placing the left and right rail along with the middle part won't make it look good.

niranjan94: For mobile screens, where the middle containers takes up the full width, we want the rail menu to be hidden and instead a different menu at the top of the content. 
This can be dealt with as separate issue. 

In this issue, you can just fix for the other screen sizes where we still can have rails.

sumedh123:
@niranjan94 I have added the media queries so that it looks good on small, medium, large sizes( > 1200px ).
Screenshots : 
Large : 
large
Medium : 

medium
Small : 
small

app/styles/partials/footer.scss
```diff
@@ -33,3 +33,68 @@ footer {
     right: 0 !important;
   }
 }
+
+
+@media screen and (min-width : 1200px) {
+    #main-container {
```
niranjan94: @sumedh123 we need these styles applied **only** for pages with a side rail. We don't want this on all pages.

app/styles/partials/footer.scss
```diff
@@ -33,3 +33,68 @@ footer {
     right: 0 !important;
   }
 }
+
```
niranjan94: And please don't put these styles in footer.scss. They have no relationship to that.

app/styles/partials/footer.scss
sumedh123: @niranjan94 What are those pages with a side rails? I am not able to find any other pages for now who are having rails.

app/styles/partials/footer.scss
sumedh123: @niranjan94 Sure. I 'll move them in override.scss

app/styles/partials/footer.scss
niranjan94: @sumedh123 the code needs to work on any page with side rails ... Have you css selectors like that

app/styles/partials/footer.scss
niranjan94: Create a new partial called a `rails.scss` and move it there.

app/styles/partials/footer.scss
sumedh123: @niranjan94 We can use SASS placeholder to put it up in a place and then just adding class to the parent of the rails would balance the rails and the middle part for any pages.

sumedh123: @niranjan94 Opinions?

niranjan94: @sumedh123 can you show an implementation of your idea ?

sumedh123: @niranjan94 I have overrided the classes of semantic ui. So any rails with those classes will be aligned as is present in above pictures. Adding placeholders will improve the quality of code. That's it. I am working on it. Still, I guess without placeholders the code will also work.

sumedh123:
@niranjan94 Can you deploy the newest commits to the site and check the deployment? Because some other contributor has changed the markup so I had to change the selectors. Also what else is needed?

niranjan94:
@sumedh123 there are conflicts. Please resolve them first.

sumedh123:
@niranjan94 Sorry for late reply. Lost my internet connectivity today. I have resolved the conflicts. Please review.

app/styles/partials/rails.scss
```diff
@@ -0,0 +1,57 @@
+@media screen and (min-width : 1200px) {
+    .ui .rail.left {
+      left : auto;
+      right : 80%;
+      padding : 0 2rem 0 0;
+      margin : 0 2rem 0 0;
+    }
+    .event .description {
```
niranjan94: Like I said before, please ensure this works on all pages with a rail. Use css [Adjacent sibling selectors](https://developer.mozilla.org/en/docs/Web/CSS/Adjacent_sibling_selectors) to make this work with all pages that have a rail.

niranjan94:
And also, please squash all your commits into one.

sumedh123:
@niranjan94 Please review.

app/styles/partials/rails.scss
```diff
@@ -0,0 +1,57 @@
+@media screen and (min-width : 1200px) {
+    .ui .rail.left {
+      left : auto;
+      right : 80%;
+      padding : 0 2rem 0 0;
+      margin : 0 2rem 0 0;
+    }
+    .rail.left + .event > .description {
```
niranjan94: Again ... You're doing it only for the public event page. Don't use `.event > .description`. Keep it generic by referring to the div next to the rail div .

sumedh123:
@niranjan94 Please review.

app/styles/partials/rails.scss
```diff
@@ -0,0 +1,57 @@
+@media screen and (min-width : 1200px) {
```
niranjan94: Keep a max-width of `1600px` here. because, after `1600px` the normal width itself is alright

app/styles/partials/rails.scss
```diff
@@ -0,0 +1,57 @@
...
+    .ui .rail.left {
+      left : auto;
+      right : 80%;
+      padding : 0 2rem 0 0;
+      margin : 0 2rem 0 0;
+    }
+    .rail.left + div {
+      width : 60%;
```
niranjan94: And this can go up a little bit. There's more space that we can use.

PR #32, opened 2017-05-09
Adds featured speakers layout to public events

### LINKED ISSUE #12: Add Featured speakers to the public event page (UI Only)
screen shot 2017-04-23 at 10 07 59 pm

Add the features speaker UI to the public event page as shown above.

**What is expected ?**
- Featured speakers with each speaker as a [component](https://guides.emberjs.com/v2.12.0/getting-started/core-concepts/#toc_components) added to the public event page.
- UI Only for now. Dummy data to be used. 

**References:**
- https://guides.emberjs.com/v2.12.0/components/defining-a-component/
- https://semantic-ui.com/views/card.html
- https://semantic-ui.com/collections/grid.html

geekyd:
#### Checklist

#### Short description:
Adds the layout for featured speakers to the public event page.

- Adds featured speakers layout for event page.

Fixes #12 
@niranjan94 Please review this.

### DIFF
--- a/app/components/public/speaker-item.js
+++ b/app/components/public/speaker-item.js
@@ -1,7 +1,12 @@
 import Ember from 'ember';
 
-const { Component } = Ember;
+const { Component, computed } = Ember;
 
 export default Component.extend({
-  classNames: ['column', 'four', 'wide']
+  classNameBindings: ['style'],
+
+  style: computed('isFeatured', function() {
+    const allClasses = this.get('isFeatured') ? 'column eight wide' : 'column four wide';
+    return allClasses;
+  })
 });
--- a/app/controllers/public/index.js
+++ b/app/controllers/public/index.js
@@ -4,11 +4,15 @@ const { Controller, computed, String } = Ember;
 
 export default Controller.extend({
 
-  speakers: [{  name: 'Speaker 1', organisation: 'Org 1' },
-  { name: 'Speaker 2', organisation: 'Org 2', socialLinks: [{ name: 'facebook', url: '#' }] },
-  { name: 'Speaker 3', organisation: 'Org 3', socialLinks: [{ name: 'linkedin', url: '#' }] },
-  { name: 'Speaker 4', organisation: 'Org 4', socialLinks: [{ name: 'twitter', url: '#' }] },
-  { name: 'Speaker 5', organisation: 'Org 5', socialLinks: [{ name: 'home', url: '#' }] }],
+  speakers: [{  name: 'Speaker 1', organisation: 'Org 1', isFeatured: false, socialLinks: [{ name: 'facebook', url: '#' }]  },
+  { name: 'Speaker 2', organisation: 'Org 2', isFeatured: false, socailLinks: [{ name: 'facebook', url: '#' }] },
+  { name: 'Speaker 3', organisation: 'Org 3', isFeatured: true, socialLinks: [{ name: 'linkedin', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
+  { name: 'Speaker 4', organisation: 'Org 4', isFeatured: true, socialLinks: [{ name: 'twitter', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
+  { name: 'Speaker 5', organisation: 'Org 5', isFeatured: false, socialLinks: [{ name: 'home', url: '#' }],  short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }],
+
+  featuredSpeakers: computed.filterBy('speakers', 'isFeatured', true),
+
+  nonFeaturedSpeakers: computed.filterBy('speakers', 'isFeatured', false),
 
   htmlSafeDescription: computed('model.event.description', function() {
     return String.htmlSafe(this.get('model.event.description'));
--- a/app/styles/partials/speaker.scss
+++ b/app/styles/partials/speaker.scss
@@ -14,3 +14,23 @@
     }
   }
 }
+
+.speaker-description {
+  padding-right: 0;
+  word-wrap: break-word;
+  max-height: 240px;
+  overflow: hidden;
+  display: none;
+
+  &:hover {
+    overflow-y: auto;
+  }
+}
+
+.show {
+  display: block;
+}
+
+.featured-speaker {
+  margin-bottom: 10px;
+}
--- a/app/templates/components/public/featured-speaker-list.hbs
+++ b/app/templates/components/public/featured-speaker-list.hbs
@@ -1,6 +1,6 @@
-<h3>{{t 'Speakers'}}</h3>
+<h3>{{t 'Featured Speakers'}}</h3>
 <div class="speaker-list ui stackable grid">
   {{#each speakers as |speaker|}}
-    {{public/speaker-item speaker=speaker}}
+    {{public/speaker-item speaker=speaker isFeatured=true}}
   {{/each}}
 </div>
--- a/app/templates/components/public/speaker-item.hbs
+++ b/app/templates/components/public/speaker-item.hbs
@@ -1,12 +1,16 @@
-<div class="speaker-item">
-
-  <img alt="{{speaker.name}}" class="ui image" src="{{if speaker.image speaker.image '/images/placeholders/avatar.png'}}">
-  <br>
-  <div class="speaker-name">{{speaker.name}}</div>
-  <div class="speaker-organisation">{{speaker.organisation}}</div>
-  <div class="social-links">
-    {{#each speaker.socialLinks as |link|}}
-      <a href="{{link.url}}" target="_blank" rel="noopener noreferrer"><i class="bordered {{link.name}} colored inverted icon"></i></a>
-    {{/each}}
+<div class="speaker-item {{if speaker.isFeatured 'ui grid stackable'}} ">
+  <div class="speaker-image column eight wide">
+    <img alt="{{speaker.name}}" class="ui image" src="{{if speaker.image speaker.image '/images/placeholders/avatar.png'}}">
+    <br>
+    <div class="speaker-name">{{speaker.name}}</div>
+    <div class="speaker-organisation">{{speaker.organisation}}</div>
+    <div class="social-links">
+      {{#each speaker.socialLinks as |link|}}
+        <a href="{{link.url}}" target="_blank" rel="noopener noreferrer"><i class="bordered {{link.name}} colored inverted icon"></i></a>
+      {{/each}}
+    </div>
+  </div>
+  <div class="column seven wide speaker-description {{if speaker.isFeatured show}}">
+    {{speaker.short_biography}}
   </div>
 </div>
--- a/tests/integration/components/public/featured-speaker-list-test.js
+++ b/tests/integration/components/public/featured-speaker-list-test.js
@@ -1,0 +1,16 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('public/featured-speaker-list', 'Integration | Component | public/featured speaker list');
+
+const speaker = [
+  { name: 'Speaker 3', organisation: 'FOSSASIA', isFeatured: true, socialLinks: [{ name: 'linkedin', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }
+];
+
+test('it renders', function(assert) {
+
+  this.set('speaker', speaker);
+  this.render(hbs `{{public/featured-speaker-list speaker=speaker}}`);
+  assert.ok(this.$().html().trim().includes('Featured Speakers'));
+});
--- a/app/components/public/featured-speaker-list.js
+++ b/app/components/public/featured-speaker-list.js
@@ -1,0 +1,6 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+});
--- a/app/styles/pages/public-event.scss
+++ b/app/styles/pages/public-event.scss
@@ -87,4 +87,8 @@
   .description {
     text-align: justify;
   }
+
+  .speakers {
+    margin-top: 10px;
+  }
 }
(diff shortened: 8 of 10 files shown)


app/styles/partials/overrides.scss
```diff
@@ -18,3 +18,7 @@ body.dimmable.undetached.dimmed {
     position: static;
   }
 }
+
+.ui.grid.grid {
```
niranjan94: Please do not do global overrides like this unless there is a specific reason. This would override grids throughout the app.

app/templates/components/public/featured-speaker-item.hbs
```diff
@@ -0,0 +1,4 @@
+{{public/speaker-item speaker=speaker class="column eight wide"}}
```
niranjan94: Move this login into the speaker-item itself

app/templates/components/public/featured-speaker-list.hbs
```diff
@@ -0,0 +1,8 @@
+<h3>{{t 'Featured Speakers'}}</h3>
+<div class="speaker-list ui grid">
+  {{#each speakers as |speaker|}}
+    {{#if speaker.featured}}
```
niranjan94: Instead of looping through the entire list and checking, provide a filtered computed property called as `featuredSpeakers` and loop through that instead.

app/templates/components/public/speaker-list.hbs
```diff
@@ -1,6 +1,6 @@
 <h3>{{t 'Speakers'}}</h3>
 <div class="speaker-list ui grid">
   {{#each speakers as |speaker|}}
-    {{public/speaker-item speaker=speaker}}
+    {{public/speaker-item speaker=speaker class="column four wide"}}
```
niranjan94: Toggle the classes in the js file based on if the speaker is featured or not

app/styles/partials/speaker.scss
```diff
@@ -14,3 +14,9 @@
     }
   }
 }
+
+.speaker-description {
+  word-wrap: break-word;
+  height: 240px;
```
niranjan94: Don't fix the height like this. Instead, use `max-height`. And also, see how this is implemented in the current site. -> https://eventyay.com/e/5642d9a1/ . We get a scroll bar on hover if the speaker description is too big. We'd like the same feature here too. 

Our aim, is to first get the all the feature of the current UI into the new frontend.

geekyd:
I created a featured-speaker-item that uses the speaker-item for showing the image, name and social links, and we have description `short_biography` in featured-speaker-item. That is how it is currently structured.

niranjan94:
@geekyd yes. But that is an unnecessary complexity. 

A speaker will be rendered by a speaker-item. 
A featured speaker is also a speaker. And hence he/she will also be rendered by a speaker-item. Only, with a slight diff in UI based on a if featured or not..

geekyd:
Yes that seems a better design approach, Thanks :smile:

app/components/public/speaker-item.js
```diff
@@ -1,7 +1,26 @@
...
 
-const { Component } = Ember;
+const { Component, computed } = Ember;
 
 export default Component.extend({
-  classNames: ['column', 'four', 'wide']
+  classNameBindings: ['style'],
+
+  style: computed(function() {
+    const isFeatured = this.get('isFeatured');
```
niranjan94: Use a ternary conditional operator and make this single line

app/components/public/speaker-item.js
```diff
@@ -1,7 +1,26 @@
...
+    if (isFeatured) {
+      return 'column eight wide';
+    } else {
+      return 'column four wide';
+    }
+  }),
+  didInsertElement() {
+    const isFeatured = this.get('isFeatured');
+    if (isFeatured) {
+      this.$('.speaker-item').addClass(' ui grid stackable');
```
niranjan94: Make these class bindings via the template itself so that even if `isFeatured` changes in real-time, we can see class name changes. (Not useful right now. But just making it future proof)

app/styles/partials/overrides.scss
niranjan94: Still pending ?

app/components/public/speaker-item.js
geekyd: The changes are to be made only for `speaker-item` or for all the elements?

niranjan94: All items please. :)

geekyd:
@niranjan94 Are there any other changes that you want me to make?

niranjan94:
@geekyd thanks for asking 😛  .. I have been giving a lot of changes bit by bit lately right 😅 Sorry about that ... I keep noticing something _after_ I tell you ... I'll try to make sure I tell everything in one go so that you don't have to keep making changes ... Cool ? :) 

This looks good for now. (Apart from the changes I have already mentioned)

geekyd:
@niranjan94 That is totally fine. Thanks for quick responses. :smile:

app/components/public/speaker-item.js
geekyd: if we introduce the classes in the template it will be affecting the `**Speakers**` section too, these classes should only be added if the section is for featured speakers. That is why I have handled it in JS.
Please check this 
screenshot from 2017-05-10 13-13-42

niranjan94: @geekyd you can use if statements in templates too. 

`<div class="{{if speaker.isFeatured 'class-one' 'class-two'}}"></div>'

app/controllers/public/index.js
```diff
@@ -5,10 +5,14 @@ const { Controller, computed, String } = Ember;
...
+  { name: 'Speaker 4', organisation: 'Org 4', featured: true, socialLinks: [{ name: 'twitter', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
+  { name: 'Speaker 5', organisation: 'Org 5', featured: true, socialLinks: [{ name: 'home', url: '#' }] }],
+
+  featuredSpeakers: [
+  { name: 'Speaker 3', organisation: 'Org 3', featured: true, socialLinks: [{ name: 'linkedin', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
+  { name: 'Speaker 4', organisation: 'Org 4', featured: true, socialLinks: [{ name: 'twitter', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }],
```
niranjan94: `featured` will be called as `isFeatured` in the API. So, let's stick to the same.

app/components/public/speaker-item.js
geekyd: Okay so while computing the `featuredSpeakers` we will add a `isFeatured` attribute to the list and the list of `speakers` should not have `isFeatured` attribute. We are using `isFeatured` to distinguish featured speakers. So to avoid the above situation we should not have `isFeatured` in the `speakers` list.

app/controllers/public/index.js
geekyd: Okay :+1:

app/templates/components/public/speaker-list.hbs
```diff
@@ -1,6 +1,6 @@
 <h3>{{t 'Speakers'}}</h3>
-<div class="speaker-list ui grid">
+<div class="speaker-list ui stackable grid">
   {{#each speakers as |speaker|}}
-    {{public/speaker-item speaker=speaker}}
+    {{public/speaker-item speaker=speaker isFeatured=false}}
```
niranjan94: @geekyd in Ember.js an undefined parameter is by default `null`. So that will be resolved as a false by any if statement. `isFeatured=false` is not needed.

app/templates/components/public/speaker-item.hbs
```diff
@@ -1,12 +1,16 @@
...
-
-  <img alt="{{speaker.name}}" class="ui image" src="{{if speaker.image speaker.image '/images/placeholders/avatar.png'}}">
-  <br>
-  <div class="speaker-name">{{speaker.name}}</div>
-  <div class="speaker-organisation">{{speaker.organisation}}</div>
-  <div class="social-links">
-    {{#each speaker.socialLinks as |link|}}
-      <a href="{{link.url}}" target="_blank" rel="noopener noreferrer"><i class="bordered {{link.name}} colored inverted icon"></i></a>
-    {{/each}}
+<div class="speaker-item {{ if speaker.isFeatured 'ui grid stackable'}} ">
```
niranjan94: No space between `{{` and `if` please

app/components/public/speaker-item.js
```diff
@@ -1,7 +1,18 @@
...
+  classNameBindings: ['style'],
+
+  style: computed(function() {
+    const allClasses = this.get('isFeatured') ? 'column eight wide' : 'column four wide';
+    return allClasses;
+  }),
+  didInsertElement() {
+    const isFeatured = this.get('isFeatured');
+    if (isFeatured) {
+      this.$('.speaker-description').removeClass('hide');
```
niranjan94: I think you missed one class.

app/controllers/public/index.js
```diff
@@ -6,10 +6,14 @@ export default Controller.extend({
 
   speakers: [{  name: 'Speaker 1', organisation: 'Org 1' },
```
niranjan94: Why two arrays ? The API response is going to be a single array of speakers. Use the `isFeatured` flag to differentiate.

app/templates/components/public/featured-speaker-list.hbs
geekyd: As you have mentioned here instead of looping through all speakers we will compute featuredSpeakers and loop over it. That is the reason for having two separate arrays.

niranjan94: Yes. So, you need a computed property named featuredSpeakers right ? Instead of two arrays ?

app/controllers/public/index.js
geekyd: Single array is a good option but that might cause us some problems. we are checking isFeatured attribute to change classes of the HTML in the `speaker-item` which is commonly used by Speakers and Featured Speakers, this will lead to change of layout for featured speaker in the Speakers section too.

app/templates/components/public/featured-speaker-list.hbs
niranjan94: We'll actually be needing two computed properties. featuredSpeakers and nonFeaturedSpeakers based on one speakers array.

niranjan94: https://www.emberjs.com/api/classes/Ember.computed.html#method_filter

app/controllers/public/index.js
niranjan94: There will be no featured speakers in the speakers section.

niranjan94: We'll actually be needing two computed properties. featuredSpeakers and nonFeaturedSpeakers based on one speakers array.

niranjan94: https://www.emberjs.com/api/classes/Ember.computed.html#method_filter

app/templates/components/public/featured-speaker-list.hbs
niranjan94: Use filterBy: https://www.emberjs.com/api/classes/Ember.computed.html#method_filterBy

app/controllers/public/index.js
niranjan94: You can use filterBy to easily filter
Use filterBy: https://www.emberjs.com/api/classes/Ember.computed.html#method_filterBy

geekyd:
@niranjan94 Please review, I have added filters. 
Screenshot:
screenshot from 2017-05-10 19-53-12

app/controllers/public/index.js
```diff
@@ -4,11 +4,15 @@ const { Controller, computed, String } = Ember;
...
+  speakers: [{  name: 'Speaker 1', organisation: 'Org 1', isFeatured: false, socialLinks: [{ name: 'facebook', url: '#' }]  },
+  { name: 'Speaker 2', organisation: 'Org 2', isFeatured: false, socailLinks: [{ name: 'facebook', url: '#' }] },
+  { name: 'Speaker 3', organisation: 'Org 3', isFeatured: true, socialLinks: [{ name: 'linkedin', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
+  { name: 'Speaker 4', organisation: 'Org 4', isFeatured: true, socialLinks: [{ name: 'twitter', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
+  { name: 'Speaker 5', organisation: 'Org 5', isFeatured: false, socialLinks: [{ name: 'home', url: '#' }],  short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }],
+
+  filterByFeaturedSpeakers: computed.filterBy('speakers', 'isFeatured', true),
```
niranjan94: Let the variable names be simple and as close to english language as possible.
`featuredSpeakers` and `nonFeaturedSpeakers`

tests/integration/components/public/featured-speaker-list-test.js
```diff
@@ -0,0 +1,16 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('public/featured-speaker-list', 'Integration | Component | public/featured speaker list');
+
+const speaker = [
+  { name: 'Speaker 3', organisation: 'FOSSASIA', featured: true, socialLinks: [{ name: 'linkedin', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }
```
niranjan94: You still have `featured: true` here

geekyd:
@niranjan94 Please review.

app/components/public/speaker-item.js
```diff
@@ -1,7 +1,12 @@
 import Ember from 'ember';
 
-const { Component } = Ember;
+const { Component, computed } = Ember;
 
 export default Component.extend({
-  classNames: ['column', 'four', 'wide']
+  classNameBindings: ['style'],
+
+  style: computed(function() {
```
niranjan94: Ahh missed one thing. You need to specify computed to observe a variable for changes. Here, we need `style` to be recomputed every time `isFeatured` changes. So, you need to specify the name of the variable to observer as the first argument of `computed`. You can specify multiple variables as multiple arguments.

geekyd:
@niranjan94 Made the requested changes, please review & merge.

PR #52, opened 2017-05-16
Adds sponsors layout to public events

### LINKED ISSUE #49: Adding sponsors to public events
Add sponsors to the public events page like on eventyay.com.
screenshot from 2017-05-14 17-04-48

geekyd:
#### Checklist

- Adds layout for sponsors to the public events page

####Screenshot
screenshot from 2017-05-16 12-07-31

Fixes #49 

@niranjan94 Please review. Thanks

### DIFF
--- a/app/components/public/sponsor-list.js
+++ b/app/components/public/sponsor-list.js
@@ -1,0 +1,10 @@
+import Ember from 'ember';
+import { orderBy, groupBy } from 'lodash';
+
+const { Component, computed } = Ember;
+
+export default Component.extend({
+  sponsorsGrouped: computed('sponsors.[]', function() {
+    return groupBy(orderBy(this.get('sponsors'), 'level'), 'type');
+  })
+});
--- a/app/controllers/public/index.js
+++ b/app/controllers/public/index.js
@@ -10,6 +10,11 @@ export default Controller.extend({
   { name: 'Speaker 4', organisation: 'Org 4', isFeatured: true, socialLinks: [{ name: 'twitter', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
   { name: 'Speaker 5', organisation: 'Org 5', isFeatured: false, socialLinks: [{ name: 'home', url: '#' }],  short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }],
 
+  sponsors: [{ name: 'Sponsor 2', url: '#', logoUrl: 'http://placehold.it/150x60', level: 2, type: 'Gold Sponsor', description: '' },
+    { name: 'Sponsor 1', url: '#', logoUrl: 'http://placehold.it/150x60', level: 1, type: 'Gold Sponsor', description: '' },
+    { name: 'Sponsor 3', url: '#', logoUrl: 'http://placehold.it/150x60', level: 1, type: 'Silver Sponsor', description: '' }
+  ],
+
   featuredSpeakers: computed.filterBy('speakers', 'isFeatured', true),
 
   nonFeaturedSpeakers: computed.filterBy('speakers', 'isFeatured', false),
--- a/app/templates/components/public/sponsor-item.hbs
+++ b/app/templates/components/public/sponsor-item.hbs
@@ -1,0 +1,3 @@
+<a href="{{sponsor.url}}">
+  <img src="{{sponsor.logoUrl}}" class="ui image sponsor-image" alt="{{sponsor.name}}">
+</a>
--- a/app/templates/components/public/sponsor-list.hbs
+++ b/app/templates/components/public/sponsor-list.hbs
@@ -1,0 +1,9 @@
+<h3 class="ui header">{{t 'Sponsors'}}</h3>
+{{#each-in sponsorsGrouped as |key sponsors|}}
+  <h4 class="ui header">{{key}}</h4>
+  <div class="ui three column stackable grid">
+    {{#each sponsors as |sponsor|}}
+      {{public/sponsor-item sponsor=sponsor}}
+    {{/each}}
+  </div>
+{{/each-in}}
--- a/app/templates/public/index.hbs
+++ b/app/templates/public/index.hbs
@@ -12,4 +12,10 @@
       {{public/speaker-list speakers=nonFeaturedSpeakers}}
     </div>
   </div>
+  <br>
+  <div class="sponsors">
+    <div class="row">
+      {{public/sponsor-list sponsors=sponsors}}
+    </div>
+  </div>
 </div>
--- a/app/components/public/sponsor-item.js
+++ b/app/components/public/sponsor-item.js
@@ -1,0 +1,7 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  classNames: ['ui', 'column']
+});
--- a/app/styles/pages/public-event.scss
+++ b/app/styles/pages/public-event.scss
@@ -91,4 +91,8 @@
   .speakers {
     margin-top: 10px;
   }
+
+  .sponsor-image {
+    margin: 0 auto;
+  }
 }
--- a/app/templates/components/public/speaker-list.hbs
+++ b/app/templates/components/public/speaker-list.hbs
@@ -1,4 +1,4 @@
-<h3>{{t 'Speakers'}}</h3>
+<h3 class="ui header">{{t 'Speakers'}}</h3>
 <div class="speaker-list ui stackable grid">
   {{#each speakers as |speaker|}}
     {{public/speaker-item speaker=speaker}}
--- a/tests/integration/components/public/sponsor-item-test.js
+++ b/tests/integration/components/public/sponsor-item-test.js
@@ -1,0 +1,15 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('sponsor-item', 'Integration | Component | sponsor item');
+
+const sponsor = [
+  { name: 'Sponsor 1', url: '#', logoUrl: 'http://placehold.it/150x60' }
+];
+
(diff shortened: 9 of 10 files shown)


app/controllers/public/index.js
```diff
@@ -10,6 +10,10 @@ export default Controller.extend({
   { name: 'Speaker 4', organisation: 'Org 4', isFeatured: true, socialLinks: [{ name: 'twitter', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
   { name: 'Speaker 5', organisation: 'Org 5', isFeatured: false, socialLinks: [{ name: 'home', url: '#' }],  short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }],
 
+  sponsors: [{ title: 'Gold Sponsors', sponsorList: [{ name: 'Sponsor 1', url: '#', logo: 'http://placehold.it/150x60' }, { name: 'Sponsor 2', url: '#', logo: 'http://placehold.it/150x60' }, { name: 'Sponsor 3', url: '#', logo: 'http://placehold.it/150x60' }] },
```
niranjan94: The response will be an array like this:

```js
[
{
    "name": "Sponsor 1",
    "logoUrl": "https://some.com/a.jpg",
    "level": 2,
    "type": "Gold Sponsor"
},
{
    "name": "Sponsor 1",
    "logoUrl": "https://some.com/a.jpg",
    "level": 2,
    "type": "Silver Sponsor"
}
]
```

This will be the API response.
You will have to group by `type` and within each type group, you should sort by `level` in `ascending`

geekyd: @niranjan94 Okay so, should I write a custom function to group the sponsors by type, and sort the sponsors in each group by using `sort` computedProperty?

niranjan94: Correct. But you don't have to use a custom function. Use [lodash's `groupBy`](https://lodash.com/docs/4.17.4#groupBy). Lodash is included in the project. Feel free to use all of the awesome utility functions there 😄 .

Since we'll be sorting and grouping, if you're using `computed.sort` we'll probably need two computed properties. (One for sorting and that passed through another which does grouping.)

Instead use one normal computed property and use lodash's `orderBy` and `groupBy`

niranjan94: @geekyd ^

geekyd: @niranjan94 Thanks for the reference. There is one problem that need to be addressed here, current api is returning the sponsors as : 

`{
"description": "",
"level": "3", 
"url": "URL of sponsor(website)", 
"sponsor_type": "", 
"logo": "URL of image", 
"id": '', 
"name": ""}
`

For webapp we used to group according to the levels and render the sponsors, will we be following a similar approach ? The level will be unique to the corresponding type?

niranjan94: We want the ability to specify order with a sponsor group too. So first sort by level. The group by type the sorted array (let the grouping be case-insensitive to the type).

geekyd:
@niranjan94 Made the required changes here is the screenshot. Please review Thanks :)
screenshot from 2017-05-16 21-01-37

app/controllers/public/index.js
```diff
@@ -10,6 +11,19 @@ export default Controller.extend({
   { name: 'Speaker 4', organisation: 'Org 4', isFeatured: true, socialLinks: [{ name: 'twitter', url: '#' }], short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' },
   { name: 'Speaker 5', organisation: 'Org 5', isFeatured: false, socialLinks: [{ name: 'home', url: '#' }],  short_biography: 'Full-stack developer with experience in PHP, Python &amp; Ember.JS. Plays around with Kubernetes and Containers. Avid open source contributor.' }],
 
+  sponsors: [{ name: 'Sponsor 2', url: '#', logoUrl: 'http://placehold.it/150x60', level: 2, type: 'Gold Sponsor', description: 'Blah Blah Blah' },
+    { name: 'Sponsor 1', url: '#', logoUrl: 'http://placehold.it/150x60', level: 1, type: 'Gold Sponsor', description: 'Blah Blah Blah' },
+    { name: 'Sponsor 3', url: '#', logoUrl: 'http://placehold.it/150x60', level: 1, type: 'Silver Sponsor', description: 'Blah Blah Blah' }
+  ],
+
+  sponsorsByCategory: computed(('sponsors'), function() {
```
niranjan94: @geekyd 

This is too complex. We're doing needles operations here. this can be simplified to just a single sortBy and a groupBy. In a javascript single page application like this, we need to take atmost care to ensure we use as less memory and resources as possible, since it all happens on the client side and we can't always expect our clients to have good PCs.

See this alternative implementation,

```js
sponsorsByCategory: computed('sponsors', function() {
    return _.groupBy(_.orderBy(this.get('sponsors'), 'level'), 'type');;
})
```

This will now return an object in the form 

```js
{
  "Gold Sponsor": [
    {
      "name": "Sponsor 1",
      "url": "#",
      "logoUrl": "http://placehold.it/150x60",
      "level": 1,
      "type": "Gold Sponsor",
      "description": "Blah Blah Blah"
    },
    {
      "name": "Sponsor 2",
      "url": "#",
      "logoUrl": "http://placehold.it/150x60",
      "level": 2,
      "type": "Gold Sponsor",
      "description": "Blah Blah Blah"
    }
  ],
  "Silver Sponsor": [
    {
      "name": "Sponsor 3",
      "url": "#",
      "logoUrl": "http://placehold.it/150x60",
      "level": 1,
      "type": "Silver Sponsor",
      "description": "Blah Blah Blah"
    }
  ]
}
```

Now, in the template, You can use [`{{#each-in object as |key value|}}  {{/each-in}}`](https://www.emberjs.com/api/classes/Ember.Templates.helpers.html#method_each-in) instead of `each` to loop through this.

app/controllers/public/index.js
```diff
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import _ from 'lodash';
```
niranjan94: Instead of importing all ... Import only what is needed. Like 

```js
import { orderBy, groupBy } from 'lodash';
```

app/controllers/public/index.js
niranjan94: And also, since we want the computed property to observe element changes in the array, 
it needs to be 
```js
sponsorsByCategory: computed('sponsors.[]', function() {
```

If you use just `sponsors`, the property will recompute only if for example `sponsors` changes from `null` to an array or back. The property will **not** recompute if for example, an item is added to the `sponsors` array.

But we want it to be recomputed if the array elements are added/removed. So we specify it as 
```js
sponsorsByCategory: computed('sponsors.[]', function() {
```

app/templates/components/public/sponsor-list.hbs
```diff
@@ -0,0 +1,9 @@
+<h3 class="ui header">{{t 'Sponsors'}}</h3>
+{{#each-in sponsors as |key value|}}
```
niranjan94: Also, instead of value, use `sponsors`. (For better readability)

app/templates/components/public/sponsor-list.hbs
```diff
@@ -0,0 +1,9 @@
...
+  <h4 class="ui header">{{key}}</h4>
+  <div class="ui three column stackable grid">
+    {{#each value as |sponsorItem|}}
```
niranjan94: instead of sponsorItem, just call it sponsor

app/templates/public/index.hbs
```diff
@@ -12,4 +12,10 @@
       {{public/speaker-list speakers=nonFeaturedSpeakers}}
     </div>
   </div>
+  <br>
+  <div class="sponsors">
+    <div class="row">
+      {{public/sponsor-list sponsors=sponsorsByCategory}}
```
niranjan94: Let the sponsors grouping be done by the sponsor-list component. This is the components job. You just should pass the plain sponsors array to it.

geekyd: Okay, I will be creating a controller for the speaker-list and handle it there.

niranjan94: No. I think you misunderstood. Components cannot have controllers. Routes have controllers. 

You can directly put the same code you used in the index controller inside the component js file.

geekyd:
@niranjan94 I have made the required changes, moved the grouping sponsors code to the component itself. Please review & merge. :)

app/components/public/sponsor-list.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember';
+import { orderBy, groupBy } from 'lodash';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  init() {
+    this._super(...arguments);
+    this.set('sponsors', groupBy(orderBy(this.get('sponsors'), 'level'), 'type'));
```
niranjan94: Please add this as a computed property. Otherwise, this will not change incase sponsors array changes.

app/templates/components/public/sponsor-list.hbs
```diff
@@ -0,0 +1,9 @@
+<h3 class="ui header">{{t 'Sponsors'}}</h3>
+{{#each-in sponsors as |key sponsor|}}
```
niranjan94: Call it `sponsorsGrouped`

app/templates/components/public/sponsor-list.hbs
```diff
@@ -0,0 +1,9 @@
...
+  <h4 class="ui header">{{key}}</h4>
+  <div class="ui three column stackable grid">
+    {{#each sponsor as |sponsor|}}
```
niranjan94: And call this as `sponsors as |sponsor|`

geekyd:
@niranjan94 Please review.

app/templates/components/public/sponsor-item.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="ui">
```
niranjan94: Is this div really required ?

geekyd: @niranjan94 No, removing this div.

geekyd:
@niranjan94 Made the requested changes.

PR #68, opened 2017-05-22
Adds sessions layout public event

### LINKED ISSUE #63: Implementing sessions layout for public events(UI)
- [x] feature request

Currently the sessions layout is missing from the public events. Add sessions to the public events as shown in the screenshot.
screenshot from 2017-05-20 15-19-10

@niranjan94 I will be working on this.

geekyd:

Adds session layout in public events page
Added components
- Session-filter
- Session-list
- Session-item

####Screenshot
screenshot from 2017-05-22 10-25-03

Fixes #63 
@niranjan94 Please review.

### DIFF
--- a/app/components/public/session-filter.js
+++ b/app/components/public/session-filter.js
@@ -1,0 +1,12 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+
+  actions: {
+    filter(trackId = null) {
+      this.set('selectedTrackId', trackId);
+    }
+  }
+});
--- a/app/components/public/session-item.js
+++ b/app/components/public/session-item.js
@@ -1,0 +1,7 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  classNames: ['ui', 'segment']
+});
--- a/app/controllers/public/sessions.js
+++ b/app/controllers/public/sessions.js
@@ -1,0 +1,30 @@
+import Ember from 'ember';
+import { chain } from 'lodash';
+
+const { Controller, computed } = Ember;
+
+export default Controller.extend({
+  sessions: [{ session_type: { length: '11:00', id: 12, name: 'Talk' }, shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: new Date(), endAt: new Date(), track: { color: 'green', fontColor: 'green', id: 1, name: 'Track 1' }, microlocation: { id: 2, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] },
+  { session_type: { length: '15:00', id: 11, name: 'Talk' }, shortAbstract: 'Welcome to open source', id: '2', title: 'Introduction to Open source', startAt: new Date(), endAt: new Date(), track: { color: 'red', fontColor: 'white', id: 2, name: 'Track 2' }, microlocation: { id: 3, name: 'Room 5' }, speakers: [{ shortBiography: '', id: 0, city: 'Delhi', name: 'Tux', speakingExperience: 'Speaker at the high conference', organisation: 'ORG 2', longBiography: '', photo: { 'iconImageUrl': '' } }] }
+  ],
+  sessionsByTracks: computed('sessions.[]', 'selectedTrackId', function() {
+    if (this.get('selectedTrackId')) {
+      return chain(this.get('sessions'))
+        .filter(['track.id', this.get('selectedTrackId')])
+        .orderBy(['track.name', 'startAt'])
+        .value();
+    } else {
+      return chain(this.get('sessions'))
+        .orderBy(['track.name', 'startAt'])
+        .groupBy('track.name')
+        .value();
+    }
+  }),
+  tracks: computed('sessions.[]', function() {
+    return chain(this.get('sessions'))
+      .map('track')
+      .uniqBy('id')
+      .orderBy('name')
+      .value();
+  })
+});
--- a/app/templates/components/public/session-filter.hbs
+++ b/app/templates/components/public/session-filter.hbs
@@ -1,0 +1,10 @@
+<div class="ui secondary menu">
+  <a href="#" class="item {{unless selectedTrackId 'active'}}" {{action 'filter'}}>
+    All
+  </a>
+  {{#each-in tracks as |id track|}}
+    <a href="#" class="item {{if (eq selectedTrackId track.id) 'active'}}" {{action 'filter' track.id}}>
+      {{track.name}}
+    </a>
+  {{/each-in}}
+</div>
--- a/app/templates/components/public/session-item.hbs
+++ b/app/templates/components/public/session-item.hbs
@@ -1,0 +1,54 @@
+{{#ui-accordion}}
+  <div class="title">
+    <div class="ui">
+      <h3 class="ui header">
+        {{session.title}}
+        <div class="sub header">
+          {{session.sessionType.name}}
+        </div>
+      </h3>
+      <div class="ui row">Talk</div>
+    </div>
+    <br>
+    <div class="ui grid">
+      <div class="row">
+        <div class="column twelve wide">
+          <div class="session-speakers">
+            {{#each session.speakers as |speaker|}}
+              <img alt="speaker" class="ui mini avatar image" src="{{if speaker.photo.iconImageUrl speaker.image '/images/placeholders/avatar.png'}}">
+            {{/each}}
+          </div>
+        </div>
+        <div class="ui right aligned column four wide" style="padding-left: 0">
+          <div class="">{{session.microlocation.name}} <i class="marker icon"></i></div>
+          <div class="">{{moment-format session.startAt 'hh:mm a / DD-MM-YYYY'}} <i class="wait icon"></i></div>
+        </div>
+      </div>
+      <div class="row">
+        <div class="column session-description">
+          {{session.shortAbstract}}
+        </div>
+      </div>
+    </div>
+  </div>
+  <div class="content">
+    {{#each session.speakers as |speaker|}}
+      <h3 class="ui header">About {{speaker.name}}:</h3>
+      <p>
+        {{#if speaker.shortBiography}}
+          {{speaker.shortBiography}}
+        {{else if speaker.longBiography}}
+          {{speaker.longBiography}}
+        {{else}}
+          {{speaker.speakingExperience}}
+        {{/if}}
+      </p>
+      <div class="social-icons">
+        <a href="#" target="_blank" rel="noopener noreferrer"><i class="facebook f icon"></i>Facebook</a>
+        <a href="#" target="_blank" rel="noopener noreferrer"><i class="linkedin f icon"></i>Linkedin</a>
+        <a href="#" target="_blank" rel="noopener noreferrer"><i class="twitter icon"></i>Twitter</a>
+        <a href="#" target="_blank" rel="noopener noreferrer"><i class="github icon"></i>Github</a>
+      </div>
+    {{/each}}
+  </div>
+{{/ui-accordion}}
--- a/app/templates/components/public/session-list.hbs
+++ b/app/templates/components/public/session-list.hbs
@@ -1,0 +1,15 @@
+<div class="ui session-container">
+  {{#if (is-array sessions)}}
+    <h3 class="ui header">{{track}}</h3>
+    {{#each sessions as |session|}}
+      {{public/session-item session=session}}
+    {{/each}}
+  {{else}}
+    {{#each-in sessions as |track trackSession|}}
+      <h3 class="ui header">{{track}}</h3>
+      {{#each trackSession as |session|}}
+        {{public/session-item session=session}}
+      {{/each}}
+    {{/each-in}}
+  {{/if}}
+</div>
--- a/app/templates/public/sessions.hbs
+++ b/app/templates/public/sessions.hbs
@@ -1,1 +1,4 @@
-{{outlet}}
+<h2 class="ui header">Sessions</h2>
+{{public/session-filter selectedTrackId=selectedTrackId tracks=tracks}}
+
+{{public/session-list sessions=sessionsByTracks}}
--- a/app/components/public/session-list.js
+++ b/app/components/public/session-list.js
@@ -1,0 +1,6 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+});
--- a/app/styles/pages/public-event.scss
+++ b/app/styles/pages/public-event.scss
@@ -99,4 +99,8 @@
   .tickets {
     margin-bottom: 20px;
   }
+
+  .session-container {
+    margin-top: 15px;
+  }
 }
--- a/tests/integration/components/public/session-filter-test.js
+++ b/tests/integration/components/public/session-filter-test.js
@@ -1,0 +1,13 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('public/session-filter', 'Integration | Component | public/session filter');
+
+const tracks = { '1': { name: 'Track 1', id: '1' } };
+
+test('it renders', function(assert) {
+  this.set('tracks', tracks);
+  this.render(hbs `{{public/session-filter tracks=tracks}}`);
+  assert.ok(this.$().html().trim().includes('Track'));
+});
--- a/tests/integration/components/public/session-item-test.js
+++ b/tests/integration/components/public/session-item-test.js
@@ -1,0 +1,13 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('public/session-item', 'Integration | Component | public/session item');
+
+const session = { session_type: { length: '11:00', id: 12, name: 'Talk' }, shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: '10:35 AM / 17-03-2017', track: { color: 'green', fontColor: 'green', id: 0, name: 'Track 1' }, microlocation: { id: 0, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] };
+
+test('it renders', function(assert) {
+  this.set('session', session);
+  this.render(hbs `{{public/session-item session=session}}`);
+  assert.ok(this.$().html().trim().includes('Room 1'));
+});
--- a/tests/integration/components/public/session-list-test.js
+++ b/tests/integration/components/public/session-list-test.js
@@ -1,0 +1,13 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('public/session-list', 'Integration | Component | public/session list');
+
+const sessions = { 'Track 1': [{ session_type: { length: '11:00', id: 12, name: 'Talk' }, shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: '10:35 AM / 17-03-2017', track: { color: 'green', fontColor: 'green', id: 1, name: 'Track 1' }, microlocation: { id: 0, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] }] };
+
+test('it renders', function(assert) {
+  this.set('sessions', sessions);
+  this.render(hbs `{{public/session-list sessions=sessions}}`);
+  assert.ok(this.$().html().trim().includes('Track'));
+});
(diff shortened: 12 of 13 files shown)


sumedh123:
@geekyd Looks good, but I think it would look better if we increase some space between the "Talk" and the image.

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,14 @@
+import Ember from 'ember';
+import { orderBy, groupBy } from 'lodash';
+
+const { Controller, computed } = Ember;
+
+export default Controller.extend({
+  sessions: [{ shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: '10:35 AM / 17-03-2017', track: { color: 'green', fontColor: 'green', id: 0, name: 'Track 1' }, microlocation: { id: 0, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] },
```
niranjan94: Like I said in the issue, the startsAt and endsAt fields will be a js Date object. not a string.

app/helpers/underscore-text.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember';
+
+const { Helper } = Ember;
+
+export function underscoreText(params) {
```
niranjan94: Please remove this helper and use the `underscore` helper from [ember-cli-string-helpers](https://github.com/romulomachado/ember-cli-string-helpers) which is already a part of this project.

app/templates/components/public/session-filter.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="ui segment">
+  {{#each-in tracks as |track|}}
```
niranjan94: I think tabs would look better. And when a user presses on a track, don't scroll to it. Instead filter the whole list to display only that track.

app/controllers/public/sessions.js
geekyd: @niranjan94 As we are going to use model for getting sessions, it would be better if we use parse the dates in the model.

niranjan94: The model's output is gonna be a js date object. (It will parse the API Response and give a js output) ... And we shouldn't do the string conversion in the model ... because the string date format we require might vary depending on where the session is being displayed in the UI.

niranjan94: Like for example, 
- in this page we may want to display it in a more human readable form like `March 23, 2017 05:30 PM`
- In the sessions table we'll display `23/03/2017 05:30 PM`
- In the form we may display in another way for it to work with the date picker.

We can't and shouldn't put all of this inside the model.

geekyd: Okay got it!

geekyd:
@niranjan94 Please review.

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,26 @@
...
+  { shortAbstract: 'Welcome to open source', id: '2', title: 'Introduction to Open source', startAt: moment(new Date().setMinutes(10)).format('h:mm a / DD-MM-YYYY'), track: { color: 'red', fontColor: 'white', id: 0, name: 'Track 2' }, microlocation: { id: 3, name: 'Room 5' }, speakers: [{ shortBiography: '', id: 0, city: 'Delhi', name: 'Tux', speakingExperience: 'Speaker at the high conference', organisation: 'ORG 2', longBiography: '', photo: { 'iconImageUrl': '' } }] }
+  ],
+  sessionsByTracks: computed('sessions.[]', function() {
+    return groupBy(orderBy(this.get('sessions'), 'startAt'), 'track.name');
+  }),
+  tracks: computed('sessions.[]', function() {
+    return keys(groupBy(orderBy(this.get('sessions'), 'startAt'), 'track.name'));
+  }),
+  actions: {
+    filterSession(element) {
```
niranjan94: Remove this action.

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,26 @@
...
+import moment from 'moment';
+import { orderBy, groupBy, pick, keys } from 'lodash';
+
+const { Controller, computed } = Ember;
+
+export default Controller.extend({
+  sessions: [{ shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: moment(new Date()).format('h:mm a / DD-MM-YYYY'), track: { color: 'green', fontColor: 'green', id: 0, name: 'Track 1' }, microlocation: { id: 2, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] },
+  { shortAbstract: 'Welcome to open source', id: '2', title: 'Introduction to Open source', startAt: moment(new Date().setMinutes(10)).format('h:mm a / DD-MM-YYYY'), track: { color: 'red', fontColor: 'white', id: 0, name: 'Track 2' }, microlocation: { id: 3, name: 'Room 5' }, speakers: [{ shortBiography: '', id: 0, city: 'Delhi', name: 'Tux', speakingExperience: 'Speaker at the high conference', organisation: 'ORG 2', longBiography: '', photo: { 'iconImageUrl': '' } }] }
+  ],
+  sessionsByTracks: computed('sessions.[]', function() {
```
niranjan94: Let this computed function depend on `sessions.[]` and `selectedTrackId`

You don't have to declare `selectedTrackId` anywhere. it is by default null. If it is null consider it to be all tracks. Else filter by that track id

app/templates/public/sessions.hbs
```diff
@@ -1 +1,4 @@
-{{outlet}}
+<h2 class="ui header">Sessions</h2>
+{{public/session-filter filterSession=(action 'filterSession') tracks=tracks}}
```
niranjan94: Change to 
```hbs
{{public/session-filter selectedTrackId=selectedTrackId tracks=tracks}}
```

Note that you're passing in the `selectedTrackId` variable

app/components/public/session-filter.js
```diff
@@ -0,0 +1,14 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  selectedTrack: 'all',
+
+  actions: {
+    filter(element, event) {
+      this.set('selectedTrack', event.target.id);
```
niranjan94: Change to 
```js
this.set('selectedTrackId', trackId);
```

app/components/public/session-filter.js
```diff
@@ -0,0 +1,14 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  selectedTrack: 'all',
+
+  actions: {
+    filter(element, event) {
```
niranjan94: Change to 
```js
filter(trackId = null) {
```

If no trackId is passed it is considered as `null`

app/templates/components/public/session-filter.hbs
```diff
@@ -0,0 +1,10 @@
+<div class="ui secondary menu">
+  <a href="#" class="item {{if (eq selectedTrack 'all') 'active'}}" id="all" onclick={{action 'filter' 'All'}}>
```
niranjan94: Change to 

```hbs
<a href="#" class="item {{unless selectedTrackId 'active'}}" {{action 'filter'}}>
```

Here if `selectedTrackId` is null it is considered to be all tracks. 
And for action you don't have to specify `onclick=`
And note that we're not passing any argument to the action. hence trackId in the action will be null.

app/templates/components/public/session-filter.hbs
```diff
@@ -0,0 +1,10 @@
...
+    All
+  </a>
+  {{#each tracks as |track|}}
+    <a href="#" class="item {{if (eq selectedTrack (underscore track)) 'active'}}" id="{{underscore track}}" onclick={{action 'filter' track}}>
```
niranjan94: Can be changed to 
```hbs
<a href="#" class="item {{if (eq selectedTrackId track.id) 'active'}}" {{action 'filter' track.id}}>
{{track.name}}
</a>
```

niranjan94:
Now from any where in that page, all I have to do is change `selectedTrackId` and everything related to it will change.

geekyd:
@niranjan94 Thank you for the detailed review and proposing a better solution. I have made the required changes. Please review it.

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,28 @@
...
+      return { name: temp.track.name, id: temp.track.id, sessions: v };
+    });
+    if (this.get('selectedTrackId')) {
+      return pick(allSessions, this.get('selectedTrackId'));
+    } else {
+      return allSessions;
+    }
+  }),
+  tracks: computed('sessions.[]', function() {
+    var allTracks = mapValues(keyBy(orderBy(this.get('sessions'), 'startAt'), 'track.id'), function(v) {
```
niranjan94: This can be better written as 
```js
chain(this.get('sessions'))
    .map('track')
    .uniqBy('id')
    .orderBy('name')
    .value()
```
_`chain` is also a lodash method._

That way we don't have to order _all_ the sessions before getting the tracks and we get all contents of the track object.

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default Controller.extend({
+  sessions: [{ shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: moment(new Date()).format('h:mm a / DD-MM-YYYY'), track: { color: 'green', fontColor: 'green', id: 1, name: 'Track 1' }, microlocation: { id: 2, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] },
+  { shortAbstract: 'Welcome to open source', id: '2', title: 'Introduction to Open source', startAt: moment(new Date().setMinutes(10)).format('h:mm a / DD-MM-YYYY'), track: { color: 'red', fontColor: 'white', id: 2, name: 'Track 2' }, microlocation: { id: 3, name: 'Room 5' }, speakers: [{ shortBiography: '', id: 0, city: 'Delhi', name: 'Tux', speakingExperience: 'Speaker at the high conference', organisation: 'ORG 2', longBiography: '', photo: { 'iconImageUrl': '' } }] }
+  ],
+  sessionsByTracks: computed('sessions.[]', 'selectedTrackId', function() {
+    var allSessions = mapValues(groupBy(orderBy(this.get('sessions'), 'startAt'), 'track.id'), function(v) {
```
niranjan94: This can be written as 

Unfiltered:
```js
chain(this.get('sessions'))
    .orderBy(['track.name', 'name'])
    .groupBy('track.name')
    .value();
```

Filtered:
```js
chain(this.get('sessions'))
    .filter(['track.id', this.get('selectedTrackId')])
    .orderBy(['track.name', 'name'])
    .groupBy('track.name')
    .value();
```

Ordering and grouping are the operation that usually take up more time. So all filtering should be done _before_ them.

niranjan94: This will give an output like 

```js
{
    'Track name': []
}
```

Where `[]` is an array of sessions. You can use `each-in` in the template to loop through this and get both track name and the sessions.

app/controllers/public/sessions.js
niranjan94: The output will be an array of track objects.

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,28 @@
+import Ember from 'ember';
+import moment from 'moment';
+import { orderBy, pick, groupBy, keyBy, mapValues, last } from 'lodash';
+
+const { Controller, computed } = Ember;
+
+export default Controller.extend({
+  sessions: [{ shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: moment(new Date()).format('h:mm a / DD-MM-YYYY'), track: { color: 'green', fontColor: 'green', id: 1, name: 'Track 1' }, microlocation: { id: 2, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] },
```
niranjan94: Again ... like I said `startAt` and `endAt` will NOT be a string. It will be a date object. You need to use [ember-moment](https://github.com/stefanpenner/ember-moment) template helpers to display it how you like.

geekyd:
@niranjan94 Please review.

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,33 @@
...
+
+export default Controller.extend({
+  sessions: [{ shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: new Date(), endAt: new Date(), track: { color: 'green', fontColor: 'green', id: 1, name: 'Track 1' }, microlocation: { id: 2, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] },
+  { shortAbstract: 'Welcome to open source', id: '2', title: 'Introduction to Open source', startAt: new Date(), endAt: new Date(), track: { color: 'red', fontColor: 'white', id: 2, name: 'Track 2' }, microlocation: { id: 3, name: 'Room 5' }, speakers: [{ shortBiography: '', id: 0, city: 'Delhi', name: 'Tux', speakingExperience: 'Speaker at the high conference', organisation: 'ORG 2', longBiography: '', photo: { 'iconImageUrl': '' } }] }
+  ],
+  sessionsByTracks: computed('sessions.[]', 'selectedTrackId', function() {
+    var allSessions = chain(this.get('sessions'))
+      .orderBy(['startAt', 'startAt'])
+      .groupBy('track.id')
+      .mapValues(function(v) {
```
niranjan94: Why is this needed ?

niranjan94: Since you're anyway only using track name in the view.

niranjan94: Just a groupBy track.name should give you all the info you need+sessions

app/controllers/public/sessions.js
```diff
@@ -0,0 +1,33 @@
...
+
+const { Controller, computed } = Ember;
+
+export default Controller.extend({
+  sessions: [{ shortAbstract: 'An introduction to the event', id: '1', title: 'Welcome to FOSSASIA', startAt: new Date(), endAt: new Date(), track: { color: 'green', fontColor: 'green', id: 1, name: 'Track 1' }, microlocation: { id: 2, name: 'Room 1' }, speakers: [{ shortBiography: 'Works for ORG 1', id: 0, city: 'Delhi', name: 'Arnold Singh', speakingExperience: 'GSOC 2015', organisation: 'ORG 1', longBiography: '', photo: { 'iconImageUrl': '' } }] },
+  { shortAbstract: 'Welcome to open source', id: '2', title: 'Introduction to Open source', startAt: new Date(), endAt: new Date(), track: { color: 'red', fontColor: 'white', id: 2, name: 'Track 2' }, microlocation: { id: 3, name: 'Room 5' }, speakers: [{ shortBiography: '', id: 0, city: 'Delhi', name: 'Tux', speakingExperience: 'Speaker at the high conference', organisation: 'ORG 2', longBiography: '', photo: { 'iconImageUrl': '' } }] }
+  ],
+  sessionsByTracks: computed('sessions.[]', 'selectedTrackId', function() {
+    var allSessions = chain(this.get('sessions'))
+      .orderBy(['startAt', 'startAt'])
```
niranjan94: Duplicate entry

app/controllers/public/sessions.js
geekyd: I'm using the id to map the tracks and not the names anymore. That's the reason we are using this function to map the track id with the sessions.

niranjan94: @geekyd I'm sorry .. I don't understand.

niranjan94: You're performing unnecessary operations here. With what I suggested here > https://github.com/fossasia/open-event-frontend/pull/68#discussion_r117942196 you can perform one operation less and yet achieve the same results.

niranjan94: For UI grouping, track name shouldn't be an issue.

geekyd: We pass tracks to the filter component which stores `selectedTrackId` the id of the selected track. We need to update the session-filter according to this information, therefore I have used track id and not tracks name to populate and bind the sessions under a track id to the id in session-filter. 

Now whenever `selectedTrackId` changes it filters the sessions of session-list according to the trackid.

niranjan94: @geekyd that's exactly what is being done here too https://github.com/fossasia/open-event-frontend/pull/68#discussion_r117942196 but with one less operation. 

With one or two sessions, a single less operation isn't gonna matter. But with hundreds of session, a single operation, especially ones like `mapValues` matter a lot

geekyd:
@niranjan94 Removed `mapValues` please check.

app/templates/components/public/session-list.hbs
```diff
@@ -0,0 +1,8 @@
+{{#each-in sessions as |id trackSession|}}
```
niranjan94: call `id` as `trackName`

niranjan94: Now. When it is filtered by a track, what is going to come here is an array and not an object. Please handle that too. 

TL,DR; 
The session-list component should be able to render both an array of sessions and an object (when it is grouped by track) 

You can use the `is-array` helper available

geekyd: There is no need to handle this as output of both filter and unfiltered data is always going to be an object and not an array.
The format:
`{ track1: Array, track2: Array ...}`

niranjan94: Crap. Did not notice that groupBy in the filtered output. We should remove that. No point grouping by track when we are filtering by track.

geekyd:
@niranjan94 Made the changes. Please review.

app/templates/components/public/session-list.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="ui session-container">
+  {{#if (is-array sessions)}}
+    <h3 class="ui header" id="{{underscore track}}">{{track}}</h3>
+    {{#each sessions as |session|}}
+      {{public/session-item session=session}}
+    {{/each}}
+  {{else}}
+    {{#each-in sessions as |track trackSession|}}
+      <h3 class="ui header" id="{{underscore track}}">{{track}}</h3>
```
niranjan94: The ID can be removed

app/templates/components/public/session-list.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="ui session-container">
+  {{#if (is-array sessions)}}
+    <h3 class="ui header" id="{{underscore track}}">{{track}}</h3>
```
niranjan94: The ID can be removed

app/components/public/session-item.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  classNames: ['ui', 'accordion', 'segment']
```
niranjan94: The `accordion` class can be removed here. This will fix the extra space on top.

screen shot 2017-05-23 at 11 59 49 pm

app/templates/components/public/session-item.hbs
```diff
@@ -0,0 +1,45 @@
+{{#ui-accordion}}
+  <div class="title">
+    <div class="ui">
+      <h4 class="ui header">{{session.title}}</h4>
+      <div class="ui row">Talk</div>
```
niranjan94: This should be the session type.

niranjan94: And add this as a subheader to your header for a better UI 

```hbs
<h4 class="ui header">
    {{session.title}}
    <div class="sub header">
        {{session.sessionType.name}}
    </div>
</h4>

File: app/templates/components/public/session-item.hbs [hunk 21]
```diff
@@ -0,0 +1,45 @@
...
+    </div>
+    <br>
+    <div class="ui grid">
+      <div class="column twelve wide">
```
niranjan94: Let the speaker image, microlocation name and start time be in one row and the session shortAbstract in a different row. That would fix the sessionAbstract speaker image spacing issue

screen shot 2017-05-24 at 12 06 41 am

app/templates/components/public/session-item.hbs
niranjan94: Also, increase from `h4` to `h3`

geekyd:
@niranjan94 Please review.

app/templates/components/public/session-item.hbs
```diff
@@ -0,0 +1,52 @@
...
+          {{/each}}
+        </div>
+      </div>
+      <div class="ui right aligned column four wide" style="padding-left: 0">
+        <div class="">{{session.microlocation.name}} <i class="marker icon"></i></div>
+        <div class="">{{moment-format session.startAt 'hh:mm a / DD-MM-YYYY'}} <i class="wait icon"></i></div>
+      </div>
+    </div>
+    <div class="ui row">
+      <div class="session-description">
```
niranjan94: There should be a column inside the row. So put this inside a div with class `column`

geekyd:
@niranjan94 Made the changes. Please review.

app/templates/components/public/session-item.hbs
```diff
@@ -0,0 +1,52 @@
...
+            <img alt="speaker" class="ui mini avatar image" src="{{if speaker.photo.iconImageUrl speaker.image '/images/placeholders/avatar.png'}}">
+          {{/each}}
+        </div>
+      </div>
+      <div class="ui right aligned column four wide" style="padding-left: 0">
+        <div class="">{{session.microlocation.name}} <i class="marker icon"></i></div>
+        <div class="">{{moment-format session.startAt 'hh:mm a / DD-MM-YYYY'}} <i class="wait icon"></i></div>
+      </div>
+    </div>
+    <div class="ui row">
```
niranjan94: @geekyd

Looks like the row is out side the grid ? 

Should be something like this.

```hbs
<div class="ui grid">
    <div class="row">
        <!-- speaker pics and microlocation/time -->
    </div>
<div class="row">
    <div class="column">
        {{session.shortAbstract}}
    </div>
</div>
```

A `row` cannot exist outside a `ui grid`

geekyd:
@niranjan94 Please review.

PR #266, opened 2017-06-18
Fixes same id assertion error on admin/permissions

### LINKED ISSUE #265: Visiting admin/permissions/event-roles route results in an assertion error in console
- [x] bug report

**Current behavior:**

Visiting admin/permissions/event-roles route results in an assertion error in console, this is being caused because same id being is assigned to multiple elements on that route.
Also, the names are not meaningful for the ids as numbers have been assigned.
image

**Expected behaviour:**

Ids/names should be distinct.

**Steps to reproduce:**

Visit `admin/permissions/event-roles` on `localhost` and check the console to see the assertion error.

CosmicCoder96:
#### Checklist

visiting `admin/permissions/event-roles` resulted in a same id assertion error because of same id being used for multiple checkboxes.

- Adds the n-times component.
- make the id's (changed to names for better form handling) unique
- make their name more logical, similar to the ones used in orga server 

Fixes #265

### DIFF
--- a/app/templates/admin/permissions/event-roles.hbs
+++ b/app/templates/admin/permissions/event-roles.hbs
@@ -17,20 +17,20 @@
     </tr>
   </thead>
   <tbody>
-    {{#each model.roles as |item|}}
+    {{#each model.roles as |role|}}
       <tr>
-        <td>{{item}}</td>
-        {{#each (range 1 6) as |number|}}
+        <td>{{role}}</td>
+        {{#n-times times=5}}
           <td>
-            {{ui-checkbox label=(t 'Create') id=(concat 'create_' number)}}
+            {{ui-checkbox label=(t 'Create')}}
             <br>
-            {{ui-checkbox label=(t 'Read') id=(concat 'read_' number)}}
+            {{ui-checkbox label=(t 'Read')}}
             <br>
-            {{ui-checkbox label=(t 'Update') id=(concat 'update_' number)}}
+            {{ui-checkbox label=(t 'Update')}}
             <br>
-            {{ui-checkbox label=(t 'Delete') id=(concat 'delete_' number)}}
+            {{ui-checkbox label=(t 'Delete')}}
           </td>
-        {{/each}}
+        {{/n-times}}
       </tr>
     {{/each}}
   </tbody>
--- a/app/templates/components/n-times.hbs
+++ b/app/templates/components/n-times.hbs
@@ -1,0 +1,3 @@
+{{#each (range 0 times) as |number|}}
+  {{yield number}}
+{{/each}}
--- a/app/components/n-times.js
+++ b/app/components/n-times.js
@@ -1,0 +1,7 @@
+import Ember from 'ember';
+
+const { Component } = Ember;
+
+export default Component.extend({
+  tagName: ''
+});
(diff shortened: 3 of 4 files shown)


CosmicCoder96:
Deployment link : https://cc96-opev-fe-omega.herokuapp.com/
@geekyd @niranjan94 Please review

CosmicCoder96:
@niranjan94 Won't we be requiring names here while saving the data into database?

niranjan94:
Nope. Not needed. The inputs will be binded to a js object. We won't be using id/name references. We need it only for validation.

CosmicCoder96:
@niranjan94 got it, thanks. By inputs being binded to js object you mean, ember will take care of it via the it assigns by default when we use ember input helpers ??

app/templates/admin/permissions/event-roles.hbs
```diff
@@ -17,18 +17,19 @@
     </tr>
   </thead>
   <tbody>
-    {{#each model.roles as |item|}}
+    {{#each model.roles as |role|}}
       <tr>
-        <td>{{item}}</td>
-        {{#each (range 1 6) as |number|}}
+        <td>{{role}}</td>
+        {{#each (range 1 6) as |index|}}
```
CosmicCoder96: @niranjan94 Is it okay, if I make a helper to simply iterate `n` times like `{{#times 4}}` as the syntax {{#each range(1 6)}} is deprecated, and defining `|index|`and not using it results in template lint error as well, I think the helper would be quite handy if ever need to simply loop too, Please let me know.

niranjan94: @CosmicCoder96 please understand the difference b/w a component and a helper :) ... You're talking about a component now.

niranjan94: Ahh .. okay. My bad. Go ahead.

geekyd: @CosmicCoder96 @niranjan94 I think it would me more appropriate if we close this PR, and do this in a separate issue. What do you think?

CosmicCoder96: @geekyd No it's fine, no need to close this one, as it will fix the issue of assertion error, which was the original objective, I will place the component for loop in a separate commit for clarity :)

CosmicCoder96:
@niranjan94 Turns out block helpers are no more supported, so have made use of a component, please review

app/templates/components/n-times.hbs
```diff
@@ -0,0 +1,3 @@
+{{#each numOfTimes as |time|}}
```
niranjan94: Instead of using a computed property, can you use `{{#each (range 1 times) as |number|}}`

CosmicCoder96:
@niranjan94 Done, Please review

CosmicCoder96:
@niranjan94 done , please review

app/templates/components/n-times.hbs
```diff
@@ -0,0 +1,3 @@
+{{#each (range 1 (inc times)) as |number|}}
```
niranjan94: Instead of incrementing the number, you could start the range from 0 right ? One less operation.

CosmicCoder96: @niranjan94  of course, apologies. Done!

CosmicCoder96:
@niranjan94 Have made the changes

PR #296, opened 2017-06-20
Migrate translation assets

niranjan94:
#### Checklist

- Migrate translation files from old frontend
- Change `i18n` to `l10n` to match with the service ( @fossasia/open-event-frontend please take note of this change )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
   - bower install
 
 script:
-  - ember l10n:convert -f false
+  - node scripts/l10n.js generate
   - COVERAGE=true ember exam --random
 
 after_success:
--- a/app/services/l10n.js
+++ b/app/services/l10n.js
@@ -1,19 +1,54 @@
 import computed from 'ember-computed';
 import L10n from 'ember-l10n/services/l10n';
-// import l10nFingerprintMap from './../utils/l10n-fingerprint-map';
+import l10nFingerprintMap from './../utils/l10n-fingerprint-map';
 
 export default L10n.extend({
 
   availableLocales: computed(function() {
     return {
-      'en': this.t('en')
+      'bn'      : this.t('Bengali/Bangla'),
+      'zh_Hans' : this.t('Chinese (Simplified)'),
+      'zh_Hant' : this.t('Chinese (Traditional)'),
+      'en'      : this.t('English'),
+      'fr'      : this.t('French'),
+      'de'      : this.t('German'),
+      'id'      : this.t('Indonesian'),
+      'ko'      : this.t('Korean'),
+      'pl'      : this.t('Polish'),
+      'es'      : this.t('Spanish'),
+      'th'      : this.t('Thai'),
+      'vi'      : this.t('Vietnamese'),
+      'hi'      : this.t('Hindi'),
+      'ja'      : this.t('Japanese'),
+      'ru'      : this.t('Russian')
     };
   }),
 
-  autoInitialize: true,
+  localStorageKey: 'current_locale',
+
+  autoInitialize: false,
 
   jsonPath: '/assets/locales',
 
   // Make this return null if you do not want to use fingerprinting
-  fingerprintMap: null
+  fingerprintMap: l10nFingerprintMap,
+
+  switchLanguage(locale) {
+    this.setLocale(locale);
+    localStorage.setItem(this.localStorageKey, locale);
+    location.reload();
+  },
+
+  init() {
+    this._super(...arguments);
+    const currentLocale = localStorage.getItem(this.localStorageKey);
+    const detectedLocale = this.detectLocale();
+    if (currentLocale) {
+      this.setLocale(currentLocale);
+    } else if (detectedLocale) {
+      this.setLocale(detectedLocale);
+    } else {
+      this.setLocale('en');
+    }
+  }
 });
--- a/scripts/l10n.js
+++ b/scripts/l10n.js
@@ -1,0 +1,62 @@
+/* eslint-disable no-console */
+const cli = require('ember-cli');
+const fs = require('fs');
+const async = require('async');
+
+/**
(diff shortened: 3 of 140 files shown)


translations/zh_Hant.po
```diff
@@ -0,0 +1,3854 @@
...
+#: app/templates/components/forms/wizard/sessions-speakers-step.l10n.hbs:1
+msgid "Private Link"
+msgstr ""
+
+#: app/templates/components/forms/wizard/sessions-speakers-step.l10n.hbs:1
+msgid "Copy"
+msgstr "複製"
+
+#: app/templates/components/forms/wizard/sessions-speakers-step.l10n.hbs:1
+msgid "S
```
CosmicCoder96: @niranjan94 This is the first time I am looking at translation files so a bit confused, why are some of the `msgstr` randomly empty

app/services/l10n.js
```diff
@@ -1,19 +1,54 @@
...
+      'ru'      : this.t('Russian')
     };
   }),
 
-  autoInitialize: true,
+  localStorageKey: 'current_locale',
+
+  autoInitialize: false,
 
   jsonPath: '/assets/locales',
```
CosmicCoder96: @niranjan94 Not sure if it is related to this PR, but there is an error in the console at this file
image

geekyd: This error exists in the current version also.

CosmicCoder96: @geekyd I am aware there is an error, however it is not the same error as this one, it changed and hence I thought it might be worth discussing
image

This is the other error, only appears on localhost, not heroku deployment

geekyd: @CosmicCoder96 Yeah, this error exists even after generating the locales.

.travis.yml
```diff
@@ -34,7 +34,7 @@ install:
   - bower install
 
 script:
-  - ember l10n:convert -f false
+  - node scripts/l10n.js generate
```
geekyd: @niranjan94 While executing the script it throws error. Also even after generating the JSON the app still gives 404 error in the get requests.
screenshot from 2017-06-21 03-38-45

app/services/l10n.js
geekyd: This error occurs due to incorrect path in the get request `assets/locales/bn.json `. The actual path is something like `assets/locales/1497996247744/bn.json` due to incorrect path we get error 404 in console.

translations/zh_Hant.po
niranjan94: Because they are not translated yet

niranjan94:
> On changing to some other language, it is still using the automatic detection. Is it out of the scope of this PR?

No. Not at all. Should be done in this PR. Will do. Thanks.

niranjan94:
@geekyd @CosmicCoder96 the errors have been resolved. Please review again.

niranjan94:
Please note that this won't be working for heroku deployments since we don't store the generated translation assets on the repo and heroku deployment at this point doesn't support autogeneration of the translation assets.

niranjan94:
@geekyd that error is unavoidable. Because both the translation service and something else might try to change a text during language change. That's why we do a `location.reload()` after change to clear everything up and re-render the entire page again in a new language.

geekyd:
@niranjan94 Okay what about this error it appears when we select `Hindi`

niranjan94:
@geekyd this too. Probably a view is being re-rendered by two things at the same time. Have faced the same issue before while using translation libraries. Won't be an issue as we anyway trigger a full page reload. (To experiment, try commenting out the page reload statement and change language. You'll actually be able to _see_ which views are failing (they wouldn't have been translated) 😉 )

geekyd:
@niranjan94 I checked both the frontend as well as the orga-server the followung languages are not working for both of the projects
- Polish
- French
- Thai
- Russian

Is there any specific reason for that?

niranjan94:
It's not that they're not working. They are still not translated.

CosmicCoder96:
@niranjan94 Will there be a separate issue for generating json files for all languages or should we take care of it in this. Because other than the console error everything looks fine, I tested it by generating some languages.

niranjan94:
@CosmicCoder96 the json files will not be added to the repo. This PR takes care of generating them before deployment to nextgen.eventyay.com

geekyd:
@niranjan94 One last thing. What do we do about the languages which are not yet translated? Are we going to add it later or what?

scripts/l10n.js
```diff
@@ -0,0 +1,62 @@
...
+
+function mergeUpdatedTranslation() {
+  async.eachSeries(translationFiles, function(fileName, callback) {
+    runEmberCommand('l10n:extract', '-g', '-l', fileName.split('.')[0]).then(() => {
+      callback();
+    }).catch(console.error);
+  });
+}
+
+function generateTranslationJsonFiles() {
```
CosmicCoder96: @niranjan94 While deploying on eventyay this script will run, will this function take care of the missing translations which @geekyd pointed out?

niranjan94:
@CosmicCoder96 @geekyd there seems to be some mis-understanding here. lemme explain completely :)

- Every language this project will be supporting will have a file inside `translations`
- Each of those languages would be in different stages of translation. Some 70%. Some 0%. Some 100%. 
- During deployment all of these language files will be converted to the required `json` format by `generateTranslationJsonFiles` function. (Irrespective of what stage their translation is in). 
- The language options which we want to give to the user will be defined at `app/services/l10n.js`. This list maybe be lesser than the number of languages in `translations` folder. For now, we have added all the languages inside `translation` folder there too. Just for test purposes. But in future, we'll have only those languages that are translated about 75%.

Hope everything is clear now.

geekyd:
@niranjan94 Got it! Thanks for explaining :)

geekyd:
@niranjan94 Please remove the conflict so that we can merge this. :)

PR #300, opened 2017-06-21
Adds confirm modal before deleting event

### LINKED ISSUE #299: Create delete modal for events
- [ ] bug report
- [x] feature request
image

When the user clicks on delete button on events/view route, the functional confirm dialog should popup.

CosmicCoder96:
#### Checklist

Adds the confirm delete modal before deleting an event on event overview page.

image
image

Fixes #299

### DIFF
--- a/app/components/modals/event-delete-modal.js
+++ b/app/components/modals/event-delete-modal.js
@@ -1,0 +1,12 @@
+import Ember from 'ember';
+import ModalBase from 'open-event-frontend/components/modals/modal-base';
+
+const { computed } = Ember;
+
+export default ModalBase.extend({
+  isSmall         : true,
+  confirmName     : '',
+  isNameDifferent : computed('confirmName', function() {
+    return this.get('confirmName').toLowerCase() !== this.get('eventName').toLowerCase();
+  })
+});
--- a/app/controllers/events/view.js
+++ b/app/controllers/events/view.js
@@ -1,0 +1,15 @@
+import Ember from 'ember';
+
+const { Controller } = Ember;
+
+export default Controller.extend({
+  actions: {
+    openDeleteEventModal() {
+      this.set('isEventDeleteModalOpen', true);
+    },
+
+    deleteEvent() {
+      this.set('isEventDeleteModalOpen', false);
+    }
+  }
+});
--- a/app/templates/components/modals/event-delete-modal.hbs
+++ b/app/templates/components/modals/event-delete-modal.hbs
@@ -1,0 +1,24 @@
+<div class="header">
+  {{t 'Are you sure you would like to delete this event?'}}
+  <div class="muted small text">
+    {{t 'Deleting the event will delete all the data associated with it.'}}
+  </div>
+</div>
+<div class="content">
+  <form class="ui form" autocomplete="off" {{action (optional formSubmit) on='submit' preventDefault=true}}>
+    <div class="field">
+      <div class="label">
+        {{t 'Please enter the full name of the event to continue'}}
+      </div>
+      {{input type='text' name='confirm_name' value=confirmName required=true}}
+    </div>
+  </form>
+</div>
+<div class="actions">
+  <button type="button" class="ui black button" {{action 'close'}}>
+    {{t 'Cancel'}}
+  </button>
+  <button type="submit" class="ui red button" disabled={{isNameDifferent}} {{action deleteEvent}}>
+    {{t 'Delete Event'}}
+  </button>
+</div>
--- a/tests/integration/components/modals/event-delete-modal-test.js
+++ b/tests/integration/components/modals/event-delete-modal-test.js
@@ -1,0 +1,13 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('modals/event-delete-modal', 'Integration | Component | modals/event delete modal');
+
+test('it renders', function(assert) {
+  this.set('isOpen', false);
+  this.set('eventName', 'sample');
+  this.set('deleteEvent', () => {});
+  this.render(hbs`{{modals/event-delete-modal isOpen=isOpen eventName=eventName deleteEvent=(action deleteEvent)}}`);
+  assert.ok(this.$().html().trim().includes('Are you sure'));
+});
--- a/app/templates/events/view.hbs
+++ b/app/templates/events/view.hbs
@@ -2,7 +2,9 @@
   <div class="ui grid stackable">
     <div class="row">
       <div class="four wide column">
-        <h2 class="ui header">Fossasia Test</h2>
+        <h2 class="ui header">
+          {{model.name}}
+        </h2>
       </div>
       <div class="twelve wide column {{unless device.isMobile 'right aligned'}}">
         {{#if device.isMobile}}
@@ -10,7 +12,7 @@
             <button class="ui button"><i class="unhide icon"></i></button>
             <button class="ui button"><i class="check icon"></i></button>
             <button class="ui button"><i class="copy icon"></i></button>
-            <button class="ui red button"><i class="trash icon"></i></button>
+            <button class="ui red button" {{action 'openDeleteEventModal'}}><i class="trash icon"></i></button>
           </div>
         {{else}}
             <button class="ui button labeled icon small">
@@ -27,7 +29,7 @@
                 {{t 'Copy'}}
               </button>
             </div>
-            <button class="ui red button labeled icon small">
+            <button class="ui red button labeled icon small" {{action 'openDeleteEventModal'}}>
               <i class="trash icon"></i>
               {{t 'Delete'}}
             </button>
@@ -58,7 +60,9 @@
       {{/tabbed-navigation}}
     </div>
   </div>
-  
+
+  {{modals/event-delete-modal isOpen=isEventDeleteModalOpen eventName=model.name deleteEvent=(action 'deleteEvent')}}
+
   <div class="ui segment">
     {{outlet}}
   </div>
--- a/tests/unit/controllers/events/view-test.js
+++ b/tests/unit/controllers/events/view-test.js
@@ -1,0 +1,9 @@
+import { test } from 'ember-qunit';
+import moduleFor from 'open-event-frontend/tests/helpers/unit-helper';
+
+moduleFor('controller:events/view', 'Unit | Controller | events/view');
+
+test('it exists', function(assert) {
+  let controller = this.subject();
+  assert.ok(controller);
+});


CosmicCoder96:
Deployment link : https://cc96-opev-frontend.herokuapp.com/events/963d3ba5
@niranjan94 I have gone with both functionalities here regarding the delete button, i.e. the typical form validation and have also disabled the button until the correct name is entered like on the orga server, please let me know which one to keep, or keep both of them

niranjan94:
No need the validation. Just disable button until name is correct. Also, make the check case insensitive.

CosmicCoder96:
@niranjan94 Done, please review again, deployment is updated

app/components/modals/event-delete-modal.js
```diff
@@ -0,0 +1,15 @@
+import ModalBase from 'open-event-frontend/components/modals/modal-base';
+import FormMixin from 'open-event-frontend/mixins/form';
+
+export default ModalBase.extend(FormMixin, {
+  isSmall : true,
+  options : {
+    closable: false
```
niranjan94: this can be closable.

app/templates/components/modals/event-delete-modal.hbs
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+      {{input type='text' name='confirm_name' value=confirmName}}
+    </div>
+  </form>
+</div>
+<div class="actions">
+  <button type="button" class="ui black button" {{action 'close'}}>
+    {{t 'Cancel'}}
+  </button>
+  <button type="submit" class="ui red button" disabled={{not-eq (lowercase eventName) (lowercase confirmName)}} {{action 'deleteEvent'}}>
```
niranjan94: Move this logic to a computed property. Let's keep only basic conditional checks on the template.

CosmicCoder96:
@niranjan94 have moved it to a computed property.

app/components/modals/event-delete-modal.js
CosmicCoder96: done

CosmicCoder96:
@sumedh123 Please review

app/components/modals/event-delete-modal.js
```diff
@@ -0,0 +1,20 @@
...
+  isSmall: true,
+
+  actions: {
+    deleteEvent() {
+      this.close();
+    }
+  },
+
+  doesNameMatch: computed('confirmName', function() {
+    return lowerCase(this.get('confirmName')) !== lowerCase(this.get('eventName'));
```
niranjan94: Instead of importing and using this, you could use `String.prototype.toLowerCase()` ? Similar performance.

app/components/modals/event-delete-modal.js
```diff
@@ -0,0 +1,20 @@
...
+import FormMixin from 'open-event-frontend/mixins/form';
+import { lowerCase } from 'lodash';
+
+const { computed } = Ember;
+
+export default ModalBase.extend(FormMixin, {
+  isSmall: true,
+
+  actions: {
+    deleteEvent() {
```
niranjan94: Bubble this action up to the controller. To be handled by the controller. 

Lets follow data-down, action-up approach

CosmicCoder96: @niranjan94 Something like this 
```
sendDelete() {
      this.sendAction('deleteEvent');
      this.close();
    }
```

niranjan94:
No. Don't have the action in the component at all. Directly call an action
from the controller.

app/components/modals/event-delete-modal.js
CosmicCoder96: @niranjan94 This is done

app/components/modals/event-delete-modal.js
CosmicCoder96: @niranjan94 Using this requires me to typecast to string, is it still worth it?

app/controllers/events/view.js
```diff
@@ -0,0 +1,15 @@
...
+
+const { Controller } = Ember;
+
+export default Controller.extend({
+  actions: {
+    openDeleteEventModal() {
+      this.set('eventDeleteModalIsOpen', true);
+    },
+
+    deleteEvent(modal) {
```
niranjan94: Don't pass model reference here please. Just set `eventDeleteModalIsOpen` as false.

app/templates/components/modals/event-delete-modal.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="header">
+  {{t 'Are you sure you would like to delete this event?'}}
+  <div class="muted small text">
+    {{t 'Deleting the event will delete all the data associated with it.'}}
+  </div>
+</div>
+<div class="content">
+  <form class="ui form" autocomplete="off" {{action 'deleteEvent' on='submit' preventDefault=true}}>
```
niranjan94: This action should also refer to the `deleteEvent` action that is being passed to the component.

app/components/modals/event-delete-modal.js
niranjan94: Why do you need to type case to string ? Both `confirmName` and `eventName` are strings right ?

CosmicCoder96:
@niranjan94 I am about to push some code, which will **fail** tests, to discuss something

app/templates/components/modals/event-delete-modal.hbs
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+      {{input type='text' name='confirm_name' value=confirmName}}
+    </div>
+  </form>
+</div>
+<div class="actions">
+  <button type="button" class="ui black button" {{action 'close'}}>
+    {{t 'Cancel'}}
+  </button>
+  <button type="submit" class="ui red button" disabled={{doesNameMatch}}  {{action deleteEvent}}>
```
CosmicCoder96: @niranjan94 This works, but fails tests as it doesn't use a quoted string for action, what is the correct alternative?

niranjan94: No. This is valid and shouldn't fail tests.

sumedh123:
@CosmicCoder96 The one with the delete button disabled untill correct name is good to go with.

app/templates/components/modals/event-delete-modal.hbs
CosmicCoder96: image

CosmicCoder96: @niranjan94 This seems to be arising from this itself ^^

niranjan94: Yes. You need to update the test for it. You need to pass in an action in the integration test too.

tests/integration/components/modals/event-delete-modal-test.js
```diff
@@ -0,0 +1,12 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('modals/event-delete-modal', 'Integration | Component | modals/event delete modal');
+
+test('it renders', function(assert) {
+  this.set('isOpen', false);
+  this.set('eventName', 'sample');
+  this.render(hbs`{{modals/event-delete-modal isOpen=isOpen eventName=eventName}}`);
```
niranjan94: An action needs to be passed here.

app/templates/components/modals/event-delete-modal.hbs
CosmicCoder96: Oh, okay then I misunderstood the error log, updating the test. And what about the

CosmicCoder96:
@niranjan94 Please review, have made the changes

tests/integration/components/modals/event-delete-modal-test.js
```diff
@@ -0,0 +1,13 @@
+import { test } from 'ember-qunit';
+import moduleForComponent from 'open-event-frontend/tests/helpers/component-helper';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('modals/event-delete-modal', 'Integration | Component | modals/event delete modal');
+
+test('it renders', function(assert) {
+  this.set('isOpen', false);
+  this.set('eventName', 'sample');
+  this.set('deleteEvent', 'deleteEvent');
```
niranjan94: A string is not an action.
See:
https://guides.emberjs.com/v2.13.0/testing/testing-components/#toc_testing-actions

app/components/modals/event-delete-modal.js
```diff
@@ -0,0 +1,14 @@
+import Ember from 'ember';
+import ModalBase from 'open-event-frontend/components/modals/modal-base';
+import FormMixin from 'open-event-frontend/mixins/form';
```
niranjan94: Form mixin not needed since we don't do any validations here.

app/templates/components/modals/event-delete-modal.hbs
```diff
@@ -0,0 +1,24 @@
...
+    {{t 'Deleting the event will delete all the data associated with it.'}}
+  </div>
+</div>
+<div class="content">
+  <form class="ui form" autocomplete="off" {{action deleteEvent on='submit' preventDefault=true}}>
+    <div class="field">
+      <div class="label">
+        {{t 'Please enter the full name of the event to continue'}}
+      </div>
+      {{input type='text' name='confirm_name' value=confirmName}}
```
niranjan94: add `required=true`. Let the browser handle this for us.

app/templates/components/modals/event-delete-modal.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="header">
+  {{t 'Are you sure you would like to delete this event?'}}
+  <div class="muted small text">
+    {{t 'Deleting the event will delete all the data associated with it.'}}
+  </div>
+</div>
+<div class="content">
+  <form class="ui form" autocomplete="off" {{action deleteEvent on='submit' preventDefault=true}}>
```
niranjan94: Instead of `deleteEvent`, point this to a dummy action. Let the user press the button to delete.

tests/integration/components/modals/event-delete-modal-test.js
CosmicCoder96: @niranjan94 Since the action is not doing anything right now, will the test action be empty?

niranjan94: Yep

app/templates/components/modals/event-delete-modal.hbs
CosmicCoder96: @niranjan94  dummy action in the component or the bubbled up to controller?

niranjan94: Just putting 

`{{action (optional dummy) on='submit' preventDefault=true}}` should work

niranjan94: You don't have to define any action.

CosmicCoder96:
@niranjan94 Done with the changes, please review

app/templates/components/modals/event-delete-modal.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="header">
+  {{t 'Are you sure you would like to delete this event?'}}
+  <div class="muted small text">
+    {{t 'Deleting the event will delete all the data associated with it.'}}
+  </div>
+</div>
+<div class="content">
+  <form class="ui form" autocomplete="off" {{action 'submit' on='submit' preventDefault=true}}>
```
niranjan94: this is not changed yet ?

CosmicCoder96: 'submit' is not defined anywhere, so it is dummy right? I have removed deleteEvent action from here

niranjan94: No. This would throw an error now if the user submits the form by pressing the enter key when typing into the input field. 

That's why do what I said 

`{{action (option anyRandomVariableThatMayNotExist on='submit' preventDefault=true}}`

CosmicCoder96: Ah! `optional` is a an ember composable helper, I though you were just calling the action name as 'optional name' . Apologies.

CosmicCoder96:
@niranjan94 Done. Please review

app/templates/components/modals/event-delete-modal.hbs
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+      {{input type='text' name='confirm_name' value=confirmName required=true}}
+    </div>
+  </form>
+</div>
+<div class="actions">
+  <button type="button" class="ui black button" {{action 'close'}}>
+    {{t 'Cancel'}}
+  </button>
+  <button type="submit" class="ui red button" disabled={{doesNameMatch}}  {{action deleteEvent}}>
```
geekyd: Please remove extra space after `disabled={{doesNameMatch}}`.

app/controllers/events/view.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
+
+const { Controller } = Ember;
+
+export default Controller.extend({
+  actions: {
+    openDeleteEventModal() {
+      this.set('eventDeleteModalIsOpen', true);
```
sumedh123: 'isEventDeleteModalOpen' would be better?

CosmicCoder96:
@niranjan94 @sumedh123 @geekyd made the changes, please review

app/components/modals/event-delete-modal.js
```diff
@@ -0,0 +1,13 @@
+import Ember from 'ember';
+import ModalBase from 'open-event-frontend/components/modals/modal-base';
+
+const { computed } = Ember;
+
+export default ModalBase.extend({
+  isSmall       : true,
+  confirmName   : '',
+  eventName     : '',
+  doesNameMatch : computed('confirmName', function() {
```
niranjan94: The variable name `doesNameMatch` is confusing. Currently, if `doesNameMatch=true` then that actually means they don't match.

CosmicCoder96: @niranjan94 How about `isNameDifferent` :) ?

niranjan94: Cool 👍

CosmicCoder96:
@niranjan94 Done 👍

app/components/modals/event-delete-modal.js
```diff
@@ -0,0 +1,13 @@
+import Ember from 'ember';
+import ModalBase from 'open-event-frontend/components/modals/modal-base';
+
+const { computed } = Ember;
+
+export default ModalBase.extend({
+  isSmall         : true,
+  confirmName     : '',
+  eventName       : '',
```
geekyd: Is this required? We are passing the `eventName` to the component itself.

CosmicCoder96:
@geekyd Have removed it, thanks for pointing that out I removed it.

PR #480, opened 2017-07-15
Fix enter key not submitting forms bug

### LINKED ISSUE #476: 'Enter' key not submitting forms
- [x] bug report
- [ ] feature request

**Current behaviour:**

Nothing happens on pressing the enter key, i.e. submit action of forms is not triggered after the user presses the enter key.
**Expected behaviour:**

Form's submit action should execute on pressing the key
**Steps to reproduce:**

Can be reproduced on login/registration form.

CosmicCoder96:
#### Checklist

Fixes the bug where enter key didn't submit the form.

Modify form mixin
Reference for changes : https://stackoverflow.com/questions/11811426/why-does-the-tagname-dom-property-return-an-uppercase-value

Fixes #476

### DIFF
--- a/app/mixins/form.js
+++ b/app/mixins/form.js
@@ -54,7 +54,7 @@ export default Mixin.create({
       }
 
       let $form;
-      if (this.get('tagName') === 'form' || (this.$() && this.$().prop('tagName'))) {
+      if ((this.get('tagName') && this.get('tagName').toLowerCase() === 'form') || (this.$() && this.$().prop('tagName').toLowerCase() === 'form')) {
         $form = this.$();
         $form.addClass('ui form');
       } else {


app/mixins/form.js
```diff
@@ -54,7 +54,7 @@ export default Mixin.create({
       }
 
       let $form;
-      if (this.get('tagName') === 'form' || (this.$() && this.$().prop('tagName'))) {
+      if (this.get('tagName') === 'FORM' || (this.$() && this.$().prop('tagName') === 'FORM')) {
```
niranjan94: Convert to lower case and compare with lowercase.

niranjan94: Because the first one `this.get('tagName')` could be set by the user and can be either in lower or upper.

CosmicCoder96:
@niranjan94 Done !

Issue #714, opened 2017-08-13
Download feature not implemented


mariobehling:
The download feature is not working. Seems to be just a dummy.

![screenshot from 2017-08-13 21-23-56](https://user-images.githubusercontent.com/1583873/29252774-c694f000-806d-11e7-8c57-6279238f00ae.png)

CosmicCoder96:
Closing via https://github.com/fossasia/open-event-frontend/pull/717

PR #717, opened 2017-08-13
feat: Integrates event export api at event/export route.

### LINKED ISSUE #657: Implement Event Export API at 'events/<event_id/export>'
- [ ] bug report
- [x] feature request

Todo:
export
Implement event export api at the export route.

sumedh123:
#### Checklist

Fixes #657 , #714

### DIFF
--- a/app/controllers/events/view/export.js
+++ b/app/controllers/events/view/export.js
@@ -1,0 +1,56 @@
+import Ember from 'ember';
+
+const { Controller, run } = Ember;
+export default Controller.extend({
+  eventDownloadUrl   : '',
+  eventExportStatus  : 'Event export not yet started.',
+  isDownloadDisabled : true,
+  data               : {
+    image    : false,
+    audio    : false,
+    video    : false,
+    document : false
+  },
+  requestLoop(exportJobInfo) {
+    run.later(() => {
+      this.get('loader')
+        .load(exportJobInfo.task_url, { withoutPrefix: true })
+        .then(exportJobStatus => {
+          if (exportJobStatus.state === 'SUCCESS') {
+            this.set('isLoading', false);
+            this.set('isDownloadDisabled', false);
+            this.set('eventDownloadUrl', exportJobStatus.result.download_url);
+            this.set('eventExportStatus', exportJobStatus.state);
+            this.get('notify').success(this.l10n.t('Event exported.'));
+          } else if (exportJobStatus.state === 'WAITING') {
+            this.requestLoop(exportJobInfo);
+            this.set('eventExportStatus', exportJobStatus.state);
+            this.get('notify').alert(this.l10n.t('Event export is going on.'));
+          } else {
+            this.set('isLoading', false);
+            this.set('eventExportStatus', exportJobStatus.state);
+            this.get('notify').error(this.l10n.t('Event export failed.'));
+          }
+        })
+        .catch(() => {
+          this.set('isLoading', false);
+          this.set('eventExportStatus', 'FAILURE');
+          this.get('notify').error(this.l10n.t('Event export failed.'));
+        });
+    }, 3000);
+  },
+  actions: {
+    startGeneration() {
+      this.set('isLoading', true);
+      let payload = this.get('data');
+      this.get('loader')
+        .post(`/events/${this.get('model.id')}/export/json`, payload)
+        .then(exportJobInfo => {
+          this.requestLoop(exportJobInfo);
+        })
+        .catch(() => {
+          this.get('notify').error(this.l10n.t('Unexpected error occurred.'));
+        });
+    }
+  }
+});
--- a/app/services/loader.js
+++ b/app/services/loader.js
@@ -13,7 +13,8 @@ export default Service.extend({
     if (config.isExternal) {
       options.url = url;
     } else {
-      options.url = config.withoutPrefix ? `${adapter.host}/${url}` : `${adapter.urlPrefix()}${url}`;
+      url = `${url[0] !== '/' ? '/' :  ''}${url}`;
+      options.url = config.withoutPrefix ? `${adapter.host}${url}` : `${adapter.urlPrefix()}${url}`;
     }
 
     if (data) {
--- a/app/templates/components/events/view/export/download-zip.hbs
+++ b/app/templates/components/events/view/export/download-zip.hbs
@@ -8,25 +8,27 @@
 </h3>
 <div class="ui hidden divider"></div>
 <div class="ui container"></div>
-<div class="ui form">
+<div class="ui form {{if isLoading 'loading'}}">
   <div class="field">
-    {{ui-checkbox class='toggle' label=(t 'Image')}}
+    {{ui-checkbox class='toggle' label=(t 'Image') checked=data.image onChange=(action (mut data.image))}}
   </div>
   <div class="field">
-    {{ui-checkbox class='toggle' label=(t 'Video')}}
+    {{ui-checkbox class='toggle' label=(t 'Video') checked=data.video onChange=(action (mut data.video))}}
   </div>
   <div class="field">
-    {{ui-checkbox class='toggle' label=(t 'Audio')}}
+    {{ui-checkbox class='toggle' label=(t 'Audio') checked=data.audio onChange=(action (mut data.audio))}}
   </div>
   <div class="field">
-    {{ui-checkbox class='toggle' label=(t 'Document')}}
+    {{ui-checkbox class='toggle' label=(t 'Document') checked=data.document onChange=(action (mut data.document))}}
   </div>
   <div class="ui basic segment less left padding">
-    <button class="ui blue button">
+    <button class="ui blue button" {{action startGeneration}}>
       {{t 'Start'}}
     </button>
-    <button class="ui button">
+    <a href="{{eventDownloadUrl}}" class="ui button {{if isDownloadDisabled 'disabled' ''}}">
       {{t 'Download'}}
-    </button>
+    </a>
+    <div class="hidden divider"></div>
+    <h4 class="ui header">Status: {{eventExportStatus}}</h4>
   </div>
 </div>
--- a/app/templates/events/view/export.hbs
+++ b/app/templates/events/view/export.hbs
@@ -1,6 +1,6 @@
 <div class="ui stackable very relaxed container grid">
   <div class="eight wide column">
-    {{events/view/export/download-zip}}
+    {{events/view/export/download-zip startGeneration=(action 'startGeneration') isDownloadDisabled=isDownloadDisabled eventExportStatus=eventExportStatus eventDownloadUrl=eventDownloadUrl data=data isLoading=isLoading}}
     <div class="ui divider"></div>
     {{events/view/export/download-common downloadType='Pentabarf XML' downloadUrl='pentabarf.xml'}}
     <div class="ui divider"></div>
--- a/tests/unit/controllers/events/view/export-test.js
+++ b/tests/unit/controllers/events/view/export-test.js
@@ -1,0 +1,10 @@
+import { test } from 'ember-qunit';
+import moduleFor from 'open-event-frontend/tests/helpers/unit-helper';
+
+moduleFor('controller:events/view/export', 'Unit | Controller | events/view/export', []);
+
+// Replace this with your real tests.
+test('it exists', function(assert) {
+  let controller = this.subject();
+  assert.ok(controller);
+});
--- a/tests/integration/components/events/view/export/download-zip-test.js
+++ b/tests/integration/components/events/view/export/download-zip-test.js
@@ -4,7 +4,11 @@ import hbs from 'htmlbars-inline-precompile';
 
 moduleForComponent('events/view/export/download-zip', 'Integration | Component | events/view/export/download zip');
 
+
 test('it renders', function(assert) {
-  this.render(hbs`{{events/view/export/download-zip l10n=l10n}}`);
+  this.on('startGeneration', function() { });
+  this.set('eventExportStatus', '');
+  this.set('eventDownloadUrl', '');
+  this.render(hbs`{{events/view/export/download-zip startGeneration=(action 'startGeneration') eventExportStatus=eventExportStatus eventDownloadUrl=eventDownloadUrl l10n=l10n}}`);
   assert.ok(this.$().html().trim().includes('Zip'));
 });
--- a/tests/unit/serializers/user-test.js
+++ b/tests/unit/serializers/user-test.js
@@ -5,6 +5,7 @@ moduleForModel('user', 'Unit | Serializer | user', {
     'serializer:user',
     'serializer:event',
     'model:event',
+    'model:session',
     'model:email-notification',
     'model:event-invoice',
     'model:notification',


tests/unit/controllers/events/view/export-test.js
```diff
@@ -0,0 +1,12 @@
+import { moduleFor, test } from 'ember-qunit';
```
geekyd: Use the helper for `modelFor`

tests/unit/controllers/events/view/export-test.js
```diff
@@ -0,0 +1,12 @@
...
+
+moduleFor('controller:events/view/export', 'Unit | Controller | events/view/export', {
```
geekyd: Make this 
```
moduleFor('controller:events/view/export', 'Unit | Controller | events/view/export', [])
```

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@fossasia/open-event-frontend Please review.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,45 @@
...
+        video    : this.get('model.exportData.dataVideo'),
+        audio    : this.get('model.exportData.dataAudio'),
+        document : this.get('model.exportData.dataDocument')
+      };
+      let exportUrl = `/events/${this.get('model').currentEventID}/export/json`;
+      let that = this;
+      this.get('loader')
+        .post(exportUrl, payload)
+        .then(data => {
+          let requestLoop = setInterval(function() {
```
niranjan94: Don't use `setInterval`. Doesn't play well with Ember's run loop. Use `Ember.run.later` instead.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,45 @@
...
+  actions           : {
+    startGeneration() {
+      let payload = {
+        image    : this.get('model.exportData.dataImage'),
+        video    : this.get('model.exportData.dataVideo'),
+        audio    : this.get('model.exportData.dataAudio'),
+        document : this.get('model.exportData.dataDocument')
+      };
+      let exportUrl = `/events/${this.get('model').currentEventID}/export/json`;
+      let that = this;
```
niranjan94: Don't do this. Use [arrow functions](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions) instead.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,45 @@
...
+      this.get('loader')
+        .post(exportUrl, payload)
+        .then(data => {
+          let requestLoop = setInterval(function() {
+            that.get('loader')
+              .load(`${data.task_url.substring(3)}`)
```
niranjan94: Don't use substrings for getting the path. Parse the URL properly and implicitly define which part of the URL to use. What does the value of `data.task_url` look like now ?

app/components/events/view/export/download-zip.js
```diff
@@ -3,4 +3,9 @@ import Ember from 'ember';
 const { Component } = Ember;
 
 export default Component.extend({
+  actions: {
+    startGeneration() {
```
niranjan94: Why this unnecessary intermediate action ? Remove it and use the passed in action directly,

app/controllers/events/view/export.js
sumedh123: @niranjan94 /v1/tasks/7025f8be-7569-42a7-9b2f-549bd0f21fc3

app/components/events/view/export/download-zip.js
sumedh123: @niranjan94 Do you want me to move all the logic from my controller to the component itself?

niranjan94: No ... Remove this action here in the component and in your component hbs file, instead of `{{action 'startGeneration'}}` just do `{{action startGeneration}}`

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,49 @@
...
+        image    : this.get('model.exportData.dataImage'),
+        video    : this.get('model.exportData.dataVideo'),
+        audio    : this.get('model.exportData.dataAudio'),
+        document : this.get('model.exportData.dataDocument')
+      };
+      let exportUrl = `/events/${this.get('model').currentEventID}/export/json`;
+      this.get('loader')
+        .post(exportUrl, payload)
+        .then(data => {
+          let download_url = data.task_url.split('/').splice(2, 4).join('/');
```
niranjan94: I have updated the loader service's config handling to be a bit more sane (in https://github.com/fossasia/open-event-frontend/commit/39c9bae55e50052bd655bc70000c68d328e49959) . Now just pass in a config object with `withoutPrefix: true` and you can pass in the task_url url directly. Without having to play with the string in anyway.

niranjan94: Its never advisable to play with the url like this. Split and splice and join is more error prone now than the previous sub string implementation.

sumedh123:
@niranjan94 The tests fail because we are using the ```{{action startGeneration}}```
``` Error: Assertion Failed: You specified a quoteless path, `startGeneration`, to the {{action}} helper which did not resolve to an action name (a string). Perhaps you meant to use a quoted actionName? (e.g. {{action "startGeneration"}}).```

File: app/templates/events/view/export.hbs [hunk 8]
```diff
@@ -1,6 +1,6 @@
 <div class="ui stackable very relaxed container grid">
   <div class="eight wide column">
-    {{events/view/export/download-zip}}
+    {{events/view/export/download-zip data=model startGeneration=(action 'startGeneration') eventExportStatus=eventExportStatus eventDownloadUrl=eventDownloadUrl}}
```
niranjan94: But you are passing in an action as a variable here ... `startGeneration`. So you will be able to refer to it within the `events/view/export/download-zip` component as `{{action startGeneration}}` directly just like 

- https://github.com/fossasia/open-event-frontend/blob/fbd8ddf1e5c579feb85441aadc5f63b4e7c5ee93/app/templates/components/widgets/forms/ticket-input.hbs#L27
- https://github.com/fossasia/open-event-frontend/blob/4676a0406e6dfda3bc79c656ae60cf0b7ba2854c/app/templates/components/event-card.hbs#L32

sumedh123: @niranjan94 I have done the same here as you told.

niranjan94: @sumedh123 You need to update your tests too. How can you expect the test to work if you don't pass in an action ?

sumedh123: Ohh sorry missed that. My bad.

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@niranjan94 Please review.

geekyd:
@sumedh123 The URL for GET request is incorrect
```
https://open-event-api.herokuapp.com//v1/tasks/5b41166d-04ec-4525-ab8a-a4649893cc63
```

Shouldn't it be 
```
https://open-event-api.herokuapp.com/v1/tasks/5b41166d-04ec-4525-ab8a-a4649893cc63
```

sumedh123:
@niranjan94 I did the changes after you made changes to the makePromise method and added the:
```
{ withourPrefix: true }
```
So see @geekyd 's comment above. It think it adds '/'.

niranjan94:
@sumedh123 then please make changes to the loader to handling a leading slash properly.

niranjan94:
@sumedh123 right now we have 

```js
      options.url = config.withoutPrefix ? `${adapter.host}/${url}` : `${adapter.urlPrefix()}${url}`;
```
Ensure a slash is added only if `url` does not have a slash at the start.

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@geekyd @niranjan94 Please review.

app/services/loader.js
```diff
@@ -13,7 +13,7 @@ export default Service.extend({
     if (config.isExternal) {
       options.url = url;
     } else {
-      options.url = config.withoutPrefix ? `${adapter.host}/${url}` : `${adapter.urlPrefix()}${url}`;
+      options.url = config.withoutPrefix ? `${adapter.host}${url}` : `${adapter.urlPrefix()}${url}`;
```
niranjan94: @sumedh123 this is not what I said... I said add a slash when there is no leading slash. And don't add a slash if there is a leading slash. Do the check here and do it.

sumedh123: @niranjan94 I tried doing that in first place, but consider a case in which we pass ```withoutPrefix: false``` but having leading url. In that case, it will remove the leading url and then the above code will be executed i.e ````${adapter.urlPrefix()}${url}``` which will make the url wrong as is happening in this case for the post request that we do to get the 'task_url'.

niranjan94: @sumedh123 forget withoutPrefix for a second. Irrespective of if it is true or false, add a leading slash to the `url` value if it doesn't have one already.

sumedh123: @niranjan94 But the thing is ```${adapter.host}${url}``` and ```${adapter.urlPrefix()}${url}``` give different output. If I do the check you told above, for the first post request which gives task_url(which also uses loader and has withoutPrefix set to null), has a url open-event-api.herokuapp.comevents/553fd33/export/json.

niranjan94: At the last? Leading slash means at the start. 

I think you're completely misunderstanding this. I'll get home and explain with a code sample.

sumedh123: @niranjan94 Before check:
event-exports

After performing a check like:
```
      if (url.split('/')[0] === '') {
        url = url.split('/').splice(1).join('/');
      }
```
event-exports4

niranjan94: I'll get home and explain.

niranjan94: @sumedh123 you're code 

```
      if (url.split('/')[0] === '') {
        url = url.split('/').splice(1).join('/');
      }
```

Itself is wrong. Did you try running it and seeing ? It is actually removing the leading slash. 

A very very simple code such as 

```js
url = `${url[0] !== '/' ? '/' :  ''}${url}`
```

will do the trick. Why do you have to split and splice and join and end up getting a wrong result ? In many cases, a simple answer is usually the best.

app/routes/events/view/export.js
```diff
@@ -1,9 +1,20 @@
...
-const { Route } = Ember;
+const { Route, RSVP, Object } = Ember;
 
 export default Route.extend({
   titleToken() {
     return this.l10n.t('Export');
+  },
+  model() {
+    return RSVP.hash({
+      exportData: Object.create({
```
niranjan94: This model is not needed at all. There is no data retrieval from server and undefined variables are _falsey_ by default in ember.

app/routes/events/view/export.js
```diff
@@ -1,9 +1,20 @@
 import Ember from 'ember';
 
-const { Route } = Ember;
+const { Route, RSVP, Object } = Ember;
 
 export default Route.extend({
   titleToken() {
     return this.l10n.t('Export');
+  },
+  model() {
```
niranjan94: Entire model definition is not needed here. the event data from the parent route `events.view` will be automatically accessible.

app/routes/events/view/export.js
niranjan94: So remove the model declaration. The variables can just exist at the controller level. You don't even have to define them anywhere.

app/routes/events/view/export.js
```diff
@@ -5,5 +5,8 @@ const { Route } = Ember;
...
+    return this.modelFor('events.view');
```
niranjan94: Not needed I guess. The parent model should be accessible if there is no model method defined here

sumedh123: @niranjan94 I want the event identifier in controller. That's why I am returning it here.

niranjan94: I understand that ... But as far as I can remember, that `this.modelFor('events.view')` will anyway be available if you don't define a model method in this route. 

Ember usually provides access to the model of the direct parent if no model method is defined. It can be accessed as usual at `model`

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,48 @@
...
+      let exportUrl = `/events/${this.get('model').id}/export/json`;
+      this.get('loader')
+        .post(exportUrl, payload)
+        .then(data => {
+          let requestLoop = run.later(() => {
+            this.get('loader')
+              .load(data.task_url, { withoutPrefix: true })
+              .then(zipDownloadUrl => {
+                if (zipDownloadUrl.state === 'SUCCESS') {
+                  $('#eventDownloadButton').removeClass('disabled');
```
niranjan94: Don't remove class like this. Use variable and bind the class with an if in the template

app/routes/events/view/export.js
niranjan94: Just give it a try. Remove the model method here. You should still be able to access the identifier at `model.identifier`

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,48 @@
+import Ember from 'ember';
+
+const { Controller, run, $ } = Ember;
+export default Controller.extend({
+  eventExportStatus : '',
+  eventDownloadUrl  : '',
+  actions           : {
+    startGeneration() {
+      let payload = {
+        image    : this.get('dataImage'),
```
niranjan94: use getProperties for this to get all at the same time.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,48 @@
...
+        video    : this.get('dataVideo'),
+        audio    : this.get('dataAudio'),
+        document : this.get('dataDocument')
+      };
+      let exportUrl = `/events/${this.get('model').id}/export/json`;
```
niranjan94: Why store it in a variable ? You can directly use it.

niranjan94: Instead of `this.get('model').id` do `this.get('model.id')`

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,48 @@
...
+      this.get('loader')
+        .post(exportUrl, payload)
+        .then(data => {
+          let requestLoop = run.later(() => {
```
niranjan94: Did you test this code ? Did you read the docs on run.later ? Run.later is analogous to `setTimeout`. It will run only once after a given delay. You should modify your code accordingly to make it work like `setInterval`

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,48 @@
...
+            this.get('loader')
+              .load(data.task_url, { withoutPrefix: true })
+              .then(zipDownloadUrl => {
+                if (zipDownloadUrl.state === 'SUCCESS') {
+                  $('#eventDownloadButton').removeClass('disabled');
+                  this.set('eventExportStatus', 'SUCCESS');
```
niranjan94: If you're anyway setting the state, you could just do a `this.set('eventExportStatus', zipDownloadUrl.state)` in a common place right ?

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,48 @@
...
+        document : this.get('dataDocument')
+      };
+      let exportUrl = `/events/${this.get('model').id}/export/json`;
+      this.get('loader')
+        .post(exportUrl, payload)
+        .then(data => {
+          let requestLoop = run.later(() => {
+            this.get('loader')
+              .load(data.task_url, { withoutPrefix: true })
+              .then(zipDownloadUrl => {
```
niranjan94: Call it `exportJobStatus` ... that makes more sense

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,48 @@
...
+      let payload = {
+        image    : this.get('dataImage'),
+        video    : this.get('dataVideo'),
+        audio    : this.get('dataAudio'),
+        document : this.get('dataDocument')
+      };
+      let exportUrl = `/events/${this.get('model').id}/export/json`;
+      this.get('loader')
+        .post(exportUrl, payload)
+        .then(data => {
```
niranjan94: call it exportJobInfo

app/templates/components/events/view/export/download-zip.hbs
```diff
@@ -10,23 +10,28 @@
 <div class="ui container"></div>
 <div class="ui form">
   <div class="field">
-    {{ui-checkbox class='toggle' label=(t 'Image')}}
+    {{ui-checkbox class='toggle' label=(t 'Image') checked=dataImage onChange=(action (mut dataImage))}}
```
niranjan94: How exactly are you passing in the variable `dataImage` and the rest to the component from the controller ?

niranjan94: I can't seem to find that anywhere

app/templates/components/events/view/export/download-zip.hbs
```diff
@@ -10,23 +10,28 @@
...
-    {{ui-checkbox class='toggle' label=(t 'Document')}}
+    {{ui-checkbox class='toggle' label=(t 'Document') checked=dataDocument onChange=(action (mut dataDocument))}}
   </div>
   <div class="ui basic segment less left padding">
-    <button class="ui blue button">
+    <button class="ui blue button" {{action startGeneration}}>
       {{t 'Start'}}
     </button>
-    <button class="ui button">
+    <a href="{{eventDownloadUrl}}" class="ui button disabled" id="eventDownloadButton">
```
niranjan94: Remove the id. Not needed

sumedh123:
@niranjan94 Will make the changes and push again. My bad. I should have tested the it locally.

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

app/templates/components/events/view/export/download-zip.hbs
sumedh123: @niranjan94 Should I include the dataImage, dataVideo, etc properties in controller? Since we are not keeping the model layer. Will that be okay?

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@niranjan94 Please review.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,53 @@
...
+  eventExportStatus  : '',
+  eventDownloadUrl   : '',
+  isDownloadDisabled : true,
+  image              : false,
+  audio              : false,
+  video              : false,
+  document           : false,
+  actions            : {
+    startGeneration() {
+      let payload = {
```
niranjan94: After making the change above, do

```js
let payload = this.get('data');
```

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,53 @@
...
+        image    : this.get('image'),
+        video    : this.get('video'),
+        audio    : this.get('audio'),
+        document : this.get('document')
+      };
+      this.get('loader')
+        .post(`/events/${this.get('model.id')}/export/json`, payload)
+        .then(exportJobInfo => {
+          let requestLoop = run.later(() => {
```
niranjan94: Again @sumedh123 ... this is running only _once_.

app/templates/components/events/view/export/download-zip.hbs
niranjan94: You need it pass it from the controller.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,53 @@
+import Ember from 'ember';
+
+const { Controller, run } = Ember;
+export default Controller.extend({
+  eventExportStatus  : '',
+  eventDownloadUrl   : '',
+  isDownloadDisabled : true,
+  image              : false,
+  audio              : false,
```
niranjan94: Instead of this, do 

```
data: {
  image: false,
  audio: false,
  video: false,
  document: false
}
```

Now you can just pass the data property into the component

app/templates/components/events/view/export/download-zip.hbs
niranjan94: See inline comments above.

app/templates/events/view/export.hbs
```diff
@@ -1,6 +1,6 @@
 <div class="ui stackable very relaxed container grid">
   <div class="eight wide column">
-    {{events/view/export/download-zip}}
+    {{events/view/export/download-zip startGeneration=(action 'startGeneration') isDownloadDisabled=isDownloadDisabled eventExportStatus=eventExportStatus eventDownloadUrl=eventDownloadUrl dataImage=image dataVideo=video dataAudio=audio dataDocument=document}}
```
niranjan94: DOn't pass everything like. See inline comments above and just pass data

app/controllers/events/view/export.js
sumedh123: @niranjan94 I have kept it in such a way that if the server response is WAITING, then it will keep running every 3 seconds else if the server returns direct failure I am stopping the loop. Also on SUCCESS response, I am terminating the loop.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,53 @@
...
+              .then(exportJobStatus => {
+                if (exportJobStatus.state === 'SUCCESS') {
+                  this.set('isDownloadDisabled', false);
+                  this.set('eventExportStatus', exportJobStatus.state);
+                  this.set('eventDownloadUrl', exportJobStatus.result.download_url);
+                  this.get('notify').success(this.l10n.t('Event exported.'));
+                  run.cancel(requestLoop);
+                } else if (exportJobStatus.state === 'WAITING') {
+                  this.set('eventExportStatus', exportJobStatus.state);
+                  requestLoop();
```
niranjan94: How can you call `requestLoop` like this ? afaik `requestLoop` is not a callable function. It just has a reference to the timer which you use for cancelling.

app/controllers/events/view/export.js
```diff
@@ -0,0 +1,53 @@
...
+                } else {
+                  this.set('eventExportStatus', exportJobStatus.state);
+                  run.cancel(requestLoop);
```
niranjan94: And this cancelling wouldn't be needed at at all. Because `run.later` is not a loop. It is to run a piece of code **once** after a given time delay. So when this code runs, there is actually nothing to cancel since it will not run again until you ask it to run.

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@niranjan94 Please review.

app/templates/components/events/view/export/download-zip.hbs
```diff
@@ -10,23 +10,28 @@
...
     </button>
-    <button class="ui button">
+    <a href="{{eventDownloadUrl}}" class="ui button {{if isDownloadDisabled 'disabled' ''}}">
       {{t 'Download'}}
-    </button>
+    </a>
+  </div>
+  <div class="ui basic segment">
+    <div class="ui header">
+      {{t 'Status'}} : {{eventExportStatus}}
```
geekyd: Is this required?

sumedh123: @geekyd I think it would be better to indicate the status to the user because in some cases the export might directly fail. Overall, we can show the status returned by API.

geekyd: We are using notify for these kind of nitifications, please use the notify service instead.

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@geekyd Did the changes. Please review. Sample event: Football Season.

app/templates/components/events/view/export/download-zip.hbs
niranjan94: We need to show the live status too. I'm sure the API returns a live status message when state is `WAITING`. Please check with server for statuses.

sumedh123: @niranjan94 Yes, it returns the states "WAITING", "FAILURE", "SUCCESS". Shall I show them up as I did previously but adding a loader will not show the waiting state since it will blur the area. It will show the success and the failure state since we are setting the loader to false when we have SUCCESS or FAILURE. Should I indicate the state too?

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@geekyd @niranjan94 Please review. I have added the status and the loader.

geekyd:
@sumedh123 Looks good to me. Please squash the commits.

niranjan94:
Deployment link: https://pr-717-open-event-frontend-fossasia.surge.sh

sumedh123:
@geekyd Done :+1:

sumedh123:
@geekyd @CosmicCoder96 The celery task fail issue: https://github.com/fossasia/open-event-orga-server/issues/4392 has been opened on the server.

sumedh123:
@geekyd @CosmicCoder96 I checked locally. I can download the events now. I guess server has resolved the issue.

PR #722, opened 2017-08-14
Making external event url field editable

### LINKED ISSUE #683: Create Event Wizzard: Not possible to add link in first field
It is not possible to add a link into the event wizzard.

screenshot from 2017-08-13 15-41-49

harshitagupta30:
#### Checklist

Currently, while we're creating the form we can not add external link in the form. This PR makes the external URL link field editable making sure that the this can be only added and edited not removed. The same pattern is followed in eventyay.

Fixes #683 
@fossasia/open-event-frontend Please review.


niranjan94:
Deployment link: https://pr-722-open-event-frontend-fossasia.surge.sh

app/templates/components/forms/wizard/basic-details-step.hbs
```diff
@@ -118,7 +118,7 @@
     <label>{{t 'Links & Social Media'}}</label>
     {{widgets/forms/link-input
       hasLinkName=true
-      fixedName=true
+      fixedName=false
```
niranjan94: What has to be done here is, instead of making the change here, 

The `disabled=fixedName` from https://github.com/fossasia/open-event-frontend/blob/development/app/templates/components/widgets/forms/link-input.hbs#L15 needs to goto https://github.com/fossasia/open-event-frontend/blob/development/app/templates/components/widgets/forms/link-input.hbs#L3

niranjan94: Looks like there was a mistake in the template itself.

harshitagupta30: @niranjan94 I have made changes as you suggested. But wanted to clear out we need to disabled this field? Because with the above piece I checked it is not disabling even when `fixedName=true` and if never want to disable then I guess we can remove the field.

niranjan94:
Deployment link: https://pr-722-open-event-frontend-fossasia.surge.sh

app/templates/components/forms/wizard/basic-details-step.hbs
niranjan94: @harshitagupta30 i do not understand what you're telling. The code is fine. Checked the deployment link too. Works as expected. 

So, which exactly are you saying as not working ?

PR #782, opened 2017-08-23
Misalignment in tabbed navigation for dynamic routing resolved

### LINKED ISSUE #775: Tabbed navigation getting misaligned on dynamic routing
- [x] bug report
- [ ] feature request

**Current behavior:** 
When navigating from one tab to another, the tabbed navigation gets misaligned. 
image

**Expected behavior:**
It should stay at its position.

harshitagupta30:
#### Checklist

This PR resolves the misalignment in tabbed navigation for dynamic routing in admin routes.

Fixes #775

### DIFF
--- a/app/components/tabbed-navigation.js
+++ b/app/components/tabbed-navigation.js
@@ -8,9 +8,7 @@ export default Component.extend({
   currentRoute: computed('session.currentRouteName', 'item', function() {
     var path = this.get('session.currentRouteName');
     if (path) {
-      this.set('item', this.$('a.active'));
-      this.$('a').addClass('vertical-item');
-      return this.$('a.active').text().trim();
+      return this.get('item');
     }
   }),
   didInsertElement() {
@@ -18,15 +16,21 @@ export default Component.extend({
     if (isMobile) {
       this.$('a').addClass('vertical-item');
     }
-    this.set('item', this.$('a.active'));
+    this.set('item', this.$('a.active').text().trim());
+  },
+  didUpdate() {
+    var isMobile = this.get('device.isMobile');
+    if (isMobile) {
+      this.$('a').addClass('vertical-item');
+    }
   },
   actions: {
     toggleMenu(mode) {
       var menu = this.$('div.menu');
       menu.toggleClass('hidden');
-
       if (mode === 'reset') {
-        this.set('item', this.$('a.active'));
+        this.$('a.item').addClass('vertical-item');
+        this.set('item', event.srcElement.text);
       }
     }
   }


niranjan94:
Deployment link: https://pr-782-open-event-frontend-fossasia.surge.sh

app/templates/components/tabbed-navigation.hbs
```diff
@@ -1,8 +1,8 @@
 {{#if device.isMobile}}
-  <div role="button" class="ui segment center aligned" {{action 'toggleMenu'}}>
+  <div role="button" class="ui segment center aligned" onClick={{action 'toggleMenu'}}>
```
niranjan94: Why is `onClick=` needed ? An action by default binds to the click event.

niranjan94:
Deployment link: https://pr-782-open-event-frontend-fossasia.surge.sh

app/templates/components/tabbed-navigation.hbs
harshitagupta30: @niranjan94 I understand that an action by default binds to the click event but if I remove onClick then the value of the is not getting updated though the problem of misalignment is resolved. If we navigate then the value remains same that is why I have explicitly called `onClick` here.

niranjan94:
Deployment link: https://pr-782-open-event-frontend-fossasia.surge.sh

app/templates/components/tabbed-navigation.hbs
niranjan94: Why is the value not getting updated ? `onClick={{action 'toggleMenu'}}` and `{{action 'toggleMenu'}}` are basically the same internally. Please investigate what is causing any issue here.

niranjan94: We should never shoot in the dark. We need to know what what is happening the way it is happening.

harshitagupta30: @niranjan94 The value is not getting updated on simply using action because we're using the jquery object `event` in `tabbed-navigation.js` which we can only pass via `onClick` method. So if I will remove the `onClick` the jquery object will not be passed and we won't be able to set it. Also, I am using the jquery object because the action lies on the dropdown and not individual items so for that we need the event which has the info about what was clicked which I am fetching using `event.srcElement.text`.

niranjan94: No. `event` is passed in even when using a normal action. `event` will directly be accessible within the action method. (Don't declare an argument)

harshitagupta30: @niranjan94 ok! Passing argument was causing the issue. I have removed it and made changes. Now it's working as expected. Can you please review again? Thanks.

niranjan94:
Deployment link: https://pr-782-open-event-frontend-fossasia.surge.sh

PR #1093, opened 2018-05-04
fix: Update ember models table from v1 to v2

### LINKED ISSUE #1092: Update ember-models-table to 2.x
Ref: https://onechiporenko.github.io/ember-models-table/v.2/#/examples

mayank8318:
#### Checklist

Update ember-models-table from v1 to v2

- Update dependency version of ember-models-table

Fixes #1092

### DIFF
--- a/app/components/ui-table.js
+++ b/app/components/ui-table.js
@@ -1,90 +1,9 @@
-import O, { observer, getWithDefault, set } from '@ember/object';
 import TableComponent from 'ember-models-table/components/models-table';
 import layout from 'open-event-frontend/templates/components/ui-table';
-
-const defaultMessages = {
-  searchLabel            : 'Search:',
-  searchPlaceholder      : 'Search',
-  columnsTitle           : 'Columns',
-  columnsShowAll         : 'Show All',
-  columnsHideAll         : 'Hide All',
-  columnsRestoreDefaults : 'Restore Defaults',
-  tableSummary           : 'Show %@ - %@ of %@',
-  allColumnsAreHidden    : 'All columns are hidden. Use <strong>columns</strong>-dropdown to show some of them',
-  noDataToShow           : 'No records to show'
-};
-
-const defaultIcons = {
-  sortAsc         : 'caret down icon',
-  sortDesc        : 'caret up icon',
-  columnVisible   : 'checkmark box icon',
-  columnHidden    : 'square outline icon',
-  navFirst        : 'chevron left icon',
-  navPrev         : 'chevron left icon',
-  navNext         : 'chevron right icon',
-  navLast         : 'chevron right icon',
-  caret           : 'caret',
-  expandRow       : 'plus icon',
-  expandAllRows   : 'plus icon',
-  collapseRow     : 'minus icon',
-  collapseAllRows : 'minus icon',
-  selectAllRows   : 'checkmark box icon',
-  deselectAllRows : 'square outline icon',
-  selectRow       : 'checkmark box icon',
-  deselectRow     : 'square outline icon'
-};
-
-const assign = Object.assign || assign;
+import Semantic from 'open-event-frontend/themes/semantic';
 
 export default TableComponent.extend({
   layout,
 
-  _setupMessages: observer('customMessages', function() {
-    const customIcons = getWithDefault(this, 'customMessages', {});
-    let newMessages = {};
-    assign(newMessages, defaultMessages, customIcons);
-    set(this, 'messages', O.create(newMessages));
-  }),
-
-  _setupIcons() {
-    const customIcons = getWithDefault(this, 'customIcons', {});
-    let newIcons = {};
-    assign(newIcons, defaultIcons, customIcons);
-    set(this, 'icons', O.create(newIcons));
-  },
-
-  simplePaginationTemplate: 'components/ui-table/simple-pagination',
-
-  numericPaginationTemplate: 'components/ui-table/numeric-pagination',
-
-  tableFooterTemplate: 'components/ui-table/table-footer',
-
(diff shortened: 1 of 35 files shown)


mayank8318:
@niranjan94 @geekyd @CosmicCoder96 I think this test is failing for everyone :
screenshot from 2018-05-05 10-01-15

package.json
```diff
@@ -77,7 +77,7 @@
     "ember-math-helpers": "^2.4.1",
     "ember-maybe-import-regenerator": "^0.1.6",
     "ember-metrics": "^0.12.1",
-    "ember-models-table": "^1.13.0",
+    "ember-models-table": "2",
```
niranjan94: Specify a proper range `^2.5.0`

mayank8318:
@niranjan94 I think I missed something because I noticed that a certain test related to `ui-table-select` is failing saying that instead of Ember.Component, Ember.Object is received. I will go through the docs again to see how to do the migration properly and update the PR accordingly.

niranjan94:
@mayank8318 yes please. Always **Test** it locally before making a PR. By test I do not mean just running the test suite, but actually run the application and try out a few basic tasks. 

Just going to `/events/live` would have shown you that the tables fail to render. Please keep the above in mind :)

mayank8318:
Yep I will definitely keep that in mind. Sorry :)

mayank8318:
@niranjan94 @geekyd The table is being rendered but not correctly. From what I could find out, it doesn't have the appropriate class tags :
(Left is the old v1 implementation     Right is the new v2 implementation)
screenshot from 2018-05-07 14-43-42

As is clear some class tags like `ui ui-table` etc are missing. I think this is happening because the v2 version doesn't have `innerTableWrapper` and `outerTableWrapper` which we are overriding for the Semactic UI theme. As a result the table looks like this - 
screenshot from 2018-05-07 15-30-58

I couldn't find any resources for this problem. Please help me here :)

mayank8318:
I had also asked this question in the ember-models-table repository and it has been answered with possible solutions. https://github.com/onechiporenko/ember-models-table/issues/313
@niranjan94 @geekyd I think the second solution is a good enough workaround in our case and we can find a concrete solution later on. Please let me know what you think of this

mayank8318:
The problem seems to arise as the `classes` object is empty here. So the CSS classes are not applied. 
screenshot from 2018-05-08 12-27-35

If I replace the use of `classes` with direct CSS in the `ui-table.hbs` it works but it means we need to hardcode it. Should I go ahead with it?

niranjan94:
@mayank8318 if I remembered right, the v2 version provides semantic UI support. Can't we use that directly instead of attempting to customize ?

mayank8318:
@niranjan94 I have made the necessary changes. Please review :)

mayank8318:
I think I will have to pass themeInstance through all the tables since it is passed in the controller of the table. Should I go ahead with it?

niranjan94:
@mayank8318 dont' pass theme instance in each of the controllers/templates. Just add the themeInstance in the main ui-table component itself.

mayank8318:
@niranjan94 I have made the changes. Please review https://deploy-preview-1093--open-event-frontend.netlify.com/

app/components/ui-table.js
```diff
@@ -1,130 +1,12 @@
...
-
-  rowTemplate: 'components/ui-table/row',
-
-  expandedRowTemplate: 'components/ui-table/expanded-row',
-
-  noDataShowTemplate: 'components/ui-table/no-data',
-
-  allColumnsHiddenTemplate: 'components/ui-table/all-columns-hidden'
-
+  themeInstance: computed(function() {
```
niranjan94: I don't think this needs to be a computed property. `themeInstance: Semantic.create()` should do.

app/controllers/events/list.js
```diff
@@ -1,27 +1,34 @@
...
     {
       propertyName : 'starts-at',
-      template     : 'components/ui-table/cell/cell-date',
+      component    : 'components/ui-table/cell/cell-date',
       title        : 'Date'
     },
     {
       propertyName     : 'roles',
-      template         : 'components/ui-table/cell/cell-roles',
+      component        : 'components/ui-table/cell/cell-roles',
```
niranjan94: Isn't just `ui-table/cell/cell-roles` working ?

app/themes/semantic.js
```diff
@@ -0,0 +1,86 @@
+import Default from 'ember-models-table/themes/semanticui';
+
+
+export default Default.extend({
+  components: {
+    'pagination-simple'  : 'components/ui-table/simple-pagination',
+    numericPagination    : 'components/ui-table/numeric-pagination',
```
niranjan94: missing quotes here ?

app/themes/semantic.js
```diff
@@ -0,0 +1,86 @@
...
+    'table-footer'       : 'components/ui-table/table-footer',
```
niranjan94: For all of this, doesn't something like `ui-table/table-footer` work ? (Without the `components/` part)

mayank8318: I didn't check. Will do it now

mayank8318: No doesn't work

app/controllers/events/list.js
mayank8318: Nope

app/themes/semantic.js
niranjan94: Okay cool! Thanks for the update @mayank8318 :)

mayank8318:
Ping @niranjan94 @CosmicCoder96 @ritikamotwani Please test and review :)

mayank8318:
@geekyd Please review now :)

PR #1159, opened 2018-05-22
Add option to add event type and event topic to admin panel

### LINKED ISSUE #1153: Allow admin to add event type, event topic and event sub topic.
Currently while creating event there are no event types and topics available. Currently there is no way where admin can add an event topic or sub topic.

Admin should be able to add event type and topics. So that they are available to choose when we create event.

**Describe alternatives you've considered**
We can have a pre-filled data in Database. Not sure if this is correct way. And how this is implemented in v1.

pradeepgangwar:
#### Checklist

Currently there is not option to add event types and topic for admins. This causes problem while creating events as we are not able to select types and topics of events. This PR fixes this.

Fixes #1153

### DIFF
--- a/app/components/modals/admin/content/new-event-sub-topic-modal.js
+++ b/app/components/modals/admin/content/new-event-sub-topic-modal.js
@@ -1,0 +1,33 @@
+import FormMixin from 'open-event-frontend/mixins/form';
+import ModalBase from 'open-event-frontend/components/modals/modal-base';
+
+export default ModalBase.extend(FormMixin, {
+  isSmall            : true,
+  autoScrollToErrors : false,
+
+  actions: {
+    addEventProperty(modelInstance) {
+      this.onValid(() => {
+        this.sendAction('addEventProperty', modelInstance);
+      });
+    }
+  },
+  getValidationRules() {
+    return {
+      inline : true,
+      delay  : false,
+      on     : 'blur',
+      fields : {
+        subTopicName: {
+          identifier : 'sub_topic_name',
+          rules      : [
+            {
+              type   : 'empty',
+              prompt : this.get('l10n').t('Please enter a sub topic name')
+            }
+          ]
+        }
+      }
+    };
+  }
+});
--- a/app/controllers/admin/content/events.js
+++ b/app/controllers/admin/content/events.js
@@ -1,0 +1,52 @@
+import Controller from '@ember/controller';
+import { camelCase, startCase } from 'lodash';
+
+export default Controller.extend({
+  actions: {
+    async loadSubTopics(topic) {
+      this.set('isLoading', true);
+      try {
+        this.set('subTopics', await topic.get('subTopics'));
+      } catch (e) {
+        this.notify.error(this.get('l10n').t('An unexpected error has occurred. SubTopics not loaded.'));
+      } finally {
+        this.set('isLoading', false);
+      }
+    },
+    openModalFor(modelInstanceOrName) {
+      const modelName = typeof modelInstanceOrName === 'string' ? modelInstanceOrName : modelInstanceOrName.constructor.modelName;
+      const modelInstance = typeof modelInstanceOrName === 'string' ? this.store.createRecord(modelInstanceOrName) : modelInstanceOrName;
+      const camelCasedValue = camelCase(modelName);
+      this.set(camelCasedValue, modelInstance);
+      modelInstance.openModal = true;
+    },
+    deleteEventProperty(eventProp) {
+      this.set('isLoading', true);
+      eventProp.destroyRecord()
+        .then(() => {
+          this.notify.success(this.get('l10n').t('This Event Property has been deleted successfully.'));
+        })
+        .catch(()=> {
+          this.notify.error(this.get('l10n').t('An unexpected error has occurred. Event Type was not deleted.'));
+        })
+        .finally(() => {
+          this.set('isLoading', false);
+        });
+    },
+    addEventProperty(modelInstance) {
+      const camelCasedValue = camelCase(modelInstance.constructor.modelName);
+      this.set('isLoading', true);
+      modelInstance.save()
+        .then(() => {
+          this.notify.success(this.get('l10n').t(`${startCase(camelCasedValue)} has been added successfully.`));
+        })
+        .catch(() => {
+          this.notify.error(this.get('l10n').t(`An unexpected error has occurred. ${startCase(camelCasedValue)} not saved.`));
+        })
+        .finally(() => {
+          this.set('isLoading', false);
+          this.set(camelCasedValue, null);
+        });
+    }
+  }
+});
--- a/app/router.js
+++ b/app/router.js
@@ -150,6 +150,7 @@ router.map(function() {
         this.route('list', { path: '/:topic_id' });
       });
       this.route('translations');
+      this.route('events');
     });
   });
   this.route('orders', function() {
--- a/app/templates/admin/content/events.hbs
+++ b/app/templates/admin/content/events.hbs
@@ -1,0 +1,157 @@
+<div class="ui stackable grid">
+  <div class="row">
+    <div class="five wide column">
+      <div class="ui center aligned fluid segment">
+        <div class="ui grid container">
+          <div class="middle aligned content eight wide column left aligned">
+            <h4 class="ui header">
+              {{t 'Event Types'}}
+            </h4>
+          </div>
+          <div class="eight wide column right aligned">
+            <button class="ui icon circular blue button {{if device.isMobile 'fluid'}}" {{action 'openModalFor' 'event-type'}}>
+              <i class="icon plus"></i>
+            </button>
+          </div>
+        </div>
+        <div class="ui divider"></div>
+        <table class="ui celled table">
+          <tbody>
+            {{#each model.eventTypes as |eventType|}}
+              <tr>
+                <td>
+                  <div class="ui grid">
+                    <div class="middle aligned content eight wide column left aligned">
+                      {{eventType.name}}
+                    </div>
+                    <div class="eight wide column right aligned">
+                      <div class="ui horizontal compact basic buttons">
+                        {{#ui-popup content=(t 'Edit') class='ui icon button' click=(action 'openModalFor' eventType)}}
+                          <i class="edit icon"></i>
+                        {{/ui-popup}}
+                        {{#ui-popup content=(t 'Delete') class='ui icon button' click=(action (confirm (t 'Are you sure you would like to delete this event type?') (action 'deleteEventProperty' eventType)))}}
+                          <i class="trash outline icon"></i>
+                        {{/ui-popup}}
+                      </div>
+                    </div>
+                  </div>
+                </td>
+              </tr>
+            {{/each}}
+          </tbody>
+        </table>
+      </div>
+    </div>
+    <div class="six wide column">
+      <div class="ui center aligned fluid segment">
+        <div class="ui grid container">
+          <div class="middle aligned content eight wide column left aligned">
+            <h4 class="ui left aligned header">
+              {{t 'Event Topics'}}
+            </h4>
+          </div>
+          <div class="eight wide column right aligned">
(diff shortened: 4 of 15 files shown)


pradeepgangwar:
Option to add subtopic is left. I am working further on that part.

niranjan94:
@pradeepgangwar is there backend support for this ? Are the APIs ready ?

app/router.js
```diff
@@ -151,6 +151,11 @@ router.map(function() {
       });
       this.route('translations');
     });
+    this.route('event-settings', function() {
```
niranjan94: Move these to content

```
/admin/content/events/topics
/admin/content/events/sub-topics
/admin/content/events/types
```

niranjan94: Or even better, you could put all three into just a single page too `/admin/content/events` since the functionality needed isn't complex.

app/routes/admin/event-settings/index.js
```diff
@@ -0,0 +1,14 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  titleToken() {
+    return this.get('l10n').t('Event Settings');
+  },
+  async model() {
+    return {
+      'eventTopics': await this.get('store').query('event-topic', {}),
```
niranjan94: Why quotes for the key ?

pradeepgangwar:
@niranjan94 Yes, APIs are ready, in fact I am able to create types and topics. But since event subtopic depends on event topic selected, so that needs a bit more implementation. I am working on that.

app/router.js
pradeepgangwar: Ok, I will try doing that. This means that form for creating all three things would be in same page? Currently I display them in same page but for new creation everything has different form.

pradeepgangwar: Thanks @niranjan94 I will fix this. :smile:

pradeepgangwar:
@niranjan94 I updated the code and now everything is in a single page with updates handled by modals. But I am having difficulty with subtopics. As subtopics depends on topic so when I select a topic from drop-down, A request is made to server and data is fetched but how do I update the dom to render the fetched data?
screenshot from 2018-05-24 01-14-58
screenshot from 2018-05-24 01-15-16

pradeepgangwar:
@niranjan94 Everything is working fine. Please review the code.
screenshot from 2018-05-24 01-14-58
screenshot from 2018-05-24 01-15-16

app/templates/admin/content/events.hbs
```diff
@@ -0,0 +1,124 @@
+<div class="ui stackable grid">
+  <div class="row">
+    <div class="five wide column">
+      <div class="ui center aligned fluid segment">
+        <h3 class="ui left aligned header">
```
niranjan94: Make these headers one level smaller. `h4`

app/controllers/admin/content/events.js
```diff
@@ -0,0 +1,133 @@
...
+        })
+        .catch(()=> {
+          this.notify.error(this.get('l10n').t('An unexpected error has occurred. Event Sub Topic not saved.'));
+        })
+        .finally(() => {
+          this.set('isLoading', false);
+        });
+    },
+    async fetchSubTopics(getEventTopic) {
+      return {
```
niranjan94: Why are you returning an object ? You can directly return subTopics right ?

app/controllers/admin/content/events.js
```diff
@@ -0,0 +1,133 @@
+import Controller from '@ember/controller';
+import { computed } from '@ember/object';
+import { later } from '@ember/runloop';
+
+export default Controller.extend({
+
+  subTopics: computed('model.topic', function() {
```
niranjan94: Remove this computed property. Since the subTopics is loaded asynchronously, use an action that gets triggered when the topic dropdown is changed to make a call to the subtopics and set the subtopic once the promise is resolved.

pradeepgangwar: Ok, I will try doing this. :)

app/controllers/admin/content/events.js
pradeepgangwar: I forgot to remove this code. This is not used right now. Sorry. :sweat_smile:

pradeepgangwar:
@niranjan94 Please check the new UI. And about removing that computed property, I got confused about how to send the event id (event subtopic need event ID for query) to action by simultaneously picking up selected id from dropdown and calling an action. 
I got this implementation from the event wizard where we have implemented this functionality like this. Let me know how this can be done.

screenshot from 2018-05-26 00-32-48

app/templates/admin/content/events.hbs
```diff
@@ -0,0 +1,157 @@
...
+          <div class="eight wide column right aligned">
+            <button class="ui icon circular blue button {{if device.isMobile 'fluid'}}" {{action 'openNewEventSubTopicModal'}}>
+              <i class="icon plus"></i>
+            </button>
+          </div>
+        </div>   
+        <div class="ui divider"></div>
+        <div class="field">
+          <label>{{t 'Event Topic'}}</label>
+          {{#ui-dropdown class='search selection' onChange=(action (mut model.topic)) as |execute mapper|}}
```
niranjan94: here instead of `(action (mut model.topic))` make it something like `(action 'loadSubTopic')`.

niranjan94: here instead of `(action (mut model.topic))` make it something like `(action 'loadSubTopic')`.

Then define an action like, 

```js
async loadSubTopic(topic) {
  this.set('subTopics', await topic.get('subTopics');
}
``` 

Make sure to change the implementation in the new subtopic modal too. Also, ensure you have a loading indicator that comes up when the subtopics are loading.

app/controllers/admin/content/events.js
```diff
@@ -0,0 +1,128 @@
...
+        this.set('subTopic', null);
+      } catch (ignored) { /* To suppress error thrown in-case this gets executed after component gets destroy */ }
+    }, 50);
+    if (!this.get('model.topic')) {
+      return [];
+    }
+    return this.get('model.topic.subTopics');
+  }),
+
+  actions: {
```
niranjan94: Too much repetition of very similar stuff. 

`openNewEventTypeModal`, `openNewEventTopicModal`, `openNewEventSubTopicModal` can become one action
`deleteEventType`, `deleteEventTopic`, `deleteEventSubTopic` can become one action
`addEventType`, `addEventTopic`, `addEventSubTopic` can become one action.

pradeepgangwar:
@niranjan94 I removed the computed property and added tests. However in the actions I was able to reduce delete type action but could not reduce add and openModal type actions. Those actions involve different calls and parameters each time. I tried different ways and one way was to add a key value pair in models of each event-type, event-topic and event-sub-topic that will identify which parameter was sent and then choose action. Although that didn't reduce code that much. Is that the correct way?

app/controllers/admin/content/events.js
```diff
@@ -0,0 +1,95 @@
...
+      this.set('isLoading', true);
+      try {
+        this.set('subTopics', await topic.get('subTopics'));
+      } catch (e) {
+        this.notify.error(this.get('l10n').t('An unexpected error has occurred. SubTopics not loaded.'));
+      } finally {
+        this.set('isLoading', false);
+      }
+    },
+    openNewEventTypeModal(eventType) {
```
niranjan94: The three can be,

```js
function openModalFor(modelInstanceOrName) {
  const modelName = typeof modelInstanceOrName === 'string' ? modelInstanceOrName : modelInstanceOrName.modelName;
  const modelInstance = typeof modelInstanceOrName === 'string' ? this.store.createRecord(modelInstanceOrName) : modelInstanceOrName;
  const camelCasedValue = camelCase(modelName); // from lodash
  this.set(camelCasedValue, modelInstance); // You can use this property itself to decide whether the model should be visible or not since `null` ==> falsy
}
```

app/controllers/admin/content/events.js
```diff
@@ -0,0 +1,95 @@
...
+          this.notify.success(this.get('l10n').t('This Event Property has been deleted successfully.'));
+        })
+        .catch(()=> {
+          this.notify.error(this.get('l10n').t('An unexpected error has occurred. Event Type was not deleted.'));
+        })
+        .finally(() => {
+          this.set('isLoading', false);
+        });
+    },
+    addEventType() {
```
niranjan94: All these add actions can just be 

```js
addEventProperty(modelInstance) {
  const camelCasedValue = camelCase(modelInstance.modelName);
  this.set('isLoading', true);
  modelInstance.save()
    .then(() => {
      this.notify.success(this.get('l10n').t(`${startCase(camelCasedValue)} has been added successfully.`));
    })
    .catch(() => {
      this.notify.error(this.get('l10n').t(`An unexpected error has occurred. ${startCase(camelCasedValue)} not saved.`));

    })
    .finally(() => {
      this.set(camelCasedValue, null);
      this.set('isLoading', false);
    })

}
```

And you can pass in the `modelInstance` in the template.

app/components/modals/admin/content/new-event-sub-topic-modal.js
```diff
@@ -0,0 +1,49 @@
+import FormMixin from 'open-event-frontend/mixins/form';
+import ModalBase from 'open-event-frontend/components/modals/modal-base';
+import { computed } from '@ember/object';
+import { later } from '@ember/runloop';
+
+export default ModalBase.extend(FormMixin, {
+  isSmall            : true,
+  autoScrollToErrors : false,
+
+  eventSubTopic: computed('model.topic', function() {
```
niranjan94: This is still pending. Remove usage of a computed property for this.

pradeepgangwar:
@niranjan94 Everything is working and code has been reduced. :smile: I was stuck at previous step for long but your solution was amazing and great. :smile: A good learning for me.

pradeepgangwar:
@niranjan94 Everything is working and code has been reduced. smile I was stuck at previous step for long but your solution was amazing. smile A good learning for me.
Please review the code.

app/controllers/admin/content/events.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+    },
+    openModalFor(modelInstanceOrName) {
+      const modelName = typeof modelInstanceOrName === 'string' ? modelInstanceOrName : modelInstanceOrName.modelName;
+      const modelInstance = typeof modelInstanceOrName === 'string' ? this.store.createRecord(modelInstanceOrName) : modelInstanceOrName;
+      const camelCasedValue = camelCase(modelName);
+      this.set(camelCasedValue, modelInstance);
+      switch (camelCasedValue) {
+        case 'eventType':
+          this.set('isNewEventTypeModalOpen', true);
```
niranjan94: Like I said before, these properties (`isNewEventTypeModalOpen` and such) are not needed. You can just directly pass in `eventType` to `isOpen`. That would be enough to control the model.

pradeepgangwar:
@niranjan94 did the changes. Please review.
I needed to add a openModal attribute to model instance. Without that it as giving me an error which says. `Assertion Failed: Cannot call set with 'name' on an undefined object` But now it works fine.

pradeepgangwar:
@niranjan94 did the changes. Please review.
I needed to add a openModal attribute to model instance. Without that it was giving me an error which says. Assertion Failed: Cannot call set with 'name' on an undefined object But now it works fine.

pradeepgangwar:
@niranjan94 Sorry for pushing my changes again, there was some problem with the edit button. I had to replace modelName with `constructor.modelName` to make it work. I didn't check that earlier. Everything seems to work now.
Please review again. :sweat_smile:

app/components/modals/admin/content/new-event-sub-topic-modal.js
```diff
@@ -0,0 +1,33 @@
...
+      });
+    }
+  },
+  getValidationRules() {
+    return {
+      inline : true,
+      delay  : false,
+      on     : 'blur',
+      fields : {
+        email: {
```
ritikamotwani: @pradeepgangwar why is it named email when the field is a name.

pradeepgangwar:
@niranjan94 @ritikamotwani Sorry for trouble again. Please review.

PR #1163, opened 2018-05-23
fix: Add port to locally generated url for Event

### LINKED ISSUE #1162: Locally generated URLs for events does not take port into account
**Describe the bug**
When using local deployment, url for events consists of only localhost and port is missing which leads us to wrong url.

**To Reproduce**
1. Create an event (if you don't have one)
2. Click on view event. 

You will see this in the general info about the event

**Expected behavior**
It should contain port number when using local deployment so that we do not have to manually enter the port to see events page.

**Additional context**
URL is a computed property of the event model calculated with this function
``` JS
url: computed('identifier', function() {
    return `${location.protocol}//${location.hostname}${this.get('router').urlFor('public', this.get('id'))}`;
  }),
```
**Originally opened here: https://github.com/fossasia/open-event-server/issues/4733**

srv-twry:
#### Checklist

Adds port to locally generated URL for Event.

URL is a computed property of the Event model. For local setup, adds the port number to the url.

Fixes #1162

### DIFF
--- a/app/models/event.js
+++ b/app/models/event.js
@@ -153,7 +153,7 @@ export default ModelBase.extend(CustomPrimaryKeyMixin, {
   }),
 
   url: computed('identifier', function() {
-    return `${location.protocol}//${location.hostname}${this.get('router').urlFor('public', this.get('id'))}`;
+    return location.origin + this.get('router').urlFor('public', this.get('id'));
   }),
 
   sessionsByState: computed('sessions', function() {


srv-twry:
fixed_url

.env.example
```diff
@@ -1,2 +1,3 @@
 GOOGLE_API_KEY="Sample Key"
 API_HOST=http://localhost:5000
+LOCAL=true
```
srv-twry: The issue with this solution is that if the developer doesn't specify LOCAL as true in the env file or the env file doesn't exist, then the port won't be added to the generated url. 
I couldn't figure out any other approach. Open for alternative approach.

niranjan94: This is not needed at all. You can just look for `location.port`. If you get a value, use it. Else don't

app/models/event.js
```diff
@@ -153,6 +153,9 @@ export default ModelBase.extend(CustomPrimaryKeyMixin, {
   }),
 
   url: computed('identifier', function() {
+    if (location.port) {
+      return `${location.protocol}//${location.hostname}:${location.port}${this.get('router').urlFor('public', this.get('id'))}`;
+    }
```
niranjan94: Actually, ... forgot about `location.origin`.

app/models/event.js
```diff
@@ -153,6 +153,9 @@ export default ModelBase.extend(CustomPrimaryKeyMixin, {
   }),
 
   url: computed('identifier', function() {
+    if (location.port) {
+      return `${location.protocol}//${location.hostname}:${location.port}${this.get('router').urlFor('public', this.get('id'))}`;
```
niranjan94: Actually, forgot about `location.origin`.

niranjan94: Actually, forgot about `location.origin`. 

```js
return location.origin + this.get('router').urlFor('public', this.get('id'));
```

niranjan94: Just use this. So, we don't need to worry about port.

PR #1213, opened 2018-05-31
fix: Remove unpublished records before leaving the route

### LINKED ISSUE #1212: Remove uncreated records from Ember data
**Describe the bug**

When we create a Sponsor in the application. We create a new record in the ember data. Now when the user clicks on the back button without filling in the necessary information, list still contains the redundant entry.

**To Reproduce**
1. Go to Specific Event dashboard and click on Speakers tab.
2. Click on Add speaker
3. Click on the back button
4. See the redundant speaker record.

**Expected behavior**

The record should be removed. I asked for possible ways to do that in the Ember JS community and was suggested two ways:
1. `{{#unless record.isNew}}` in the list
2. call rollbackAttributes on the record before leaving the route

**Screenshots**

before_speaker
after_speaker

srv-twry:
#### Checklist

When we create a Sponsor in the application. We create a new record in the ember data. Now when the user clicks on the back button without filling in the necessary information, list still contains the redundant entry.
Similar problem happens in multiple places in the application.

- In the deactivate hook, we use ```model.rollbackAttributes``` in case of Sessions/Speakers and Discount code.

@niranjan94 You had suggested me [here](https://github.com/fossasia/open-event-frontend/issues/1212#issuecomment-393551876) to use `unloadRecord` but I wasn't sure how to query the record from EmberData beforing calling unload on it. Hence i used the instructions described in this stackoverflow thread: https://stackoverflow.com/questions/27870072/ember-data-unsaved-model-is-not-discarded-when-navigating-back-with-the-back-but

Fixes #1212


niranjan94:
You don't need to query the record again. You just need the model. Even for attributes roll back you use the model right? Same way, unload it..

For a record that isn't persisted to the database, removing should be done by unloading. Not a prop roll back.

niranjan94:
Also, use `resetController` instead of `deactivate`. So that it is fired even on model changes.

app/routes/events/view/sessions/create.js
```diff
@@ -23,5 +23,10 @@ export default Route.extend({
       tracks       : await eventDetails.query('tracks', {}),
       sessionTypes : await eventDetails.query('sessionTypes', {})
     };
+  },
+  resetController() {
```
niranjan94: The first argument of this is `controller`. Use that instead of `this.controller`

app/routes/events/view/tickets/discount-codes/create.js
```diff
@@ -9,5 +9,8 @@ export default Route.extend({
       event   : this.modelFor('events.view'),
       usedFor : 'ticket'
     });
+  },
+  resetController() {
```
niranjan94: Same comment as above.

srv-twry:
@niranjan94 @CosmicCoder96  Please review

Issue #1244, opened 2018-06-05
Call for speaker private link not working


mayank8318:
**Describe the bug**
The private link provided during the wizard gives a 404 error.

**To Reproduce**
1. Create an event with a private call for speaker
2. Copy the link provided in the wizard
3. Publish the event
4. Try opening the link provided above

**Expected behavior**
Call for speakers should open as expected.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser Chrome

PR #1251, opened 2018-06-06
feat: Make download as iCal functional

### LINKED ISSUE #1211: Make export as iCal functional
Make export as iCal functional.

Use the same approach as used in export as a zip.

**Describe alternatives you've considered**
None

**Additional context**
To be done after https://github.com/fossasia/open-event-server/pull/4812 gets merged.

mayank8318:
#### Checklist

Makes the download as iCal button functional in the Export tab

Fixes #1211

### DIFF
--- a/app/templates/components/events/view/export/download-common.hbs
+++ b/app/templates/components/events/view/export/download-common.hbs
@@ -3,10 +3,16 @@
     {{t 'Download Event as'}} {{downloadType}}
   </div>
   <div class="sub header">
-    {{t 'Once the event is live and the schedule is published,'}} {{downloadType}} {{t ' version of the schedule will be available at:'}}
+    {{t 'Once the event is live and the schedule is published,'}} {{downloadType}} {{t
+    ' version of the schedule will be available at:'}}
   </div>
 </h3>
 <div class="ui secondary segment url">
-  <pre>http://localhost:5000/e/280a30d7/schedule/{{downloadUrl}}</pre>
+  <pre>{{displayUrl}}</pre>
 </div>
-<a class="ui link" href="#"> {{t 'Alternatively, you can download the'}} {{downloadType}} {{t ' here'}}</a>
+<a class="ui link" href="#" onclick={{action 'startExportTask'}}> {{t 'Alternatively, you can download the'}} {{downloadType}} {{t ' here'}}</a>
+<div class="ui basic segment less left padding">
+  <a href="{{eventDownloadUrl}}" class="ui blue button {{if isDownloadDisabled 'disabled' ''}} {{if isLoading 'loading'}}">
+    {{t 'Download '}}{{downloadType}}
+  </a>
+</div>
--- a/app/components/events/view/export/download-common.js
+++ b/app/components/events/view/export/download-common.js
@@ -1,4 +1,66 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { run } from '@ember/runloop';
+import ENV from 'open-event-frontend/config/environment';
 
 export default Component.extend({
+  isDownloadDisabled : true,
+  eventDownloadUrl   : '',
+  isLoading          : false,
+  fileFormat         : computed(function() {
+    switch (this.get('downloadType')) {
+      case 'iCalendar':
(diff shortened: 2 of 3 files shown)


mayank8318:
@niranjan94 I wasn't sure if I should put the download logic in the component or the controller. A lot of it is similar to the json download logic in the controller but it will increase the complexity of the controller due to format checks etc. So, I put in the component. Which one is a better implementation?

mayank8318:
@niranjan94 I am, its in the download button

app/templates/components/events/view/export/download-common.hbs
```diff
@@ -3,10 +3,16 @@
     {{t 'Download Event as'}} {{downloadType}}
   </div>
   <div class="sub header">
-    {{t 'Once the event is live and the schedule is published,'}} {{downloadType}} {{t ' version of the schedule will be available at:'}}
+    {{t 'Once the event is live and the schedule is published,'}} {{downloadType}} {{t
+    ' version of the schedule will be available at:'}}
   </div>
 </h3>
 <div class="ui secondary segment url">
   <pre>http://localhost:5000/e/280a30d7/schedule/{{downloadUrl}}</pre>
```
niranjan94: This URl shouldn't be hardcoded/

mayank8318: But I don't think this has been implemented yet because these public links do not work anywhere in the project, not even in version 1. How should I go about it? I am facing the same issue in https://github.com/fossasia/open-event-frontend/issues/1244

mayank8318:
@niranjan94 I have made it dynamic of sorts but we aren't creating these files on the server yet. Once I finish all three I will make an issue on the server and solve it. I am displaying what would be the eventual URL. Is this ok?

app/templates/components/events/view/export/download-common.hbs
niranjan94: The URL won't be something like `http://localhost:5000/e/280a30d7/schedule/{{downloadUrl}}`

niranjan94: The URL won't be something like `http://localhost:5000/e/280a30d7/schedule/{{downloadUrl}}` 

It would be the URL returned by the Backend. (Either s3, or google cloud storage or a direct static asset URL). Not an FE url. FE won't be having these files with itself.

niranjan94: And this is in no way related to https://github.com/fossasia/open-event-frontend/issues/1244

niranjan94: https://github.com/fossasia/open-event-frontend/issues/1244 is completely an FE issue

mayank8318: @niranjan94 Sorry, yes #1244 is related to the FE since the route is missing. Regarding this issue, I can display the ical_url once the event is published but if it is still in preview the server won't have this saved in the event database. So how should I predict the right one if it isn't hardcoded?

mayank8318: Because on the frontend we are saying that "once the event is published" the user can download it from that URL. So we are basically predicting the URL before the job has been done right? So, if we  do not know the location before hand what should I display there?

niranjan94: You shouldn't have to predict anything. You can just the URL the server returns as a result of the export task.

niranjan94: You shouldn't have to predict anything. You can just the URL the server returns as a result of the export task.

It is the server's responsibility to store the most recently exported iCal,xCal etc URLs against the event somewhere in the DB so that the FE can use that for other places.

niranjan94: > Because on the frontend we are saying that "once the event is published" the user can download it from that URL. So we are basically predicting the URL before the job has been done right? 

This shouldn't be the case. You should display the URL _only_ after generation. Please make changes accordingly.

mayank8318: Ok that clarifies things :) I will make the changes

Issue #1242, opened 2018-06-05
Hide hidden tickets from public route regardless of who is logged in


pradeepgangwar:
**Describe the bug**
Currently hidden tickets are visible to organizer on public route. If organizer is logged in he/she also should not see hidden tickets.

niranjan94:
@pradeepgangwar the adapter is not a route a level. It is at a model level. 

This can be easily implemented. You can just override the `headers` computed property within the adapter for the ticket model (do it only for this model). And don't add the the `Authorization` header if the current route it a public route.

Issue #1144, opened 2018-05-19
Tests fail because of "Cannot read property 'afterContentPaint' of null"


mayank8318:
**I'm submitting a ...**  (check one with "x")
- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, instead ask your query in out Gitter channel at https://gitter.im/fossasia/open-event-frontend

**Current behavior:**
Some tests are failing because of "Cannot read property 'afterContentPaint' of null".
This seems to be an ongoing issue with no proper fix https://github.com/dollarshaveclub/ember-router-scroll/issues/72

**Expected behavior:**
The tests should run properly.

**Steps to reproduce:**
Run `Ember test`

**Would you like to work on the issue?**
I will try to update dependencies and see if it works otherwise I will wait for a solution in the parent repository https://github.com/dollarshaveclub/ember-router-scroll/issues/72

srv-twry:
@mayank8318 Any update on this ?

niranjan94:
Seems to be resolved by https://github.com/fossasia/open-event-frontend/commit/f678cb17ebcaf33b7aa75c4497a89316a43ffeb3 

https://travis-ci.org/fossasia/open-event-frontend/builds/391272394

PR #1252, opened 2018-06-06
Hide side menu tabs on public routes if no data is present

### LINKED ISSUE #650: Modify side bar to hide tabs for missing data
- [x] bug report
- [ ] feature request

**Current behavior:**
screenshot from 2017-08-10 05-49-28

**Expected behavior:**
The tabs should be hidden if the data is not present.

### LINKED ISSUE #1243: Hide tabs in side menu on public route if no data is present
**Describe the bug**
Currently speakers and sessions are visible in side menu even if there are no entries for that. We should not be showing this unless we have approved speakers and sessions.

pradeepgangwar:
#### Checklist

Currently all the side menu tabs are displayed when they are enabled by organizer even if they do not contain any information. Purpose of this PR is to display tabs only if they contain some data inside them. This was discussed during recent open event wizard meeting. 

- Hides tabs if no data is present.
- CFS is enabled only if it has a announcement with it.

Fixes #650

### DIFF
--- a/app/templates/components/public/side-menu.hbs
+++ b/app/templates/components/public/side-menu.hbs
@@ -4,12 +4,12 @@
       {{#scroll-to href='#info' class='item active'}}
         {{t 'Info'}}
       {{/scroll-to}}
-      {{#if event.isTicketingEnabled}}
+      {{#if event.tickets.length}}
         {{#scroll-to href='#tickets' class='item'}}
           {{t 'Tickets'}}
         {{/scroll-to}}
       {{/if}}
-      {{#if event.isSessionsSpeakersEnabled}}
+      {{#if event.speakers.length}}
         {{#scroll-to href='#speakers' class='item'}}
           {{t 'Speakers'}}
         {{/scroll-to}}
@@ -18,18 +18,18 @@
       <a class="item" href="{{href-to 'public.index'}}">
         {{t 'Info'}}
       </a>
-      {{#if event.isTicketingEnabled}}
+      {{#if event.tickets.length}}
         <a class="item" href="{{href-to 'public.index'}}#tickets">
           {{t 'Tickets'}}
         </a>
       {{/if}}
-      {{#if event.isSessionsSpeakersEnabled}}
+      {{#if event.speakers.length}}
         <a class="item" href="{{href-to 'public.index'}}#speakers">
           {{t 'Speakers'}}
         </a>
       {{/if}}
     {{/if}}
-    {{#if event.isSessionsSpeakersEnabled}}
+    {{#if event.sessions.length}}
       {{#link-to 'public.sessions' class='item'}}
         {{t 'Sessions'}}
       {{/link-to}}
--- a/app/components/public/side-menu.js
+++ b/app/components/public/side-menu.js
@@ -5,6 +5,6 @@ export default Component.extend({
     this._super(...arguments);
     let speakersCall = await this.get('event.speakersCall');
     this.set('shouldShowCallforSpeakers',
-      this.get('event.isSessionsSpeakersEnabled') && (speakersCall.privacy === 'public'));
+      speakersCall.announcement && (speakersCall.privacy === 'public'));
   }
 });


pradeepgangwar:
- Here taking tickets.length into consideration makes sure that if if we have ticketing enabled with all hidden tickets we should not be showing tickets tab, because hidden tickets will not be visible on public route.
- Taking speaker.length into consideration makes sure that we show speakers tab only if we have speakers with us. Just having session speakers enabled does not guarantee having speakers with us.
- Similar argument for sessions.
- For CFS, we discussed during meeting that CFS without announcement is invalid. So I added that extra condition to check if announcement field is not empty.

pradeepgangwar:
@niranjan94 @mayank8318 Please review.

app/templates/components/public/side-menu.hbs
```diff
@@ -17,25 +17,22 @@
...
   {{/if}}
-  {{#if event.isSessionsSpeakersEnabled}}
+  {{#if event.sessions.length}}
     {{#link-to 'public.sessions' class='item'}}
       {{t 'Sessions'}}
     {{/link-to}}
   {{/if}}
-  {{#link-to 'public.schedule' class='item'}}
-    {{t 'Schedule'}}
-  {{/link-to}}
```
srv-twry: Why delete this ?

srv-twry:
side-question: There are two tests which fail usually i.e. about 75-80% of the time.
If we are not able to fix it and are ignoring it anyways, why can't we just disable it ?

@pradeepgangwar @ritikamotwani @mayank8318

app/templates/components/public/side-menu.hbs
pradeepgangwar: We do not have support to add schedule for a event. We just get a empty page when clicked on this. In our meeting we were told to remove it temporarily.

srv-twry: alright.

niranjan94:
> side-question: There are two tests which fail usually i.e. about 75-80% of the time. If we are not able to fix it and are ignoring it anyways, why can't we just disable it ?

@srv-twry please open an issue for fixing it. We _must_ be fixing it. That said, we can disable it until that issue is resolved.

srv-twry:
@niranjan94 There is already a created issue with details on the problem: #1144

niranjan94:
@CosmicCoder96 you're right. This is not the right way to do it. Esp. if an event has a lot a session, we'll be unnecessarily loading all events just to decide if we need to show a link or not. Instead the server could expose a `statistic` endpoint for an event that would just give the counts of all these. FE can use that.

pradeepgangwar:
I see there are some modifications happening in event-statistics api. I will wait it to be merged and then open a PR soon. 
Ref: https://github.com/fossasia/open-event-server/pull/4724

pradeepgangwar:
@CosmicCoder96 @niranjan94 I see that [event-statistic-general](https://open-event-api-dev.herokuapp.com/#event-statistics-event-statistics-details-get) endpoint allows only authorized user to access the endpoint. But if we are going to integrate it for public route then we will need that it can be accessed without login. Should I make necessary changes on server or something else is to be done?

CosmicCoder96:
@pradeepgangwar The current event statistics end point on server doesn’t give us the desired data. I think the best approach would be generate a new end point on the server which contains info required explicitly by this end point. All of this information is on a public route, so the end point shouldn’t require authorisation.

srv-twry:
@CosmicCoder96 @pradeepgangwar I think instead of adding a new endpoint, we can easily create two schemas for the same endpoint. The first schema will contain whatever data you want without any authorization. 
the second one would require authorization. So if the request to the endpoint has a JWT token we will use the second one or else we will use the first schema
It will be extremely easy to implement and use rather than creating a whole new endpoint.

CosmicCoder96:
@srv-twry 
>  So if the request to the endpoint has a JWT token we will use the second one or else we will use the first schema
But what if a user is logged in, and wants to visit the public route?

srv-twry:
@CosmicCoder96 We can do something similar may be : https://github.com/fossasia/open-event-frontend/issues/1242

srv-twry:
As far as i understand, any request from the public route shouldn't send a JWT token. The public route should look the same to everyone regardless of whether he/she is the admin, organizer, other registered user or an unregistered user.

srv-twry:
@CosmicCoder96 Actually @pradeepgangwar and @niranjan94 are trying to do the same for Tickets in this PR: https://github.com/fossasia/open-event-frontend/pull/1270. Hence I would wait a couple of days to see if they are able to customize it. If they can, then I will go ahead with this otherwise I will create a new endpoint according to your suggestions.

CosmicCoder96:
@bhaveshAn what has changed since the last discussion, for it to be approved or reviewed ?

bhaveshAn:
> I see there are some modifications happening in event-statistics api. I will wait for it to be merged and then update this PR.
Ref: fossasia/open-event-server#4724
> what has changed since the last discussion, for it to be approved or reviewed ?

@pradeepgangwar @CosmicCoder96 https://github.com/fossasia/open-event-server/pull/4724 just focused on creating event-type and event-topic statistics on `v1/admin/statistics/event-types` and `v1/admin/statistics/event-topics` while event statistics are the same `v1/admin/statistics/events`

CosmicCoder96:
>@pradeepgangwar @CosmicCoder96 fossasia/open-event-server#4724 just focused on creating event-type and event-topic statistics on v1/admin/statistics/event-types and v1/admin/statistics/event-topics while event statistics are the same v1/admin/statistics/events /v1/events/{event_identifier}/general-statistics

@bhaveshAn That's what I said here.
https://github.com/fossasia/open-event-frontend/pull/1252#pullrequestreview-127502835

srv-twry:
@pradeepgangwar Please resolve the conflicts and update this PR. 
I think we should merge this one and create a follow up issue on this. We can either add a new endpoint or handle JWT tokens in the frontend. We need this right now.

PR #1259, opened 2018-06-08
Make Stripe OAuth functional in event wizard

### LINKED ISSUE #1180: Make the Connect with Stripe button in the wizard functional
Make the Connect with Stripe button in the wizard functional. It isn't functional right now.

1. Fetch the ```client_id``` from the settings.
2. Create a route(redirect uri) to handle the response.
3. Add functionality to the button to direct to the OAuth Endpoint to register.
4. Handle the response(log it for now) in the redirect route created in the step 2.

**Parent issue**
#1172

pradeepgangwar:
#### Checklist

Currently in event creation wizard organizer do not have support to add their stripe account to enable payments. This PR makes connect to stripe button functional.

- Adds a new library torii which makes OAuth implementation simpler.
- Defines new functions and methods to make stripe button functional.

Fixes #1180

### DIFF
--- a/app/torii-providers/stripe.js
+++ b/app/torii-providers/stripe.js
@@ -1,0 +1,14 @@
+import stripeConnect from 'torii/providers/stripe-connect';
+import { alias } from '@ember/object/computed';
+import { inject } from '@ember/service';
+import { configurable } from 'torii/configuration';
+
+export default stripeConnect.extend({
+  settings: inject(),
+
+  clientId: alias('settings.stripeClientId'),
+
+  redirectUri: configurable('redirectUri', function() {
+    return `${window.location.origin}/torii/redirect.html`;
+  })
+});
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "qunit-dom": "^0.6.3",
     "sanitize-html": "^1.14.1",
     "semantic-ui-ember": "^3.0.3",
+    "torii": "^0.10.1",
     "url-parse": "^1.4.1",
     "xgettext-template": "^3.4.0"
   },
--- a/app/components/forms/wizard/basic-details-step.js
+++ b/app/components/forms/wizard/basic-details-step.js
@@ -11,10 +11,13 @@ import {
 import { countries } from 'open-event-frontend/utils/dictionary/demography';
 import FormMixin from 'open-event-frontend/mixins/form';
 import { orderBy, filter, find } from 'lodash';
+import { inject as service } from '@ember/service';
 
 export default Component.extend(FormMixin, {
 
   currentTimezone: moment.tz.guess(),
+
+  torii: service('torii'),
 
   getValidationRules() {
     return {
@@ -221,6 +224,19 @@ export default Component.extend(FormMixin, {
     this.getForm().form('remove prompt', 'discount_code');
   }),
   actions: {
+    connectStripe() {
+      this.get('data.event.stripeAuthorization.content') ? '' : this.set('data.event.stripeAuthorization', this.store.createRecord('stripe-authorization'));
+      this.get('torii').open('stripe')
+        .then(authorization => {
+          this.set('data.event.stripeAuthorization.stripeAuthCode', authorization.authorizationCode);
+        })
+        .catch(error => {
+          this.get('notify').error(this.get('l10n').t(`${error.message}. Please try again`));
+        });
+    },
+    disconnectStripe() {
+      this.get('data.event.stripeAuthorization.content').destroyRecord();
+    },
     saveDraft() {
       this.onValid(() => {
         this.set('data.event.state', 'draft');
@@ -318,6 +334,9 @@ export default Component.extend(FormMixin, {
     }
     if (!this.get('isCreate') && this.get('data.event.tax') && !this.get('data.event.tax.content')) {
       this.set('data.event.tax', this.store.createRecord('tax'));
+    }
+    if (!this.get('isCreate') && this.get('data.event.stripeAuthorization') && !this.get('data.event.stripeAuthorization.content')) {
+      this.set('data.event.stripeAuthorization', this.store.createRecord('stripe-authorization'));
     }
   }
 });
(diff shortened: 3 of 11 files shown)


pradeepgangwar:
Anybody reviewing current implementation please take care of following points. Features and modifications left:
- Disconnect button does not work. A function is to be implemented.
- Code has been not checked for corner cases. I will work on this soon.

package.json
```diff
@@ -100,6 +100,7 @@
     "qunit-dom": "^0.6.2",
     "sanitize-html": "^1.14.1",
     "semantic-ui-ember": "^2.1.0",
+    "torii": "^0.10.1",
```
niranjan94: We really don't need `torii` for this. We only have to handle the redirect URI. That's it. Which can be done just fine on Ember. Please remove torii dependency

pradeepgangwar: Okay :sweat_smile: . Only some minor things are left. I finish this quickly so that this PR can be reviewed as a whole. Thanks.

pradeepgangwar:
@niranjan94 @srv-twry @ritikamotwani @mayank8318 All the functionality related to stripe OAuth button seem to be working. Please review. :smile:

config/environment.js
```diff
@@ -77,6 +77,15 @@ module.exports = function(environment) {
     ifa: {
       enabled : false,
       inline  : false
+    },
+
+    torii: {
+      providers: {
+        'stripe-connect': {
+          clientId    : process.env.STRIPE_CLIENT_ID,
```
srv-twry: This client_id is supposed to be fetched from the settings endpoint :)

pradeepgangwar: @srv-twry I tried that but could not get idea about how to do it, since this file resides outside app and stripe picks its settings from this file. @niranjan94 Please help me if there is a way to do this.

niranjan94: @pradeepgangwar you will need to extend the stripe service and customize it to make it work as per our needs.

niranjan94: Extend the `stripe-connect` provider. Define the `clientId` property as an alias for the settings client ID property

niranjan94: https://github.com/Vestorly/torii/blob/9369efa96eeaf1a5404bab9994ed7c8f805edb1e/addon/providers/oauth2-code.js#L74

app/torii-providers/stripe.js
```diff
@@ -0,0 +1,29 @@
...
+
+  name    : 'stripe-connect',
+  baseUrl : 'https://connect.stripe.com/oauth/authorize',
+
+  requiredUrlParams : [],
+  optionalUrlParams : ['stripe_landing', 'always_prompt'],
+
+  responseParams: ['code', 'state'],
+
+  clientId: configurable('clientId', async function() {
```
pradeepgangwar: @niranjan94 I tried writing my custom provider. Everything seems to work fine. I am able to load clientId from settings here. But when I see the clientId in url (that gets formed), it is `object [Object]` . I printed various outputs of clientId just before return statement and it was proper string with correct clientId. But when url gets build I get it as object [Object]. I tried it for several hours with different options but none worked. I think I am missing something here. Please help.

niranjan94: Hey @pradeepgangwar . Let me pull this in locally and try it out.

app/torii-providers/stripe.js
```diff
@@ -0,0 +1,29 @@
+import Oauth2 from 'torii/providers/oauth2-code';
+import { configurable } from 'torii/configuration';
+import { inject as service } from '@ember/service';
+
+export default Oauth2.extend({
```
niranjan94: Extend the stripe provider provided by torii. Just extend. Don't copy paste what is there. That will lead to code duplication. We only want to change one property there.

app/torii-providers/stripe.js
niranjan94: Just make this 

```js
settings: inject(),

clientId: alias('settings.stripeClientId')
```

(Make sure to import `alias`)

pradeepgangwar:
@niranjan94 that worked like charm :smile: :fireworks: . Thanks for the help!.
Please review.

pradeepgangwar:
@niranjan94 that worked smile fireworks . Thanks for the help!.
Please review.

app/torii-providers/stripe.js
```diff
@@ -0,0 +1,26 @@
+import stripeConnect from 'torii/providers/stripe-connect';
+import { alias } from '@ember/object/computed';
+import { inject } from '@ember/service';
+import { configurable } from 'torii/configuration';
+
+function currentUrl() {
```
niranjan94: Why do you need this ?

pradeepgangwar: If i don't include this. It sends redirect to /create/torii/redirect.html and gives error that it cannot parse it. Since on stripe dashboard I have given it as localhost:4200/torii/redirect.html . So either I change it on my stripe dashboard or change it here (I did this to omit the /create part from url)? Please suggest.

pradeepgangwar: @niranjan94 I did some more research on this and here is what is found:
1. From docs of torii they said that
> the redirect URL you register with the OAuth provider(s) that you use should be: your app base URL/torii/redirect.html

Other redirect URIs were deprecated by torii and have potential vulnerability as given [here](https://github.com/Vestorly/torii/blob/master/README.md#oauth-redirects). 

2. Also if I use redirect uri as `http://localhost:4200/create/torii/redirect.html it shows this error message 
screenshot from 2018-06-13 15-08-54

Please suggest me if any changes are required.

niranjan94: That's fine. What i mean is, why do you need to override it here when it is already done at https://github.com/Vestorly/torii/blob/master/addon/providers/oauth2-code.js#L107. You don't even have to define redirectUri here. It is already part of the super class

pradeepgangwar: @niranjan94 The definition given there also takes into account windows.location.pathname which will eventually include `create` and `events/<event_id>/edit/basic-details` (editing created event) before `torii/redirect.html` . So I thought overriding existing function so that redirect uri is always `http://localhost:4200/torii/redirect.html` (removed windows.location.pathname from function).

niranjan94: Okay. Couldn't you use 

```js
return `${window.location.origin}/torii/redirect.html` 
```

pradeepgangwar:
@niranjan94 did the changes. 
Please review.

niranjan94:
Good work!

PR #1265, opened 2018-06-09
fix: Private CFS not working

### LINKED ISSUE #1244: Call for speaker private link not working

mayank8318:
#### Checklist

Add a route for private CFS link

- Re-direct to the cfs page if the hash in the url is correct.

Fixes #1244 

@niranjan94 There is one flaw in this which was there from before. The public cfs route doesn't check the privacy. So directly accessing `/cfs` from the address bar works irrespective of the privacy. I don't know how to handle this efficiently. I can add a check in the public route and define the whole private route but that would involve code repetition. Maybe we can pass an additional variable to the public route which would signify if the hash was correct or not in case of a private cfs.
```javascript
if(cfs.privacy === 'public' || (cfs.privacy === 'private' && isVerified === true)) {
// go ahead
}
``` 
But I am not really sure about the second method. How should I approach this?

### DIFF
--- a/app/routes/public/cfs/index.js
+++ b/app/routes/public/cfs/index.js
@@ -5,6 +5,21 @@ export default Route.extend({
     return this.get('l10n').t('Call for Speakers');
   },
 
+  async beforeModel(transition) {
+    let hash = transition.params['public.cfs'].speaker_call_hash;
+    const eventDetails = this.modelFor('public');
+    const speakersCall = await eventDetails.get('speakersCall');
+    /*
+    The following should show the CFS page to the user:
+     - CFS is public and no hash is entered
+     - CFS is public and a valid hash is entered
+     - CFS is private and a valid hash is entered
+    */
+    if (!((speakersCall.privacy === 'public' && (!hash || speakersCall.hash === hash)) || (speakersCall.privacy === 'private' && hash === speakersCall.hash))) {
+      this.transitionTo('not-found');
+    }
+  },
+
   async model() {
     const eventDetails = this.modelFor('public');
     if (this.get('session.isAuthenticated')) {
--- a/app/router.js
+++ b/app/router.js
@@ -39,6 +39,10 @@ router.map(function() {
     this.route('sessions', function() {
       this.route('list', { path: '/:track_id' });
     });
+    this.route('cfs', { path: '/cfs/:speaker_call_hash' }, function() {
+      this.route('new-speaker');
+      this.route('new-session');
+    });
     this.route('cfs', function() {
       this.route('new-speaker');
       this.route('new-session');


niranjan94:
@mayank8318 You can check it in the public cfs route's `beforeModel` hook. If it public cfs, allow user to proceed. Else replace route with the not-found route

mayank8318:
@niranjan94 But I will also land on the same page if I enter the right hash since I am redirecting to this route.

niranjan94:
Don't do it like that @mayank8318 . 

Instead have it like this,

`/e/<event-id>/cfs` - for public cfs. 
`/e/<event-id>/cfs/<hash>` - for private cfs. 

Make sure both are handled by the cfs route. Now in the beforeModel hook, you can check the cases and allow/disallow. 

---

To be clear, 

instead of

```js
this.route('cfs', function() {
  this.route('new-speaker');
  this.route('new-session');
  this.route('private', { path: '/:speaker_call_hash' });
});	  
```

I'm proposing,

```js
this.route('cfs', { path: '/cfs/:speaker_call_hash' }, function() {
  this.route('new-speaker');
  this.route('new-session');
});	  
```

app/routes/public/cfs/index.js
```diff
@@ -5,6 +5,15 @@ export default Route.extend({
     return this.get('l10n').t('Call for Speakers');
   },
 
+  async beforeModel(transition) {
+    let hash = transition.params['public.cfs'].speaker_call_hash;
+    const eventDetails = this.modelFor('public');
+    const speakersCall = await eventDetails.get('speakersCall');
+    if (speakersCall.hash !== hash) {
+      this.transitionTo('not-found');
```
niranjan94: This should only be done in the following cases.

- CFS Privacy setting is private and the hash is either empty or wrong
- Hash is not empty and incorrect

---

The following should show the CFS page to the user

- CFS is public and no hash is entered
- CFS is public and a valid hash is entered
- CFS is private and a valid hash is entered

app/templates/components/public/side-menu.hbs
```diff
@@ -37,7 +37,7 @@
     {{t 'Schedule'}}
   {{/link-to}}
   {{#if shouldShowCallforSpeakers }}
-    {{#link-to 'public.cfs' class='item'}}
+    {{#link-to 'public.cfs' event.speakersCall.hash class='item'}}
```
niranjan94: This is wrong. You shouldn't be putting the hash here. This would allow anyone to get the private cfs url by just going to the public event page.

app/routes/public/cfs/index.js
mayank8318: `/cfs` doesn't get mapped to 
```javascript
 this.route('cfs', { path: '/cfs/:speaker_call_hash' }, function() {
      this.route('new-speaker');
      this.route('new-session');
    });
```
I get a 404 instead, so I was passing the hash in the public url but that is major security breach as pointed out. Couldn't find any resource where one route was mapped to two different urls

mayank8318: Is it okay if I do this - 
```javascript
this.route('cfs', function() {
      this.route('new-speaker');
      this.route('new-session');
    });
this.route('cfs', { path: '/cfs/:speaker_call_hash' }, function() {
      this.route('new-speaker');
      this.route('new-session');
    });
```
https://stackoverflow.com/questions/24923120/emberjs-using-multiple-paths-urls-for-one-route?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

niranjan94: Hey @mayank8318 ... Sorry for the late reply. Yep. You can implement it like this. This is fine

mayank8318:
@niranjan94 Please review now :)

mayank8318:
@pradeepgangwar @ritikamotwani @srv-twry @agbilotia1998 rebasing dismissed your reviews, kindly review again :)

PR #1309, opened 2018-06-21
feat: Redirect to previous route after login

### LINKED ISSUE #1294: Redirect to previous page after login
**Describe the bug**

Right now when the user logs in, he/she is redirected back to the public page after successful.

**To Reproduce**
1. Login from any route

**Expected behavior**

The user should be redirected back to the previous page.

Working on it.

srv-twry:
#### Checklist

Right now when the user logs in, he/she is redirected back to the public page.

Redirect to previous route after login

Fixes #1294

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -1,7 +1,10 @@
 import Route from '@ember/routing/route';
 import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';
+import { inject as service } from '@ember/service';
+import { merge, values, isEmpty } from 'lodash';
 
 export default Route.extend(ApplicationRouteMixin, {
+  session: service(),
   title(tokens) {
     if (!tokens) {
       tokens = [];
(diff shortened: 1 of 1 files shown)


srv-twry:
@pradeepgangwar @ritikamotwani @mayank8318 Please review. 
I have to use the `window` object but I don't have any other options. I have tried everything that I could think of.

1. Tried using the `transition.targetName` but it doesn't work for the case when the user goes to the `Create Event` section and is automatically redirected to the `Login` section since the transition never really occurs. I also created an issue in the Ember simple auth and got some guidance but the suggestions don't really work very well. It also doesn't work for routes which take params. I had opened an issue in the ember-simple-auth and I was suggested to use this method. Issue: https://github.com/simplabs/ember-simple-auth/issues/1618

2. I tried using the `transition.intent.url` it works fine for most cases but for transition from parent to child routes, it is undefined hence can't use it in order to redirect to url such as `http://localhost:4200/e/fac99f4d/coc` although it works great for url such as `http://localhost:4200/e/fac99f4d/`.

```Javascript
if (transition.targetName !== 'login' && transition.intent.url) {
       this.set('session.previousRouteName', transition.intent.url);
      }
```

PR #1312, opened 2018-06-22
fix: Fix the Date filters in explore route

### LINKED ISSUE #1296: Date filters don't work in Explore Events tab
**Describe the bug**
Date filters dont work in Explore Events tab

**To Reproduce**
1. Go to explore events
2. Try the date filters

**Expected behavior**
Filters should work

 - OS: Ubuntu 18.04
 - Browser Chrome

mayank8318:
#### Checklist

Fixes the date filters in the explore route

- Give the user option of selecting dates manually instead of providing options like Today, yesterday etc.

Fixes #1296

### DIFF
--- a/app/components/explore/side-bar.js
+++ b/app/components/explore/side-bar.js
@@ -1,29 +1,24 @@
 import Component from '@ember/component';
-import { computed } from '@ember/object';
-import { getDateRanges } from 'open-event-frontend/utils/dictionary/filters';
+import moment from 'moment';
 
 export default Component.extend({
 
   classNames: ['ui', 'fluid', 'explore', 'vertical', 'menu'],
-
-  dateRanges: computed(function() {
-    return getDateRanges.bind(this)();
-  }),
 
   actions: {
     selectCategory(category, subCategory) {
       this.set('category', (category === this.get('category') && !subCategory) ? null : category);
       this.set('sub_category', (!subCategory || subCategory === this.get('sub_category')) ? null : subCategory);
     },
+
     selectEventType(eventType) {
       this.set('event_type', eventType === this.get('event_type') ? null : eventType);
     },
-    selectDateRange(dateRange) {
-      let isCustomDate = null;
-      if (dateRange === 'custom_dates') {
-        isCustomDate = dateRange;
+
+    dateValidate(date) {
+      if (moment(date).isAfter(this.get('endDate'))) {
+        this.set('endDate', date);
       }
-      this.set('date_range', dateRange === this.get('date_range') ?  isCustomDate : dateRange);
     }
   }
 });
--- a/app/controllers/explore.js
+++ b/app/controllers/explore.js
@@ -1,93 +1,14 @@
 import Controller from '@ember/controller';
-import EmberObject, { computed } from '@ember/object';
 import moment from 'moment';
 
 export default Controller.extend({
-  queryParams  : ['category', 'sub_category', 'event_type', 'date_range'],
+  queryParams  : ['category', 'sub_category', 'event_type', 'start_date', 'end_date'],
   category     : null,
   sub_category : null,
   event_type   : null,
-  date_range   : null,
+  start_date   : moment().toISOString(),
+  end_date     : null,
 
-  categoryFilter: computed('category', function() {
-    return EmberObject.create(
-      {
-        name : 'event-topic',
-        op   : 'has',
-        val  : {
-          name : 'name',
-          op   : 'eq',
-          val  : this.get('category')
-        }
-      }
-    );
-  }),
-
-  subCategoryFilter: computed('sub_category', function() {
-    return EmberObject.create(
-      {
-        name : 'event-sub-topic',
-        op   : 'has',
-        val  : {
-          name : 'slug',
-          op   : 'eq',
-          val  : this.get('sub_category')
-        }
-      }
-    );
-  }),
-
-  typeFilter: computed('event_type', function() {
-    return EmberObject.create(
-      {
-        name : 'event-type',
-        op   : 'has',
-        val  : {
-          name : 'name',
-          op   : 'eq',
-          val  : this.get('event_type')
-        }
-      }
-    );
-  }),
-
-  filterOptions: [
-    {
-      name : 'starts-at',
-      op   : 'ge',
-      val  : moment().toISOString()
-    },
-    {
-      name : 'state',
-      op   : 'eq',
-      val  : 'published'
-    }
-  ],
-
-  filteredEvents: computed('category', 'sub_category', 'event_type', 'date_range',  function() {
-    let category = this.get('category');
-    let subCategory = this.get('sub_category');
-    let eventType = this.get('event_type');
-    let filterOptions = this.get('filterOptions').slice(0, 2);
-
-    if (category) {
-      filterOptions.pushObject(this.get('categoryFilter'));
-
-      if (subCategory) {
-        this.get('filterOptions').pushObject(this.get('subCategoryFilter'));
-      }
-    }
-
-    if (eventType) {
-      this.get('filterOptions').pushObject(this.get('typeFilter'));
-    }
-
-    return this.get('store').query('event', {
-      sort   : 'starts-at',
-      filter : filterOptions
-    });
-
-  }),
   actions: {
     shareEvent(event) {
       this.set('eventToShare', event);
--- a/app/templates/components/explore/side-bar.hbs
+++ b/app/templates/components/explore/side-bar.hbs
@@ -54,37 +54,30 @@
   {{/ui-accordion}}
 </div>
 <div class="item">
-  {{#ui-accordion}}
-    <span class="title">
-      <i class="dropdown icon"></i>
-      {{t 'Date'}}
-    </span>
-    <div class="content menu">
-      {{#each dateRanges as |dateRange|}}
-        <a href="#" class="link item {{if (eq date_range dateRange.key) 'active'}}" {{action 'selectDateRange' dateRange.key}}>
-          {{dateRange.name}}
-          {{#if (and (eq dateRange.key 'custom_dates') (eq date_range 'custom_dates'))}}
-            <div class="explore sub menu">
-              <div class="ui form">
-                <div class="grouped fields">
-                  <div class="item field">
-                    <label class="required" for="start_date">{{t 'Starts'}}</label>
-                    {{widgets/forms/date-picker type='text'
-                                                rangePosition='start'
-                                                inputId='start_date'}}
-                  </div>
-                  <div class="item field">
-                    <label class="required" for="end_date">{{t 'Ends'}}</label>
-                    {{widgets/forms/date-picker type='text'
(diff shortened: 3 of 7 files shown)


app/controllers/explore.js
```diff
@@ -74,20 +70,32 @@ export default Controller.extend({
...
     }
 
-    return this.get('store').query('event', {
+    let filteredResults = [];
+
+    await this.get('store').query('event', {
       sort   : 'starts-at',
       filter : filterOptions
+    }).then(results => {
+      filteredResults = results.filter(event => {
```
niranjan94: Please do this on the server side.

niranjan94: To be done on the server-side.

https://flask-rest-jsonapi.readthedocs.io/en/latest/filtering.html

mayank8318:
@niranjan94 made the changes :)

app/components/explore/side-bar.js
```diff
@@ -1,13 +1,15 @@
...
+import { observer } from '@ember/object';
+import moment from 'moment';
 
 export default Component.extend({
 
   classNames: ['ui', 'fluid', 'explore', 'vertical', 'menu'],
 
-  dateRanges: computed(function() {
-    return getDateRanges.bind(this)();
+  dateChange: observer('start_date', 'end_date', function() {
```
niranjan94: Do not use an observer for this. Always use an observer only when there's really no other option.

In this case, you should be using a one way binding for the date picker and make it such that it can accept a call back action with the selected date value. (similar to how the ui-dropdown component works) In that action you can do whatever you do here.

This way the data also flows correctly and is easy to catch bugs.

Observers can easily cause weird issues  that there's even an eslint rule to prevent its usage :P https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-observers.md

mayank8318: @niranjan94 But I will still need an observer in the date-picker, right? I think I am not getting the required implementation here. I need something like this in the date-picker right -
```javascript
[observe for change in start_date] {
  this.sendAction('callback_name');
}
```

mayank8318:
@niranjan94 changed the implementation, please review :)

app/templates/components/explore/side-bar.hbs
```diff
@@ -54,37 +54,31 @@
...
+  <div class="content menu">
+    <div class="explore sub menu">
+      <div class="ui form">
+        <div class="grouped fields">
+          <div class="item field">
+            <label class="required" for="start_date">{{t 'Starts'}}</label>
+            {{widgets/forms/date-picker type='text'
+                                        rangePosition='start'
+                                        value=start_date
+                                        valueCheck=end_date
```
niranjan94: Remove this property. date-picker does not need this.

app/templates/components/explore/side-bar.hbs
```diff
@@ -54,37 +54,31 @@
...
+                                        valueCheckCallback=(action 'dateValidate')
```
niranjan94: Call it it `onChange` to match with other components. The callback should receive the date selected in the datepicker.

mayank8318:
@niranjan94 made the changes :)

app/controllers/explore.js
```diff
@@ -37,6 +38,78 @@ export default Controller.extend({
     );
   }),
 
+  dateFilterWithEndDate: computed('start_date', 'end_date', function() {
+    return EmberObject.create(
```
niranjan94: Why do you need this to be an `EmberObject` ? Just a plain object would do.

app/controllers/explore.js
```diff
@@ -37,6 +38,78 @@ export default Controller.extend({
...
+              ]
+            }
+          ]
+      }
+
+    );
+  }),
+
+  dateFilterBasic: computed('start_date', function() {
+    return EmberObject.create(
```
niranjan94: Why do you need this to be an `EmberObject` ? Just a plain object would do.

app/controllers/explore.js
```diff
@@ -37,6 +38,78 @@ export default Controller.extend({
...
+            op   : 'ge',
+            val  : this.get('start_date')
+          }
+        ]
+      }
+    );
+  }),
+
   typeFilter: computed('event_type', function() {
     return EmberObject.create(
```
niranjan94: Why do you need this to be an `EmberObject` ? Just a plain object would do.

app/controllers/explore.js
```diff
@@ -52,42 +125,44 @@ export default Controller.extend({
...
     let eventType = this.get('event_type');
+    let endDate = this.get('end_date');
     let filterOptions = this.get('filterOptions').slice(0, 2);
 
     if (category) {
       filterOptions.pushObject(this.get('categoryFilter'));
 
       if (subCategory) {
-        this.get('filterOptions').pushObject(this.get('subCategoryFilter'));
+        filterOptions.pushObject(this.get('subCategoryFilter'));
```
niranjan94: Use native `push`

mayank8318:
@niranjan94 made the changes :)

mayank8318:
@niranjan94 But I am forming the filters dynamically right? How will I handle that?

niranjan94:
> @niranjan94 But I am forming the filters dynamically right? How will I handle that?

You can still do that based on the params inside the `model` fn in the route right?

mayank8318:
@niranjan94 Oh okay so if `param` contains that string say category, for example, I know that the user has selected a particular category and then I can add it to the filter option. Is this right?

niranjan94:
@mayank8318 `params` is an object. Its going to contain the values of all the query params that were defined. 

Example,

```
{
 	category: "whatever",
    sub_category: "whatever",
    event_type: "whatever",
    from_date: "whatever",
    to_date: "whatever"
}
```

You will be doing exactly what you're already doing within the computed properties, except, it'll be within the `model` fn. Make sure the `queryParams` array on the controller contains all of them and the the queryParams definiton on the route contains `refreshModel` as `true` for all the params. 

For reference: 
- https://guides.emberjs.com/release/routing/query-params/
- https://guides.emberjs.com/release/routing/query-params/#toc_opting-into-a-full-transition

mayank8318:
@niranjan94 Yes, I am able to implement that but it refreshes the page every time some filter is changed. I don't think this is a good user experience. For eg, if I open the categories drop down and click on some category, the whole page is refreshed. I get the desired search results but I have to restart my searching process by re-clicking on the dropdown icon.

mayank8318:
@niranjan94 No, the list isn't getting updated

niranjan94:
Make sure you use `filteredEvents` and not `model.filteredEvents` in the template/controller.

mayank8318:
@niranjan94 Oh ok now it is working. So, just out of curiosity - `controller.set('filteredEvents', model.filteredEvents);` just initialises the filteredEvents variable of the controller right?

niranjan94:
@mayank8318 yes. Correct

app/templates/explore.hbs
```diff
@@ -1,6 +1,6 @@
 <div class="ui stackable grid">
   <div class="four wide column">
-    {{explore/side-bar model=model category=category sub_category=sub_category event_type=event_type date_range=date_range}}
+    {{explore/side-bar model=model category=category sub_category=sub_category event_type=event_type start_date=start_date end_date=end_date}}
```
niranjan94: use camelCase

```hbs
{{explore/side-bar model=model category=category subCategory=sub_category eventType=event_type startDate=start_date endDate=end_date}}
```

mayank8318: Should I change it everywhere?

mayank8318: I mean even the query params?

mayank8318: @niranjan94 These are not in camel case because the template is already using the camel-case version for iterating - 
```javascript
 {{#if (eq category Category.name)}}
          {{#each Category.subTopics as |subCategory|}}
            <a href="#"
               class="link inner item {{if (eq sub_category subCategory.slug) 'active'}}"
              {{action 'selectCategory' Category.name subCategory.slug}}>
              {{subCategory.name}}
            </a>
          {{/each}}
        {{/if}}
```

niranjan94: All query params - snake_case
Every other js variable - camelCase

mayank8318: But what about the camel cased variables(same name) in the template?

niranjan94: Which template ? 
- If it is on the route's template, and it refers to the query param variable snake_case if fine since its the same variable.
- Anywhere else, camelCase

mayank8318: The `explore/side-bar` template, the one which contains all the filtering options(FE wise). We are passing the values of the query params to it in the highlighted line. The problem is that the name of the query params is same as the model property if written in camel case. That's why we are passing variables in snake_case only. I have already pointed out an example(variable `sub_category` and `subCategory`) I will have to rename that `subCategory` to something else maybe `subCategoryItem`. Is that okay?

Variables which have this problem -> `sub_category` and `event_type`. Rest can be converted to camel case without a problem.

mayank8318:
@niranjan94 I have converted `start_date` and `end_date` to camel case

mayank8318:
@CosmicCoder96 @niranjan94 @ritikamotwani @pradeepgangwar Please review

mayank8318:
@ritikamotwani @srv-twry @pradeepgangwar @bhaveshAn @mishuvs Please test the functionality here - 
https://deploy-preview-1312--open-event-frontend.netlify.com/explore

ritikamotwani:
@mayank8318 
image
This is the UI for the date filtering for the explore route and to filter according to the start and end date you have to click on the custom dates and the accordion appears
image
In this PR for the date filtering you have to enter the start and the end date none of the other things is there.

mayank8318:
@ritikamotwani Yes correct. IMO this makes more sense since we were giving this option anyways.

mayank8318:
That extra feature can be added in another PR. This implements the basic functionality.

ritikamotwani:
@mayank8318 the current UI makes more sense as if some user wants to get tomorrow's event clicking on tomorrow will be more convenient for him than entering the start and end date! This basic filtering can be reviewed for now but an issue should be created for implementing it on the basis of the current UI

mayank8318:
@ritikamotwani ok I will make an issue for that

ritikamotwani:
@mayank8318 a doubt
I tried filtering events from 13th July to 14th July 
image
I am getting events of 14th June too 
image

pradeepgangwar:
@ritikamotwani I had the same doubt. Actually the end date of that event should be between the filtered dates. So implementation is like this: If any portion of event timeline falls between the selected time period then event is displayed.

ritikamotwani:
@pradeepgangwar okay!

mayank8318:
@pradeepgangwar right! If any portion of the event timeline falls between the selected time period then that event is considered active and included in the list.

srv-twry:
@niranjan94 Merged it. In case you find anything wrong, you can follow up by creating an issue or in the channel. Thanks.

PR #1313, opened 2018-06-23
[ALT] Redirect to previous route on login via transition

niranjan94:
Alternate solution for https://github.com/fossasia/open-event-frontend/pull/1309 without touching the `window` object or handling `paths`

### DIFF
--- a/app/components/nav-bar.js
+++ b/app/components/nav-bar.js
@@ -4,7 +4,7 @@ export default Component.extend({
   actions: {
     logout() {
       this.get('authManager').logout();
-      this.get('routing').transitionTo('index');
+      this.get('router').transitionTo('index');
     }
   }
 });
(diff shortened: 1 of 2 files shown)


pradeepgangwar:
@niranjan94 This works, as when I transition to a login route from any other route it redirects back to same route. But when I am on a specif events route for ex. `/e/<event_id>/cfs` I get 500 error saying `Error while processing route: public.cfs.index eventDetails.get is not a function` and on  `/e/<event_id>/sessions` I get redirected to `/e/undefined/sessions`. I think event_id does not retain and both the issues are related. In both of the errors we actually were on redirected route when error occurred.

niranjan94:
@pradeepgangwar let me try that case out locally.

maxlorenz:
What is the status regarding https://github.com/fossasia/open-event-frontend/pull/1309 has been merged? @niranjan94

srv-twry:
@maxlorenz #1309 just solves the problem. I am saving the previous route's URL. It is not the best way to do things in Ember. @niranjan94 is trying to do it in an elegant way in this PR. So IMO, we can leave this PR open.

PR #1318, opened 2018-06-24
fix: records not visible after creation

### LINKED ISSUE #1317: Template does not update after creating a record (does not show created record)
**Describe the bug**
At some places in app template does not show created record until we refresh.

**Expected behavior**
Template should update as soon as any data/record is changed.

pradeepgangwar:
#### Checklist

Currently in some parts of app, records are not visible after creation until we refresh the page.

- Make necessary changes to make to make it work.

Fixes #1317


app/routes/admin/content/events.js
```diff
@@ -6,12 +6,12 @@ export default Route.extend({
   },
   async model() {
     return {
-      'eventTopics': await this.get('store').query('event-topic', {
+      'eventTopics': await this.get('store').findAll('event-topic', {
```
mayank8318: @pradeepgangwar Will we have to change it at other places as well?

niranjan94: You cannot use `findAll` here. `findAll` will not support query parameters. You _must_ use `query`

niranjan94: Reload the model if need-be but don't use findAll when you need to query stuff.

pradeepgangwar:
I tried to fix this where I was able to find such behaviour. If you find it
anywhere else let me know. I was aware of these areas only.

app/routes/admin/content/events.js
```diff
@@ -13,5 +13,10 @@ export default Route.extend({
 
       'eventTypes': await this.get('store').query('event-type', {})
     };
+  },
+  setupController(controller, model) {
+    this._super(...arguments);
+    controller.set('model.eventTypes', model.eventTypes.toArray());
+    controller.set('model.eventTopics', model.eventTopics.toArray());
```
pradeepgangwar: @niranjan94 I removed `findAll` but was not able to reload model for this route. After reading docs I figured out that `reload()` is available for record and not collection of records. So I thought of converting model to array and adding/removing record from array each time a record is changed.
I will be happy to know if there is any other way of approaching this. Please review these changes. :)

niranjan94: I don't get why this conversion is needed. The result of the query is already an object that extended `Ember.ArrayProxy`. So, you can actually push to it directly. 

https://emberjs.com/api/ember/3.2/classes/ArrayProxy/methods/addObject?anchor=addObject

pradeepgangwar: @niranjan94 I tried `addObject` directly on `model.eventTypes` without using `toArray()`. But here is what I get:
```
events.js:115 Error: The result of a server query (on event-type) is immutable.
    at Class.replace (-private.js:9800)
    at Class.insertAt (array.js:566)
    at Class.pushObject (array.js:573)
    at Class.addObject (array.js:652)
    at events.js:113
    at tryCatcher (rsvp.js:200)
    at invokeCallback (rsvp.js:372)
    at publish (rsvp.js:358)
    at rsvp.js:14
    at invoke (backburner.js:205)
```
I searched on google and found some links but they are used `toArray()`. Some of them are:
https://github.com/emberjs/data/issues/370
https://github.com/offirgolan/ember-light-table/issues/423

Let me know if I missed something and if there are any changes that might fix this issue.

mayank8318:
@pradeepgangwar What is the status of this PR?

pradeepgangwar:
@mayank8318 I have made the changes and it is ready to be reviewed.

PR #1407, opened 2018-07-14
Add helper text to add session/speaker form

### LINKED ISSUE #1406: In the form to add a new session/speaker helper text is missing
**Describe the bug**

In case there are no existing sessions or speakers of an event helper text should come under teh heading choose an existing session.

 - OS: Ubuntu
 - Browser: Chromium

ritikamotwani:
#### Checklist

- In the add speaker form if there are no sessions for the event helper text should come.
- In the create session form if there are no speakers for the event helper text should come.

Fixes #1406

### DIFF
--- a/app/components/forms/session-speaker-form.js
+++ b/app/components/forms/session-speaker-form.js
@@ -302,6 +302,16 @@ export default Component.extend(FormMixin, {
     return groupBy(this.get('fields').toArray(), field => field.get('form'));
   }),
 
+  // Clicking on the add session button creates a blank record which increases the length of the speaker's list by 1.
+  noSpeakerExists: computed('speakers', function() {
+    return (this.get('speakers').length === 1);
+  }),
+
+  // Clicking on the add speaker button creates a blank record which increases the length of the session's list by 1.
+  noSessionExists: computed('sessions', function() {
+    return (this.get('sessions').length === 1);
+  }),
+
   shouldShowNewSpeakerDetails: computed('speakerDetails', 'newSpeakerSelected', function() {
     return this.get('newSpeakerSelected') && !this.get('speakerDetails');
   }),
--- a/app/templates/components/forms/session-speaker-form.hbs
+++ b/app/templates/components/forms/session-speaker-form.hbs
@@ -69,6 +69,9 @@
         {{/if}}
       </div>
     {{/each}}
+    {{#if noSpeakerExists}}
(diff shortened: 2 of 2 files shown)


app/components/forms/session-speaker-form.js
```diff
@@ -302,6 +302,14 @@ export default Component.extend(FormMixin, {
     return groupBy(this.get('fields').toArray(), field => field.get('form'));
   }),
 
+  noSpeakerExists: computed('speakers', function() {
+    return (this.get('speakers').length === 1);
+  }),
+
+  noSessionExists: computed('sessions', function() {
+    return (this.get('sessions').length === 1);
```
mayank8318: Why === 1 ?

srv-twry: Add a comment above the computed property since this can be confusing.

ritikamotwani: Okay I'll add a comment and also if someone finds a more appropriate approach using unloadRecord or something can replace this method till then I think we can merge this as for the events with no sessions the heading looks very absurd without the helper text.

ritikamotwani:
@srv-twry please review.

app/components/forms/session-speaker-form.js
```diff
@@ -302,6 +302,16 @@ export default Component.extend(FormMixin, {
     return groupBy(this.get('fields').toArray(), field => field.get('form'));
   }),
 
+  // Clicking on the add session button creates a blank record which increases the length of the speaker's list by 1.
+  noSpeakerExists: computed('speakers', function() {
```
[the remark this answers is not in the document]
ritikamotwani: @pradeepgangwar this is not the route. The route is public and we don't create a speaker when the session is created in the public route as the speaker is the logged in user.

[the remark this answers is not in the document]
pradeepgangwar: We can change this later. We need to change a lot of logic in session-speaker-form.js to do this my way. But since this will work for now we can go for this. 
Although that 1 record comes because in above code snippet when e are creating the new record for speaker we also set its relationship with event. And hence it also gets added to our query list here. Hence we always get that record with event although its empty.
https://github.com/fossasia/open-event-frontend/blob/0edc77cabbd31f49cbbb746a1a1f30aeb724d5fd/app/routes/events/view/sessions/create.js#L15-L16

PR #1518, opened 2018-07-29
feat: Extract order-form to a component and add validations

### LINKED ISSUE #1510: Add validations for attendee form builder

Right now the event organizer can have any integer as the order expiry time. time should not be allowed to be negative as it doesn't make sense. Moreover we should also set an upper limit to the expiry time.

**Additional context**

Parent: #1452

srv-twry:
#### Checklist

Right now the event organizer can have any integer as the order expiry time. time should not be allowed to be negative as it doesn't make sense. Moreover we should also set an upper limit to the expiry time.

- Extract order-form to a component.
- Add validations.

Fixes #1510

### DIFF
--- a/app/components/forms/events/view/order-form.js
+++ b/app/components/forms/events/view/order-form.js
@@ -1,0 +1,30 @@
+import Component from '@ember/component';
+import FormMixin from 'open-event-frontend/mixins/form';
+
+export default Component.extend(FormMixin, {
+  getValidationRules() {
+    return {
+      inline : true,
+      delay  : false,
+      on     : 'blur',
+      fields : {
+        orderExpiryTime: {
+          identifier : 'orderExpiryTime',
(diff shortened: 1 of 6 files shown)


app/components/forms/events/view/order-form.js
```diff
@@ -0,0 +1,30 @@
...
+            }
+          ]
+        }
+      }
+    };
+  },
+  actions: {
+    submit(data) {
+      this.onValid(() => {
+        this.sendAction('save', data);
```
CosmicCoder96: instead of this use closure actions like in rest of the app.

CosmicCoder96: @srv-twry hmm, maybe open an issue and change them all, `this.sendAction` is really not required  if we are following DDAU approach like in most of the ember tables.

PR #1661, opened 2018-09-16
Revert "Update ember-cli-babel to the latest version 🚀"

CosmicCoder96:
Reverts fossasia/open-event-frontend#1641


manishekaneja:
@CosmicCoder96 It is solving the problem but I just want to mention that the latest version I can found out is 6.17.1. If any issue is faced in 6.17.1 also can you share it ??

PR #1879, opened 2019-01-17
Generate translation files for l10n

### LINKED ISSUE #1806: Languages are missing and empty languages
In https://eventyay.com/admin/content/translations other languages are missing.

Nevertheless unavailable languages are also showing up on the bottom dropdown, but they don't work and it jumps back to English immediately.

screenshot from 2019-01-14 21-50-34

CosmicCoder96:
Generates all the missing translation files while deploying.

Fixes #1806

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ COPY package.json yarn.lock ./
 RUN yarn install
 
 COPY . .
-RUN node scripts/l10n.js
+RUN node scripts/l10n.js generate
 RUN touch .env && JOBS=1 npx ember build -prod
 
 ##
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,13 +3,14 @@ LABEL maintainer="Niranjan Rajendran <me@niranjan.io>"
...
 
-RUN apk add git python-dev make g++
+RUN apk add git python-dev make g++ gettext
 
 COPY package.json yarn.lock ./
 
 RUN yarn install
 
 COPY . .
+RUN node scripts/l10n.js
```
iamareebjamal:
```suggestion
RUN node scripts/l10n.js generate
```

PR #1909, opened 2019-01-19
Order summary: Add discount to discounted tickets only

CosmicCoder96:
Currently, on the orders summary page, all the tickets amount is reduced even if the discount was applied to only a few of them. Although the final order of the amount is determined correctly, the break up does not add up due to incorrectly applied discounts.


mariobehling:
Good catch!

PR #1788, opened 2019-01-10
Fix Event Imports

YogeshSharma0201:
#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The acceptance, integration, unit tests and linter pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

#### Short description of what this resolves:

Event Import is working now and previous import jobs are being displayed.
Only zip files can be used for importing as of now and other importing methods are not implemented on backend. 

#### Changes proposed in this pull request:

- Add model for import jobs
- Added neccessary actions and function to handle import 

import

Fixes #767
Backend Changes PR: fossasia/open-event-server#5514

PR #1945, opened 2019-01-22
Fix import Event 

YogeshSharma0201:
#### Checklist

This PR is subpart of PR #1788, for easier testing and review.
Backend Changes PR fossasia/open-event-server/pull/5531


app/components/events/event-import-section.js
```diff
@@ -28,7 +28,9 @@ export default Component.extend(FormMixin, {
   },
   actions: {
     submit() {
+      let _this = this;
       this.onValid(() => {
+        _this.sendAction('uploadFile', event.target.getElementsByTagName('input')[0].files);
```
CosmicCoder96: don't use`sendAction` use ember closure actions.

app/controllers/events/import.js
```diff
@@ -0,0 +1,65 @@
...
+      ext = ext[ext.length - 1].toLowerCase();
+      if (ext === 'xml') {
+        endpoint = 'import/pentabarf';
+      } else if (ext === 'ical') {
+        endpoint = 'import/ical';
+      }
+      data.append('file', file);
+
+      this.set('importStatus', 'Uploading file.. Please don\'t close this window');
+      this.set('importError', '');
```
niranjan94: Use [`setProperties`](https://emberjs.com/api/ember/3.5/functions/@ember%2Fobject/setProperties) instead of multiple set statements so that all of them happen within a single property change batch.

app/controllers/events/import.js
```diff
@@ -0,0 +1,65 @@
...
+      this.set('file', true);
+
+      this.get('loader').post(
+        `${`${ENV.APP.apiHost}/${ENV.APP.apiNamespace}/events/`}${endpoint}`,
+        data,
+        { isExternal: true, isFile: true }
+      ).then(data => {
+        _this.importTask(`tasks/${data.task_url.split('/')[3]}`);
+      }).catch(e => {
+        this.set('');
```
niranjan94: Use [`setProperties`](https://emberjs.com/api/ember/3.5/functions/@ember%2Fobject/setProperties) instead of multiple set statements so that all of them happen within a single property change batch.

app/controllers/events/import.js
```diff
@@ -0,0 +1,65 @@
...
+import ENV from 'open-event-frontend/config/environment';
+
+export default Controller.extend({
+  importStatus   : '',
+  importError    : '',
+  btnImportEvent : false,
+  file           : false,
+  fileName       : '',
+  importTask(taskUrl) {
+    let _this = this;
```
niranjan94: This is not required. You can just use `this` within [Arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions). The context is binded correctly.

app/controllers/events/import.js
```diff
@@ -0,0 +1,65 @@
...
+          this.set('');
+          this.set('importError', e.message);
+          this.set('btnImportEvent', false);
+          this.set('file', false);
+        });
+    }, 3000);
+  },
+  actions: {
+    uploadFile(files) {
+      let _this = this;
```
niranjan94: This is not required. You can just use `this` within [Arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions). The context is binded correctly.

app/controllers/events/import.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+      data.append('file', file);
+
+      this.set('importStatus', 'Uploading file.. Please don\'t close this window');
+      this.set('importError', '');
+      this.set('btnImportEvent', true);
+      this.set('file', true);
+
+      this.get('loader').post(
+        `${`${ENV.APP.apiHost}/${ENV.APP.apiNamespace}/events/`}${endpoint}`,
```
niranjan94: Why nested string concats ... ? Can't it be 

```javascript
`${ENV.APP.apiHost}/${ENV.APP.apiNamespace}/events/${endpoint}`
```

And why do you even have to mention `apiHost` and `apiNamespace` here ? The loader service takes care of this automatically as long as you don't pass in `isExternal: true` in the options hash.

app/services/loader.js
```diff
@@ -48,6 +49,9 @@ export default Service.extend({
         } else {
           if (config.isFormData) {
             fetchOptions.body = objectToFormData(data);
+          } else if (config.isFile) {
+            fetchOptions.headers.Authorization = adapter.ajaxOptions().headers.Authorization;
```
niranjan94: Headers already taken care of as long as you don't put `isExternal` in the options hash. So remove this please.

app/controllers/events/import.js
YogeshSharma0201: @niranjan94 i have changed that.

app/controllers/events/import.js
```diff
@@ -0,0 +1,64 @@
...
+        endpoint = 'import/pentabarf';
+      } else if (ext === 'ical') {
+        endpoint = 'import/ical';
+      }
+      data.append('file', file);
+
+      this.setProperties({
+        'importStatus'   : 'Uploading file.. Please don\'t close this window',
+        'importError'    : '',
+        'btnImportEvent' : true,
```
niranjan94: Call this something more clear ... maybe `isUploading` or `isImporting`

app/controllers/events/import.js
```diff
@@ -0,0 +1,64 @@
...
+    run.later(() => {
+      this.get('loader')
+        .load(taskUrl)
+        .then(data => {
+          if (data.state !== 'SUCCESS') {
+            this.set('importStatus', `Status: ${  data.state}`);
+            this.importTask(taskUrl);
+          } else {
+            this.set('importStatus', `Status: ${  data.state}`);
+            document.location = `/events/${  data.result.id}`;
```
niranjan94: Use ember’s transition to route method instead of this

app/controllers/events/import.js
```diff
@@ -0,0 +1,64 @@
...
+    }, 3000);
+  },
+  actions: {
+    uploadFile(files) {
+      let [file] = files;
+      var data = new FormData();
+      var endpoint = 'import/json';
+      var ext = file.name.split('.');
+      ext = ext[ext.length - 1].toLowerCase();
+      if (ext === 'xml') {
```
niranjan94: What about `.zip`, `.xcal` cases ? The upload input accepts those too

YogeshSharma0201: @niranjan94 actually only `.zip` files can be successfully imported, its endpoint is `import/json`. Other cases don't even have a valid endpoint, but eventyay had this code so i thought they can be implemented later on the server side. Should i only accept `.zip` then?

CosmicCoder96: @YogeshSharma0201 add the .zip and .xcal cases, it's fine if server doesn't accept them for now.

YogeshSharma0201:
@CosmicCoder96 i have included .xcal case and .zip is already added, its api is import/json.

PR #2003, opened 2019-01-27
Orders: Make flow similar to eventyay Phase - 1

### LINKED ISSUE #2000: Phase 1: Order flow: Make frontend ready for accepting non registered users as buyers
- Modify the pop up asking non-logged in users to log in or sign up, to instead ask for an email.
- Functionality to handle pre-registered users who are simply not logged in will be handled in a separate issue. (#2001)
- Enable order now button for unverified users.
- Generate a random password, and along with the entered user create a new user on the server.
- Fetch user id from the server, and authenticate the session with him.
- Post successful authentication, persist the session with the new user itself.
- Ensure, that after persisting the current user in the session, the initial session which is about to get invalidated does not cause the app to redirect new user to home page. Keep him on the order page itself.
- Move the `placeOrder` action into the controller of the route where all the controlling logic *should* reside.
**Parent issue:** #1994

CosmicCoder96:
1. Modifies the pop up asking non-logged in users to log in or sign up, to instead ask for an email.
2. Functionality to handle pre-registered users who are simply not logged in will be handled in a separate issue.
3. Enables order now button for unverified users.
4. Generates a random password, and along with the entered user create a new user on the server.
5. Fetches user id from the server, and authenticate the session with him.
6. Posts successful authentication, persist the session with the new user itself.
7. Ensures, that after persisting the current user in the session, the initial session which is about to get invalidated does not cause the app to redirect new user to home page. Keep him on the order page itself.
8. Moves the placeOrder action into the controller of the route where all the controlling logic should reside.

Fixes #2000

### DIFF
--- a/app/components/modals/login-signup-modal.js
+++ b/app/components/modals/login-signup-modal.js
@@ -1,6 +1,39 @@
 import ModalBase from 'open-event-frontend/components/modals/modal-base';
+import FormMixin from 'open-event-frontend/mixins/form';
 
-export default ModalBase.extend({
+export default ModalBase.extend(FormMixin, {
+  getValidationRules() {
+    return {
+      inline : true,
+      delay  : false,
+      on     : 'blur',
+      fields : {
+        identification: {
+          identifier : 'email',
+          rules      : [
+            {
+              type   : 'empty',
+              prompt : this.get('l10n').t('Please enter your email ID')
+            },
+            {
+              type   : 'email',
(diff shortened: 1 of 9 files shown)


kushthedude:
@CosmicCoder96 The user which is getting created with this is gettin stored in this , I would like to recommend that rather than creating a random password and store user data , Like in many apps we see a temporary session is created for the temporary user on the basis of Email-ID provided and after the user complete the order , The session and his details are destroyed after tickets have been sent to him .

PR #2014, opened 2019-01-29
schedule link is only visible when schedule is published

### LINKED ISSUE #1798: Only show "Schedule" link and page when there is a Schedule published
The schedule link and page show up even if there is not schedule published.

screenshot from 2019-01-14 21-23-12

shubham3211:
If  the user has not published the schedule the schedule link would not be visible.

Fixes #1798


shubham3211:
@mariobehling  @CosmicCoder96  Please review.

PR #2104, opened 2019-02-04
Fix date alignment in the session card.

Anupam-dagar:
Fixes: #2074

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The acceptance, integration, unit tests and linter pass locally with my changes 
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

#### Short description of what this resolves:
The orientation of date in sessions card was not proper. This PR fixes the orientation of the date in the sessions card.

#### Changes proposed in this pull request:

- Fixed the alignment of the date in sessions card.

## Before
52085721-30deb000-25cb-11e9-877e-8ca9deaa5204
## After
sessions - test - open event 04-02-2019 16-50-44

PR #2106, opened 2019-02-04
Meaningful message instead of blank field when session is not scheduled.

Anupam-dagar:
#### Checklist

This commit adds a "Session not scheduled yet." message to the session
card when session is not scheduled.

- Changed the if condition for the scheduler block.
- Added an else block to display the message.

Fixes: #2105 #2110 

sessions - test - open event 04-02-2019 17-45-30

Regarding the alignment in the image, it will get resolved automatically once #2104 is merged.

### DIFF
--- a/app/models/session.js
+++ b/app/models/session.js
@@ -4,13 +4,12 @@ import ModelBase from 'open-event-frontend/models/base';
 import { belongsTo, hasMany } from 'ember-data/relationships';
 import { computedDateTimeSplit } from 'open-event-frontend/utils/computed-helpers';
 
-const detectedTimezone = moment.tz.guess();
 
 export default ModelBase.extend({
   title         : attr('string'),
   subtitle      : attr('string'),
-  startsAt      : attr('moment', { defaultValue: () => moment.tz(detectedTimezone).add(1, 'months').startOf('day') }),
-  endsAt        : attr('moment', { defaultValue: () => moment.tz(detectedTimezone).add(1, 'months').hour(17).minute(0) }),
+  startsAt      : attr('moment', { defaultValue: () => null }),
+  endsAt        : attr('moment', { defaultValue: () => null }),
   shortAbstract : attr('string'),
   longAbstract  : attr('string'),
   language      : attr('string'),


app/templates/components/public/session-item.hbs
```diff
@@ -20,9 +20,11 @@
           </div>
         </div>
         <div class="ui right aligned column four wide less left padding">
-          {{#if session.startsAt}}
+          {{#if session.microlocation.name}}
```
CosmicCoder96: Leave this be, rest is fine.

Anupam-dagar: I guess by default there is a time set for the sessions, so by keeping `startsAt` in the if condition, it will always get true and the block will execute. That's why I shifted to microlocation name since on the session getting scheduled, the session will have a location and a time slot for sure.

CosmicCoder96: actually, for unscheduled sessions it is supposed to be null, seems weird that they will have a startTime by default, maybe open an issue for that too. It _should_ be null.

app/templates/components/public/session-item.hbs
```diff
@@ -23,6 +23,8 @@
           {{#if session.startsAt}}
             <div class=""><i class="marker icon"></i>{{session.microlocation.name}}</div>
             <div class=""><i class="wait icon"></i>{{moment-format session.startsAt 'hh:mm a / DD-MM-YYYY'}}</div>
+          {{else}}
+            <div>{{t 'Session not scheduled yet.' }}</div>
```
CosmicCoder96: No empty div's please, they serve no purpose.

Anupam-dagar:
@CosmicCoder96 I have removed the empty div.
I have added another commit as a fix for the issue #2110 which makes the default value for all the submitted session start and end time to `null`.

app/templates/components/public/session-item.hbs
Anupam-dagar: Opened #2110 to set the unscheduled sessions start and end time to null.

PR #2161, opened 2019-02-07
admin can edit all the speaker and non admin speaker can only edit himself

### LINKED ISSUE #2149: Edit speakers not visible properly
**Describe the bug**
Edit speakers should be visible properly. 

**Expected behavior**
Edit speakers should be visible properly and next line should be from the left, the buttons are overlaping

**Screenshots**
edit-speaker

**Additional context**
I am working on this issue.

shubham3211:
edit speaker button moved next to speaker name if the user is admin else if the speaker is not admin he can only edit himself
editspeakers

Fixes #2149

### DIFF
--- a/app/templates/my-sessions/view.hbs
+++ b/app/templates/my-sessions/view.hbs
@@ -7,12 +7,14 @@
       {{/if}}
     {{/link-to}}
     {{#each model.speakers as |speaker|}}
-      {{#link-to 'events.view.speakers.edit' model.event.id speaker.id}}
-        <button class="ui blue button {{if device.isMobile 'fluid' 'right floated'}}">{{t 'Edit Speaker- '}}{{speaker.name}}</button>
-        {{#if device.isMobile}}
-          <div class="ui hidden fitted divider"></div>
-        {{/if}}
-      {{/link-to}}
+      {{#if (eq speaker.email authManager.currentUser.email)}}
+        {{#link-to 'events.view.speakers.edit' model.event.id speaker.id}}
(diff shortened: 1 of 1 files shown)


shubham3211:
@CosmicCoder96  Done :+1:

CosmicCoder96:
@shubham3211 actually I just realized, on my sessions route, these multiple speakers shouldn't be there in the first place, according to eventyay v1, you can only edit yourself as a speaker in a session, editing speakers will fail for a non-admin user, this seems like a bug. Can you instead make it so that the user can edit only himself as a speaker and not others. So only one edit speaker button should be there, and that should be for the user himself.

app/templates/my-sessions/view.hbs
```diff
@@ -7,12 +7,14 @@
       {{/if}}
     {{/link-to}}
     {{#each model.speakers as |speaker|}}
-      {{#link-to 'events.view.speakers.edit' model.event.id speaker.id}}
-        <button class="ui blue button {{if device.isMobile 'fluid' 'right floated'}}">{{t 'Edit Speaker- '}}{{speaker.name}}</button>
-        {{#if device.isMobile}}
-          <div class="ui hidden fitted divider"></div>
-        {{/if}}
-      {{/link-to}}
+      {{#if (and (eq speaker.email authManager.currentUser.email) (not-eq authManager.currentUser.isAnAdmin true))}}
```
CosmicCoder96: Admin doesn't have access to my-sessions route for any given user, so remove the admin part.

ritikamotwani: @shubham3211 any updates?

Issue #2215, opened 2019-02-21
Create account gives unclear message for invalid email


divyam-goel:
**Describe the bug**
While creating account if we enter an invalid email address then the message returned is: An unexpected error occurred. A more helpful message should be shown which helps user understand his/her mistake.

**To Reproduce**
1. Go to '.../register'
2. Enter any invalid email address and hit Register
3. See error

**Expected behavior**
A more helpful message like "Invalid Email address. Please try again with a valid email address." should be returned.

**Screenshots**
![error](https://user-images.githubusercontent.com/32808710/53182887-5c7c1580-3620-11e9-9d4f-8c42c4d804ed.png)

**Desktop:**
 - OS: Ubuntu 16.04
 - Browser chrome

kushthedude:
I would like to take this issue. @divyam-goel

divyam-goel:
Yeah, sure. Go ahead :)

Anupam-dagar:
@kushthedude since there was no WIP PR opened, I took a look at this issue and opened a PR for the same.

kushthedude:
@Anupam-dagar No problem was a bit busy due to college fest

PR #2245, opened 2019-03-02
fix error message for invalid email on register form.

Anupam-dagar:
Fixes: #2215

#### Checklist

This changes the error message shown on entering an invalid email address while registering a user.

- change error message for invalid email address

### DIFF
--- a/app/components/forms/register-form.js
+++ b/app/components/forms/register-form.js
@@ -17,7 +17,8 @@ export default Component.extend(FormMixin, {
           identifier : 'email',
           rules      : [
             {
-              type   : 'email',
+              type   : 'regExp',
+              value  : /\S+@\S+\.\S+/,
               prompt : this.get('l10n').t('Please enter a valid email address')
             }
           ]
--- a/app/controllers/register.js
+++ b/app/controllers/register.js
@@ -28,7 +28,7 @@ export default Controller.extend({
           if (reason && reason.hasOwnProperty('errors') && reason.errors[0].status === 409) {
             this.set('errorMessage', this.get('l10n').t('User already exists.'));
           } else {
-            this.set('errorMessage', this.get('l10n').t('An unexpected error occurred.'));
+            this.set('errorMessage', this.get('l10n').t('Invalid Email address. Please try again with a valid email address.'));
           }
         })
         .finally(() => {


kushthedude:
@Anupam-dagar the solution you have provided still is not proper, as told by @uds5501 The error message should have been appeared like this
screenshot 2019-03-02 at 3 23 03 pm
Rather than a Invalid email
It was just my opinion still @CosmicCoder96 may tell the final review.

Anupam-dagar:
@kushthedude can you steps which you performed beacuse that unexpected error has occured shouldnt have appeared this time. I'll fix the tooltip also.

kushthedude:
Thats just due to my previous login attempt ,not related to the PR.
I was just saying rather than changing the text in the tag, You could have made the tooltip to display whenever user try to login with wrong E-mail.

Anupam-dagar:
@kushthedude the issue refers to the event when the user has submitted the request using `register` button, at that time the current replacement looks good to me rather than the tooltip which is already functional while filling up the form.
Also the use case which you entered is correct in itself, it's not valid email and the tooltip is shown when it proceeds next and on submission the error. Put an input like `testemail@host`, in this case the tooltip isn't shown. But the email is wrong and hence the error to be shown after submission is correct.
Maybe it's good to show the tooltip in that case.

kushthedude:
@Anupam-dagar ,
That was exactly my point,When you enter email like adc@adc , A tootip showing email is invalid should be there rather than a dialog box.
But that was just my suggestion the final review still be dependent on mentors :).

Anupam-dagar:
@CosmicCoder96 I have added a regex for email validation. It accepts any value of the type `string@string.string`.

CosmicCoder96:
um... can you point me to the issue description which this PR solves, also add it as `Fixes #`

Anupam-dagar:
@CosmicCoder96 I changed the position of Fixes at the first line of PR description.
The issue is here https://github.com/fossasia/open-event-frontend/issues/2215

Anupam-dagar:
This means I should do a lookup for the email entered when register button is clicked and if the user exists I should display that the user already exists.

CosmicCoder96:
> This means I should do a lookup for the email entered when register button is clicked and if the user exists I should display that the user already exists.

No, server should return a 404 when a user does not exist and FE should handle it.

> Also is an email of the form string@string considered valid?

Yes. https://stackoverflow.com/questions/20573488/why-does-html5-form-validation-allow-emails-without-a-dot

Anupam-dagar:
I guess if the user email is not in database, then user should be allowed to register, right?

CosmicCoder96:
@Anupam-dagar yes.

Anupam-dagar:
I'll push a commit for that issue also, but for the issue #2215, is changing the message only enough?

kushthedude:
But since ‘string@string’ is considered a valid email than why should it return the changed message.

Anupam-dagar:
So I guess the issue is misleading, the issue shouldn't be regarding incorrect message when invalid email is entered as `string@string` is a valid email.

The main issue arises when the user try to register with an email of the type `string@string`, then the registration process doesn't get completed.
This probably needs a server side fix too, I'll do a thorough research and get back on this.

CosmicCoder96:
Please reopen it when the server side changes, if required are complete.

PR #2391, opened 2019-03-26
Display unverified user message on create event page

### LINKED ISSUE #2312: Warning to verify Email
**Describe the bug**
There should be a global warning after a user registers, if not global, there should be a message in the login form which comes after registering to verify the Email Id. For now, the email verification warning only shows in the user's profile. As unverified Users are not allowed to create an event, I think it should be global.

**Additional context**
Working on it.

prateekj117:
#### Checklist

An unverified User warning is now displayed on the Create Event page as well as unverified Users are not allowed to create an event. 

Fixes #2312

### DIFF
--- a/app/templates/create.hbs
+++ b/app/templates/create.hbs
@@ -1,4 +1,9 @@
 <div class="ui container">
+  <div class="ui stackable grid">
+    <div class="sixteen wide column">
+      {{unverified-user-message}}
+    </div>
+  </div>
   {{#if model.event.name}}
     <h2 class="weight-300">{{t 'Creating'}} <span class="weight-500">{{model.event.name}}</span></h2>
   {{else}}


prateekj117:
@CosmicCoder96 Please review.

app/templates/create.hbs
```diff
@@ -1,4 +1,9 @@
 <div class="ui container">
+  <div class="ui stackable grid">
```
iamareebjamal: This will unconditionally show to the user. Where's the logic here?

prateekj117: @iamareebjamal It's in unverified-user-message file itself.

prateekj117: @iamareebjamal It's in unverified-user-message file itself.

iamareebjamal:
I think the entire div should be visible only if the message is there. @CosmicCoder96 will tell better

prateekj117:
@iamareebjamal That's how it was implemented in the profile page. Also, I tried, the message doesn't display when the user is verified and does not disturb the UI.

prateekj117:
@CosmicCoder96 Ok.

PR #2492, opened 2019-04-06
Adding Check-In and Check-Out message individually

### LINKED ISSUE #2471: Implementation of Exact Message with Specific Pop-up , Whenever a attendee checks-in or checks-out .

- No , not related to a problem , but an enhancement if mentors would like to consider.
- Whenever an attendee checkouts , There should specific pop-up like "Attendee Checkout Successfully" which should be in red dialogue box.
- Whenever an attendee is checked in Dialogue box should say "Attendee checked-in successfully" 

Implementing this enhancement .

**Additional context**

image
Something Like This

kushthedude:
#### Checklist

It adds individual 'Check-In' and 'Check-Out' message whenever attendee changes his status .

- Added a new action 'ToggleCheckOut'
- Changed button name to 'Check-Out' from 'Cancel'

Fixes #2471


kushthedude:
@CosmicCoder96 Review Please

kushthedude:
> * ‘Toggle’ means variable is switched between two possible values - an attendee can either be checkedin or checked out at a given point of time. There is no third state. If you make two functions called toggle check in and toggle check out - it makes no sense and code is basically duplicated. Do this inside the same original function.

Did changes @CosmicCoder96

app/controllers/events/view/tickets/attendees/list.js
```diff
@@ -37,19 +37,29 @@ export default Controller.extend({
...
+        attendee.save()
+          .then(() => {
+            this.notify.success(this.get('l10n').t('Attendee Checked-Out Successfully'));
+            this.send('refreshRoute');
+          })
+          .catch(() => {
+            this.notify.error(this.get('l10n').t('An unexpected error has occurred'));
+          });
+      } else {
+        attendee.toggleProperty('isCheckedIn');
```
CosmicCoder96: what's the point of putting this statement in both if and else blocks. Write it once outside both if it is executed irrespective of the condition being true or false.

kushthedude: If the toggle property i would outside the if , else block and then i need to put attendee.save outside too as the functionality will create error and if i put .save() outside if, else  , Build error would occur due to .then and .catch .
That is why i repeated the .toggleProperty in both if , else .
@CosmicCoder96

CosmicCoder96: > If the toggle property I would outside the if 
I don't know what that means.
>  then i need to put attendee.save outside too
Why would you need to put it "outside too"

kushthedude: @CosmicCoder96 Did the changes

kushthedude:
@CosmicCoder96 review

kushthedude:
> ```js
> ```js
> attendee.toggleProperty('isCheckedIn');
> if(attendee.isCheckedIn) {
>   attendee.set('checkinTimes', newCheckinTimes);
> }
> attendee.save()
>   .then(savedAttendee => {
>     this.notify.success(this.get('l10n').t(`Attendee ${savedAttendee.isCheckedIn ? 'Checked-In' : 'Checked-Out'} Successfully`));
>     this.send('refreshRoute');
>   })
>   .catch(() => {
>     this.notify.error(this.get('l10n').t('An unexpected error has occurred'));
>   });
> ```
> 
> 
> This is what I meant.
> ```

I will implement the ternary operator ASAP

kushthedude:
@CosmicCoder96 Implemented the ternary operator. Review Please

Issue #2345, opened 2019-03-20
No validation on phone number in contact info


Haider8:
**Describe the bug**

There is no validation on phone number, user can enter number of any length and alphabets in the input.
**To Reproduce**
1. Go to 'Settings' in the user dropdown.
2. Click on 'Contact Info' tab.
3. Enter phone number of any length and also alphabets.
4. See error.

**Expected behavior**

It shouldn't accept the invalid inputs from the user.

**Screenshots**

![phone-number-valid](https://user-images.githubusercontent.com/43299408/54676911-cefcea00-4b27-11e9-957d-4609879c8e35.png)

**Additional context**

Working on it

CosmicCoder96:
@Anupam-dagar Proceed.

PR #2542, opened 2019-04-09
Add phone number field validation using regex.

Anupam-dagar:
Reference Link for the regex:
https://stackoverflow.com/a/29835355
https://regexr.com/38pvb

This regex doesn't checks for helpline number like `1800-MY-APPLE`.
I guess we need to assume to some extent the kind of phone number to be entered.

#### Checklist

This adds phone number field validation using regex.

- Add regex for phone number field validation.

Fixes: #2345

### DIFF
--- a/app/components/settings/contact-info-section.js
+++ b/app/components/settings/contact-info-section.js
@@ -1,5 +1,6 @@
 import Component from '@ember/component';
 import FormMixin from 'open-event-frontend/mixins/form';
+import { validPhoneNumber } from 'open-event-frontend/utils/validators';
 
 export default Component.extend(FormMixin, {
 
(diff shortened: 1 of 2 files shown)


app/components/settings/contact-info-section.js
```diff
@@ -30,8 +30,9 @@ export default Component.extend(FormMixin, {
           identifier : 'phone',
           rules      : [
             {
-              type   : 'empty',
-              prompt : this.get('l10n').t('Please enter your phone')
+              type   : 'regExp',
```
ritikamotwani: @Anupam-dagar , I don't think you can remove the empty condition. 
Shouldn't there be two rules like in the email 1 for empty and the other for checking the valid regexp?

Also regexp isn't directly added in the component file.
Check how it's done for validating websites, twitter, facebook handles in session-speakers form
https://github.com/fossasia/open-event-frontend/blob/development/app/components/forms/session-speaker-form.js

Anupam-dagar: I agree with the 2 rules, I was implementing them but did a syntax mistake. Now with reference to the correct file, I got the correct syntax.

Anupam-dagar:
@ritikamotwani I have updated my PR.

Issue #2546, opened 2019-04-09
Password reset shows unexpected error instead of wrong current password.


Anupam-dagar:
**Describe the bug**

While changing password if current password is incorrect, the error `Unexpected error. Password did not change.` shows up. User should be notified `Wrong current password` if wrong password is entered by the user.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to '/settings/password'
2. Enter wrong current password and new passwords
3. Click on save

**Expected behaviour**

Correct error should be shown to user.

**Screenshots**

![Screenshot from 2019-04-09 22-16-12](https://user-images.githubusercontent.com/21174572/55819291-174f6c80-5b16-11e9-8136-8b39ee652bbf.png)

**Additional context**

I am working on this issue.

PR #2547, opened 2019-04-09
Fix current password error in password reset settings.

Anupam-dagar:
#### Checklist

This PR fixes the error shown to the user when incorrect current password is entered during password reset process.

The if block checking for if error is returned by the server was checking for wrong attribute. The attribute received from server is `errors` while `error` was being checked in the frontend code. I fixed the attribute checked in `if` block and displayed the error along with added extra custom error message.

## Screenshots
Screenshot from 2019-04-09 22-16-37

Fixes: #2546

### DIFF
--- a/app/controllers/settings/password.js
+++ b/app/controllers/settings/password.js
@@ -16,8 +16,8 @@ export default Controller.extend({
           this.get('notify').success(this.get('l10n').t('Password updated successfully'));
         })
         .catch(error => {
-          if (error.error) {
-            this.get('notify').error(this.get('l10n').t(error.error));
+          if (error.errors) {
+            this.get('notify').error(this.get('l10n').t(`${error.errors[0].detail}`));
           } else {
             this.get('notify').error(this.get('l10n').t('Unexpected error. Password did not change.'));
           }


app/controllers/settings/password.js
```diff
@@ -16,8 +16,8 @@ export default Controller.extend({
           this.get('notify').success(this.get('l10n').t('Password updated successfully'));
         })
         .catch(error => {
-          if (error.error) {
-            this.get('notify').error(this.get('l10n').t(error.error));
+          if (error.errors) {
```
pradeepgangwar: @Anupam-dagar It may be possible that error is because of reasons other than `wrong current password` e.g. `server not responding` etc. How do you plan to catch those errors and display them correctly.

Anupam-dagar: @pradeepgangwar I guess then it's correct only to show the error received by the server and on the server side we can make this error better. I wrote the latter part just to justify it more clearly.
Should I make a PR on the server side to improve the error message?
I'll update my PR with only showing the error received from the server.

pradeepgangwar: Yes, it looks better to do this on server and fetch error from there.

Anupam-dagar: I have done the required frontend changes, will open a server PR next.

Issue #2670, opened 2019-04-16
Avoid Top progress bar which shows for lazy loading 


mariobehling:
When more items on the page are loaded the top blue bar also shows loading again. This makes the site appear even slower.

![Screenshot from 2019-04-16 13-19-49](https://user-images.githubusercontent.com/1583873/56186787-d04c0480-604a-11e9-9816-dc94fb81a940.png)

mohitm15:
@mariobehling what should be the expected behaviour according to you??

uds5501:
@mariobehling looking into it

kushthedude:
@mariobehling I checked it out and yup thats persisting when the items are loaded , a refresh happens again . @uds5501 If you are not taking it up , May i take it ?

uds5501:
@kushthedude Actually I am working on it

kushthedude:
Cool

uds5501:
@kushthedude this error is not reproduce able on my system. so if you can rectify it, please go ahead and work on it :)

kushthedude:
@uds5501 check on next.eventyay.com once

uds5501:
@kushthedude the issue exists on next.eventyay.com but not on my local repo

kushthedude:
So should i take it @uds5501

uds5501:
@kushthedude go ahead

uds5501:
@CosmicCoder96 Now that lazy loading has been taken out of play from index, does this issue still needs to be addressed?

kushthedude:
@uds5501 there are several more places where this issue is still happening .

uds5501:
@kushthedude where else are we using lazy loading?

kushthedude:
Not about lazy loading but there several places where tables get loaded after getting load once

mariobehling:
Process has already been improved.

PR #2726, opened 2019-04-18
Fix: Remove Extra network requests on initial ember table load

CosmicCoder96:
Makes sure that ember model table doesn't execute it's own query hook, if route model already takes care of it.

### DIFF
--- a/app/components/ui-table-server.js
+++ b/app/components/ui-table-server.js
@@ -13,6 +13,8 @@ export default ModelsTable.extend({
 
   isLoading: false,
 
+  isInitialLoad: true,
+
   isError: false,
 
   metaPagesCountProperty: 'page[size]',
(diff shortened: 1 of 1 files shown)


kushthedude:
I think this fixes #2670 , As tables were reloading after they have been loaded @CosmicCoder96

CosmicCoder96:
@kushthedude Not completely, that issue's scope is all lazy loading calls, ember tables don't have lazy loading.

kushthedude:
On my local repo , there were only issues with ember tables , but on `next.eventyay.com` this issue exist at most of the places and i think the github branch and the deployment is not synced right now .

CosmicCoder96:
On next.eventyay there are 6000 users, 4000 events, 8000 sessions. Load time is negligible on localhost. You can't observe it.

kushthedude:
Oops , i was not knowing next.eventyay is in production, thought we were still using eventyay V.1 😅

kushthedude:
@CosmicCoder96 Ember Tables are still getting a reload after getting loaded completely .

CosmicCoder96:
@kushthedude While changing tabs, yes. That needs to be handled separately. For instance, if you go to Manage events, it won't load twice. However if you switch tabs within My events route, it loads twice while switching them. This PR fixed only a part of it.

kushthedude:
Yup switching tabs causes it to reload twice , Do we need to do similar type of fix for that @CosmicCoder96

CosmicCoder96:
@kushthedude yes.

PR #2742, opened 2019-04-19
Revert "Make URL validation work on sponsors step"

CosmicCoder96:
Reverts fossasia/open-event-frontend#2665


prateekj117:
@CosmicCoder96 @ritikamotwani I think there is a workaround for this. We can use sponsor.identifier as an inputId as done by socialLinks in basic Details Page.

PR #2776, opened 2019-04-23
session is not clickable if there is no speaker

### LINKED ISSUE #2668: Session should not be clickable if it has no speaker
**Describe the bug**
Session is clickable even if it has no speaker

**Expected behaviour**
Session should not be clickable if it has no speaker

**Screenshots**
session

**Additional context**
I am working on it

shubham3211:
session is not clickable if there is no speaker

Fixes #2668

### DIFF
--- a/app/templates/components/public/session-item.hbs
+++ b/app/templates/components/public/session-item.hbs
@@ -68,7 +68,7 @@
           <a href={{speaker.twitter}} target="_blank" rel="noopener noreferrer"><i class="twitter icon"></i>Twitter</a>
         {{/if}}
         {{#if speaker.github}}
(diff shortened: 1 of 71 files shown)


shubham3211:
@CosmicCoder96 @ritikamotwani  @pradeepgangwar please review

app/templates/components/public/session-item.hbs
```diff
@@ -11,7 +11,7 @@
     </div>
     <br>
     <div class="ui grid stackable">
-      {{#if hideImage}}
+      {{#if (and hideImage session.speakers)}}
```
CosmicCoder96: check this while computing `hideImage`, because we are in essence trying to hide the image, is session doesn't have speakers. Keep the least amount of logic in template files.

PR #2778, opened 2019-04-23
Index route: remove function interfering with cache shoe

CosmicCoder96:
This refactor ensures that fastboot shoebox is used correctly.


kushthedude:
@CosmicCoder96 This needs to be reverted, As this is causing the error #2863

PR #2769, opened 2019-04-21
Make event roles inline.

Anupam-dagar:
#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The acceptance, integration, unit tests and linter pass locally with my changes 
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

#### Short description of what this resolves:
This makes the event roles labels inline.

#### Changes proposed in this pull request:

- Make the labels inline.

## Screenshots
Screenshot from 2019-04-21 17-16-35

Fixes: #2768

PR #2790, opened 2019-04-24
Revert "Make event roles inline."

CosmicCoder96:
Reverts fossasia/open-event-frontend#2769


kushthedude:
@CosmicCoder96 What was the issue in it ?

PR #2807, opened 2019-04-25
deleted speakers removed from sessions

### LINKED ISSUE #2667: Deleted speakers are visible in sessions
**Describe the bug**
Deleted speakers are visible in sessions

**Expected behaviour**
Deleted speakers should not be visible in sessions

**Screenshots**
speaker1

**Additional context**
I am working on it

shubham3211:
deleted speakers removed from sessions

Fixes #2667

### DIFF
--- a/app/routes/events/view/sessions/list.js
+++ b/app/routes/events/view/sessions/list.js
@@ -55,7 +55,6 @@ export default Route.extend({
     }
 
     let queryObject = {
-      include      : 'speakers',
       filter       : filterOptions,
       'page[size]' : 10
     };


app/routes/events/view/sessions/list.js
```diff
@@ -55,7 +55,6 @@ export default Route.extend({
     }
 
     let queryObject = {
-      include      : 'speakers',
```
CosmicCoder96: how will this remove only the `deleted` speakers? This will stop including speakers at all. Also, deleted speakers should never be returned by the server in the first place unless `get_trashed=true` is set. Investigate.

uds5501:
@CosmicCoder96 This is a server issue.

PR #2946, opened 2019-05-18
Bring all dependencies inline with each other and stabilize versions

niranjan94:
- Align ember-cli dependencies with each other (3.10.x)
- Switch to ES5 getter for property access where ever possible (single-level prop access)
- Update other dependencies
- Remove deprecated usages of `merge` and `didTransition`
- Replace `ember-browserify` with `ember-auto-import`
- Add ember types for a better IDE experience


package.json
```diff
@@ -19,103 +19,113 @@
...
     "ember-cli-deploy": "^1.0.0",
     "ember-cli-deploy-build": "^1.0.0",
     "ember-cli-deploy-git": "^1.2.0",
     "ember-cli-deploy-html-manifest": "0.0.6",
     "ember-cli-deploy-revision-data": "^1.0.0",
-    "ember-cli-deprecation-workflow": "^1.0.0",
-    "ember-cli-document-title": "^0.4.0",
-    "ember-cli-dotenv": "^2.0.0",
+    "ember-cli-deprecation-workflow": "^1.0.1",
+    "ember-cli-document-title-northm": "^1.0.3",
```
CosmicCoder96: @niranjan94 I replaced this because `ember-cli-document-title` is no longer maintained. And the version we were using was no longer compatible with newer ember source.
https://github.com/kimroen/ember-cli-document-title/issues/83#issuecomment-456837023

niranjan94: Awesome. Did not notice that.. Will change

niranjan94:
@CosmicCoder96 I think we can merge this before more conflicts start coming in, especially given that it touches a lot of files

PR #2948, opened 2019-05-18
fix: Save user details in order confirmation step

### LINKED ISSUE #2945: Orders: Buyer info is missing after successful order
image

uds5501:
Fixes #2945

- add a order.user.save() method to save the user model

### DIFF
--- a/app/controllers/orders/new.js
+++ b/app/controllers/orders/new.js
@@ -6,6 +6,11 @@ export default Controller.extend({
       try {
         this.set('isLoading', true);
         let order = data;
+        let current_user = this.get('authManager.currentUser');
+        let userChanges = current_user.changedAttributes();
+        if (userChanges.firstName || userChanges.lastName){
+          await current_user.save();
+        }
         let { attendees, paymentMode } = data;
         for (const attendee of attendees ? attendees.toArray() : []) {
           await attendee.save();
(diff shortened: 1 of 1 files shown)


app/controllers/orders/new.js
```diff
@@ -6,6 +6,7 @@ export default Controller.extend({
       try {
         this.set('isLoading', true);
         let order = data;
+        await order.user.get('save');
```
CosmicCoder96: `this.get('authManager.currentUser').save()`
and this save should be conditional --> only if the value for the name  was previously null, and now there is a name provided or edited, we make this extra call.

uds5501: @CosmicCoder96 Okay I will test this out

uds5501: @CosmicCoder96 Okay perfect, it works. Pushing the changes soon

uds5501:
@shreyanshdwivedi @CosmicCoder96 Please review now

app/controllers/orders/new.js
```diff
@@ -6,6 +6,12 @@ export default Controller.extend({
       try {
         this.set('isLoading', true);
         let order = data;
+        let current_user = this.get('authManager.currentUser');
```
CosmicCoder96: current_user and order.user are the same things, just use a check for dirtyAttributes for the user model.

uds5501: @CosmicCoder96 on it

uds5501: Completed the required changes

PR #3011, opened 2019-05-25
fix: removes download and invoice buttons from placed orders

uds5501:
Fixes fossasia/open-event-server#5956

Removes the download and invoice buttons from after the discussion here :
https://github.com/fossasia/open-event-server/issues/5956#issuecomment-495939739

### DIFF
--- a/app/templates/orders/view.hbs
+++ b/app/templates/orders/view.hbs
@@ -34,16 +34,18 @@
         </div>
       </div>
       <br>
-      <a class="ui labeled icon {{if isLoadingTickets 'loading'}} button" href="#" {{action downloadTickets model.event.name model.order.identifier}} target="_blank" rel="noopener nofollow">
-        <i class="ticket alternate icon"></i>
-        {{t 'Download Tickets'}}
-      </a>
-      <br>
-      <br>
-      <a href="#" {{action 'downloadInvoice' model.event.name model.order.identifier }} class="ui labeled icon blue {{if isLoadingInvoice 'loading'}} button">
-        <i class="print alternate icon"></i>
-        {{t 'Print Invoice'}}
-      </a>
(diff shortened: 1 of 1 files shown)


uds5501:
@shreyanshdwivedi @mrsaicharan1 Review?

app/templates/orders/view.hbs
```diff
@@ -34,16 +34,18 @@
...
-        <i class="ticket alternate icon"></i>
-        {{t 'Download Tickets'}}
-      </a>
-      <br>
-      <br>
-      <a href="#" {{action 'downloadInvoice' model.event.name model.order.identifier }} class="ui labeled icon blue {{if isLoadingInvoice 'loading'}} button">
-        <i class="print alternate icon"></i>
-        {{t 'Print Invoice'}}
-      </a>
+      {{#if (eq model.order.status 'completed')}}
```
shreyanshdwivedi: @uds5501 I think we are redirecting pending tickets to `/pending` rather than `/view`. Can you check?

shreyanshdwivedi: Ah! @uds5501 I don't see any `Download Tickets` option on the pending page.

uds5501: @shreyanshdwivedi We want download and invoices ONLY in completed orders, correct?

shreyanshdwivedi: Yes @uds5501. We want tickets only for `Completed` or `Placed` orders but for `Pending` tickets we are already not showing any option to download tickets (for reference - screenshot attached above).
`/view` route is accessed only by `Placed` and `Completed` orders

uds5501: Wait a second, we determined that placed orders aren't supposed to be generating tickets or invoices either, correct? @shreyanshdwivedi

PR #3015, opened 2019-05-26
UI: Reduce excessive use of containers :: admin,events,overview

### LINKED ISSUE #697: Tables: Top areas of tables using a lot of space
Tables: Top areas of tables is using a lot of space. Please move number of items and search bar to same row as in previous implementation.

screenshot from 2017-08-13 17-04-04

Previous implementation

screenshot from 2017-08-13 17-05-49

CosmicCoder96:
Partial resolves #697

### DIFF
--- a/app/templates/events/view.hbs
+++ b/app/templates/events/view.hbs
@@ -89,7 +89,7 @@
                               isOpen=isEventDeleteModalOpen
                               eventName=model.name
                               deleteEvent=(action 'deleteEvent')}}
-  <div class="ui {{if isLoading 'loading'}} segment">
+  <div class="ui {{if isLoading 'loading'}} basic segment">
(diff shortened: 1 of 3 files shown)


app/templates/events/view.hbs
```diff
@@ -89,7 +89,7 @@
                               isOpen=isEventDeleteModalOpen
                               eventName=model.name
                               deleteEvent=(action 'deleteEvent')}}
-  <div class="ui {{if isLoading 'loading'}} segment">
+  <div class="ui {{if isLoading 'loading'}} basic segment">
```
kushthedude: @CosmicCoder96 I think this container should be unchanged , As I tried it , Without proper segmented boundaries all section of event are not looking nice

PR #3028, opened 2019-05-28
fix: Refactor buyer name fields in order-form

### LINKED ISSUE #2707: Username changing along with Ticket Buyer Name
**Describe the bug**

When ordering a ticket, the ticket buyer name is changing along with the user name on top

**To Reproduce**
ezgif com-video-to-gif (2)

**Expected behaviour**

The User name shouldn't be changing at all

 - OS: [e.g. iOS] Ubuntu16.04
 - Browser [e.g. chrome, safari] Chrome

**Additional context**

Working on it

uds5501:
- This avoid name changes in the nav bar because the name has been pre loaded

Fixes #2707

This utilises computed property to make sure that user name doesn't change during filling of forms.

#### Checklist

### DIFF
--- a/app/components/forms/orders/order-form.js
+++ b/app/components/forms/orders/order-form.js
@@ -14,14 +14,26 @@ import {
 export default Component.extend(FormMixin, {
   router: service(),
 
+  buyerFirstName: computed('buyerHasFirstName', function() {
+    if (this.buyerHasFirstName) {
+      return this.get('data.user.firstName');
+    } else {
+      return '';
+    }
+  }),
+  buyerLastName: computed('buyerHasLastName', function() {
+    if (this.buyerHasLastName) {
+      return this.get('data.user.lastName');
+    } else {
+      return '';
+    }
+  }),
   buyer: computed('data.user', function() {
     return this.get('data.user');
   }),
-
   buyerHasFirstName: computed(function() {
     return this.get('data.user.firstName');
   }),
-
   buyerHasLastName: computed(function() {
     return this.get('data.user.lastName');
   }),
@@ -474,13 +486,16 @@ export default Component.extend(FormMixin, {
   actions: {
     submit(data) {
       this.onValid(() => {
+        let currentUser = this.get('data.user');
+        currentUser.set('firstName', this.buyerFirstName);
+        currentUser.set('lastName', this.buyerLastName);
         this.sendAction('save', data);
       });
     },
     modifyHolder(holder) {
       if (this.sameAsBuyer) {
-        holder.set('firstname', this.buyer.content.firstName);
-        holder.set('lastname', this.buyer.content.lastName);
+        holder.set('firstname', this.buyerFirstName);
+        holder.set('lastname', this.buyerLastName);
         holder.set('email', this.buyer.content.email);
       } else {
         holder.set('firstname', '');
--- a/app/components/forms/orders/order-form.js
+++ b/app/components/forms/orders/order-form.js
@@ -1,5 +1,6 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { readOnly, oneWay } from '@ember/object/computed';
 import { run } from '@ember/runloop';
 import { inject as service } from '@ember/service';
 import FormMixin from 'open-event-frontend/mixins/form';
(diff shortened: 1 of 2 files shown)


uds5501:
@shreyanshdwivedi @kushthedude @mrsaicharan1 Peer review time!! :smile:

uds5501:
@CosmicCoder96 in that case, this issue shouldn't be valid then, if I am understanding you?

CosmicCoder96:
@uds5501 No this issue is valid, the buyer name changing the username is wrong.

uds5501:
@CosmicCoder96 I think I maybe incorrect in the interpretation of the term *sync* but you mean that it should change and correctly signify the users : 

- When a new user is logged in
- When a profile name is updated
- User logs out 

Correct?
If that is the case then this implementation works locally. Although if there is some other case which I might've not covered, please let me know

app/components/nav-bar.js
```diff
@@ -1,10 +1,22 @@
...
 
 export default Component.extend({
   actions: {
     logout() {
       this.authManager.logout();
       this.routing.transitionTo('index');
     }
-  }
+  },
+  nameShown: computed('authManager.currentUser', function() {
```
niranjan94: Not a valid fix. This should be corrected at the source of the problem which is the order form. 

`authManager.currentUser` should be considered the source of truth for the current logged in user. the navbar and all other UI elements should always reflect what it contains. 

but the order form on the other hand, _must not_ modify `authManager.currentUser` directly when a user input is still happening. update of `authManager.currentUser` should only happen when the the data on the server itself changes. (ie either buy an update request to the server modifying the user data or after a GET request retrieving the latest user data)

uds5501: @niranjan94 Understood, I will try to formulate another approach . I think I can utilise the `buyer` attribute in the order-form component

uds5501:
This is working for me in my local, @shreyanshdwivedi @kushthedude and @mrsaicharan1 could you please verify it ? If it's working fine, then @CosmicCoder96 please review!

app/components/forms/orders/order-form.js
```diff
@@ -14,14 +14,26 @@ import {
 export default Component.extend(FormMixin, {
   router: service(),
 
+  buyerFirstName: computed('buyerHasFirstName', function() {
```
niranjan94: Never mutate a plain computed property with only a getter directly. That is an incorrect usage. Use [`oneWay`](https://api.emberjs.com/ember/3.10/functions/@ember%2Fobject%2Fcomputed/oneWay) here instead. That exists exactly for this purpose.

app/components/forms/orders/order-form.js
```diff
@@ -14,14 +14,26 @@ import {
...
+    if (this.buyerHasFirstName) {
+      return this.get('data.user.firstName');
+    } else {
+      return '';
+    }
+  }),
+  buyerLastName: computed('buyerHasLastName', function() {
```
niranjan94: See comment above

app/components/forms/orders/order-form.js
```diff
@@ -14,14 +14,26 @@ import {
...
+    if (this.buyerHasLastName) {
+      return this.get('data.user.lastName');
+    } else {
+      return '';
+    }
+  }),
   buyer: computed('data.user', function() {
```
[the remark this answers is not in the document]
uds5501: @niranjan94 sure, in this ticket I will handle this file particularly, I shall open another issue for such cases once this is done

app/components/forms/orders/order-form.js
uds5501: @niranjan94 If I am understanding your concern correctly, I should use it in this way, right?
```js
buyerFirstName: oneway('buyerHasFirstName')
```

kushthedude: @uds5501 No ,
```
 buyerFirstName: computed.oneWay('buyerHasFirstName')
```

uds5501:
@kushthedude @niranjan94 please have a look now

uds5501:
@CosmicCoder96 Please review

mrsaicharan1:
@uds5501 Is this working locally? I still see that the name changes in netlify deployment. Can you please check it out? Would be great if you could confirm.

uds5501:
@mrsaicharan1 weird, let me have a look.

uds5501:
@mrsaicharan1 I tried ordering a ticket without logging in. Looks like it's working fine. Can you please pinpoint where is it going wrong?
image
image

mrsaicharan1:
@uds5501 Try logging in and changing the first name and last name while ordering. navbar name changes.

mrsaicharan1:
May-30-2019 13-35-33

uds5501:
@mrsaicharan1 PHEW. Yes, that's something which it's supposed to do. remember this PR? #2948.
This issue pertains to the fact that while typing in "Buyer First Name", the navbar was changing along with it. And from your GIF, this is clear that it works as intended

mrsaicharan1:
> @mrsaicharan1 PHEW. Yes, that's something which it's supposed to do. remember this PR? #2948.
> This issue pertains to the fact that while typing in "Buyer First Name", the navbar was changing along with it. And from your GIF, this is clear that it works as intended

But once you log back in, it still remains. Is that supposed to happen? If it is, then it's fine!

uds5501:
@mrsaicharan1 Yes, it does. Actually, if a user has no name and he's filling up the buyer form, the user name gets updated and hence the navbar is updated as well.

uds5501:
@niranjan94 please review

uds5501:
@CosmicCoder96 @niranjan94 @SaptakS 
The related issues has been added to this week's milestone, please take a look and if there is anything to change, let me know!

uds5501:
@niranjan94 Please review!

PR #3029, opened 2019-05-28
fix: Speakers Call Page fix

mrsaicharan1:
Fixes #3016 

#### Changes proposed in this pull request:

- Changed the access method to `transition.params`

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The acceptance, integration, unit tests and linter pass locally with my changes 
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

PR #3035, opened 2019-05-29
Revert "fix: Speakers Call Page fix"

CosmicCoder96:
Reverts fossasia/open-event-frontend#3029

### DIFF
--- a/app/routes/public/cfs/index.js
+++ b/app/routes/public/cfs/index.js
@@ -6,7 +6,7 @@ export default Route.extend({
   },
 
   async beforeModel(transition) {
-    let hash = transition.params.public.cfs.speaker_call_hash;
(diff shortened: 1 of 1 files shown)


kushthedude:
@CosmicCoder96 Why the following was reverted ?

mrsaicharan1:
> @CosmicCoder96 Why the following was reverted ?

Because of the access method `let hash = transition.to.params['public.cfs'].speaker_call_hash;`

kushthedude:
> > @CosmicCoder96 Why the following was reverted ?
> 
> Because of the access method. The correct way was `let hash = transition.to.params['public.cfs'].speaker_call_hash;`

@mrsaicharan1 But the hash was creating some error and you made the PR to fix that, If I am right?

PR #3036, opened 2019-05-29
Fix CFS : define params

CosmicCoder96:
Access params in CFS Correctly.


mrsaicharan1:
It seems to be working fine now. Looks like I didn't check `transition.to.params`.   
@CosmicCoder96 Was `transition.params` -> `transition.to.params` due to a dependency upgrade?Also, When does this promise get rejected? `transition.to.params['public.cfs']`

CosmicCoder96:
@mrsaicharan1 Yes, due to ember source upgrade, and accessing params does not return a promise, they are objects.

PR #3096, opened 2019-06-10
[WIP] fix: ensure display of correct ticket price on view/attendees

### LINKED ISSUE #2732: Incorrect ticket price is shown on view attendees page
**Describe the bug**

 On the ticket price column, the total price for the order is shown in place of the individual ticket price.

**To Reproduce**
1. Go to 'Create and publish event'
2. Click on 'tickets<<add order'
3. Place order successfully'
4. Go back to tickets<<attendees.
5. See error

**Expected behaviour**

On the ticket price column, individual ticket price should be visible.

**Screenshots**

Screenshot 2019-04-19 06:14:30

 - OS: Ubuntu 18.04
 - Browser: Chrome
 - Version : 71.0398

**Additional context**

Working on it.

shreyanshdwivedi:
Fixes #2732 

On the ticket price column, the total price for the order is shown in place of the individual ticket price.

 - On the ticket price column, individual ticket price is visible.
 - Discount Code is shown on applied ticket only

#### Checklist

### DIFF
--- a/app/helpers/discount-value.js
+++ b/app/helpers/discount-value.js
@@ -1,0 +1,24 @@
+import Helper from '@ember/component/helper';
+
+export async function discountValue(params) {
+  let [attendee] = params;
+  let ticket = await attendee.ticket;
+  let order = await attendee.order;
+  if (ticket && order) {
+    let discountCode = await order.discountCode;
+    if (discountCode) {
+      let discountCodeTickets = await discountCode.tickets;
+      if (discountCodeTickets && discountCodeTickets.includes(ticket)) {
+        let { value } = discountCode;
+        if (discountCode.type === 'amount') {
+          return value;
+        } else {
+          return (ticket.price * value) / 100;
+        }
+      }
+    }
+  }
+  return 0;
+}
+
+export default Helper.helper(discountValue);
--- a/app/templates/components/ui-table/cell/events/view/tickets/attendees/cell-price.hbs
+++ b/app/templates/components/ui-table/cell/events/view/tickets/attendees/cell-price.hbs
@@ -1,8 +1,26 @@
-<div class="weight-400">
-  {{currency-symbol paymentCurrency}}  {{format-number record.order.amount}}
-</div>
 {{#if record.order.discountCode.code}}
-  <div class="ui mini yellow label">
-    {{record.order.discountCode.code}}
+  {{#each record.ticket.discountCodes as |discountCode|}}
+    {{#if (eq discountCode.code record.order.discountCode.code)}}
+      {{#if (eq discountCode.type 'amount')}}
+        <div class="weight-400">
+          {{currency-symbol paymentCurrency}}  {{format-number (sub record.ticket.price discountCode.value)}}
+        </div>
+      {{else}}
+        <div class="weight-400">
+          {{currency-symbol paymentCurrency}}  {{format-number (mult (div (sub 100 discountCode.value) 100) record.ticket.price)}}
+        </div>
+      {{/if}}
+      <div class="ui mini yellow label">
+        {{discountCode.code}}
+      </div>
+    {{/if}}
+  {{else}}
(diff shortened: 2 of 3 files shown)


app/templates/components/ui-table/cell/events/view/tickets/attendees/cell-price.hbs
```diff
@@ -1,8 +1,26 @@
-<div class="weight-400">
-  {{currency-symbol paymentCurrency}}  {{format-number record.order.amount}}
-</div>
 {{#if record.order.discountCode.code}}
-  <div class="ui mini yellow label">
-    {{record.order.discountCode.code}}
+  {{#each record.ticket.discountCodes as |discountCode|}}
+    {{#if (eq discountCode.code record.order.discountCode.code)}}
```
CosmicCoder96: this is too much logic inside a template file, use a helper to get the discount code for an order, always keep minimal logic inside templates.

shreyanshdwivedi: @CosmicCoder96 I tried writing a helper but Idk why the ticket and order object are not returning value of attributes even. Not even basic attributes ( I checked in local by printing console logs).
Where am I doing it wrong?
@uds5501 @mrsaicharan1 @kushthedude maybe you people can help?

shreyanshdwivedi: I'll fix the code and write a helper test once I fix the logic

shreyanshdwivedi: @niranjan94 @CosmicCoder96 can you please tell me where am I doing wrong in the helper I added
> I tried writing a helper but Idk why the ticket and order object are not returning value of attributes even. Not even basic attributes ( I checked in local by printing console logs).

shreyanshdwivedi: @CosmicCoder96 I'm able to get discount value when printed the logs. But the FE shows `[object promise]`. What does this mean?
How can I return resolved value of promise?
Screenshot_2019-06-15 All Attendees Tickets Order Events Open Event

app/helpers/discount-value.js
```diff
@@ -1,19 +1,21 @@
...
-  const { ticket, order } = attendee;
-  if (order.get('discountCode') && ticket.get('discountCodes')) {
-    discountCode = order.get('discountCode');
-    ticket.get('discountCodes').forEach(discount => {
-      if(discount.get('code') === discountCode.get('code')) {
-        if (discount.get('type') === 'amount') {
-          return discount.get('value');
-        } else {
-          return (ticket.get('price') * discount.get('value'))/100;
+export async function discountValue(params) {
```
niranjan94: Don't make helpers async. Also, so much business logic in a helper is a bad idea. Do all this in the controller before you pass the data down to the template.

shreyanshdwivedi: @niranjan94 actually I need to calculate discount for each entry (row). The component `cell-price.js` doesn't return even basic values to the template. 
I don't think we can use controller here as we need to calculate the values corresponding to each row dynamically and controller will be executed only once.

niranjan94: @shreyanshdwivedi the rows are coming in from the route -> controller -> component right ? Why not make sure the rows already come in with the calculated discount amount ? It is possible. 

Never misuse helpers like this. They are not meant for such things.

mariobehling:
@shreyanshdwivedi Thank you for working on this. Unfortunately your approach looks like a dead end. Please open a new PR with a correct fix. The code approach is not usable. It would be better you start fresh. Should be easier after you got the insights in this PR. After several days it should not be a WIP anymore also. Please try a new approach. Closing it therefore.

PR #3137, opened 2019-06-13
feat: toggle filters in mobile devices 

### LINKED ISSUE #3136: Filter removal from explore page in mobile devices
This is a sub issue of #3122 focussing on:
- Toggle Filters for the mobile view as done in Event Brite.

uds5501:
Fixes #3136 

- Add component level logic for hiding/showing filters
- Toggle filters in small devices.

#### Checklist

Two mobile views of the implementation, with and without the filters
#### Without : 
Screenshot from 2019-06-13 18-18-44
#### With :
Screenshot from 2019-06-13 18-18-49

### DIFF
--- a/app/components/explore/side-bar.js
+++ b/app/components/explore/side-bar.js
@@ -9,7 +9,8 @@ export default Component.extend({
 
   customStartDate: moment().toISOString(),
 
-  customEndDate: null,
+  customEndDate : null,
+  showFilters   : false,
 
   hideClearFilters: computed('category', 'sub_category', 'event_type', 'startDate', 'endDate', 'location', function() {
     return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location !== null);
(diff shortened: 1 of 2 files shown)


app/components/explore/side-bar.js
```diff
@@ -19,6 +20,10 @@ export default Component.extend({
     return getDateRanges.bind(this)();
   }),
 
+  showFiltersOnMobile: computed('device.isMobile', 'showFilters', function() {
```
niranjan94: Use [`computed.and`](http://api.emberjs.com/ember/3.10/functions/@ember%2Fobject%2Fcomputed/and) instead

uds5501: actually had to change the logic because the filters weren't showing up on desktop.. Please review now? @niranjan94

app/components/explore/side-bar.js
```diff
@@ -19,6 +20,10 @@ export default Component.extend({
...
+    return ((this.device.isMobile && this.showFilters) || !this.device.isMobile);
```
niranjan94: @uds5501 could this logic be simplified into 

```js
!this.device.isMobile || this.showFilters
```

uds5501:
@kushthedude I think I tried following EventBrite UI as closely as possible without disturbing the basic UI of the code, they went with button so I think it's best we go with button too :sweat_smile: .

Of course, I am prepared to make the change if the mentors agree to that. @CosmicCoder96 @niranjan94 @SaptakS @pradeepgangwar

Issue #3098, opened 2019-06-10
Fix and improve event filters / browse events search


mariobehling:
The page https://eventyay.com/explore offers different ways to search for events. Please simplify the search options as follows:
* Add options show "All Categories" and "All Event Types"
* provide a tick box for filters
* move date filters to first position
* add filter option for Call for Speakers "Open", "Closed" and "Not Available"
* add filter option "Free" or "Paid"

![Screenshot from 2019-06-10 15-20-13](https://user-images.githubusercontent.com/1583873/59199211-b4219b00-8b95-11e9-848a-bedaec54a332.png)

kushthedude:
On it

kushthedude:
@mariobehling Can you please clarify the following, If we need a tickbox to `Show or Hide Filter Section` or the way we are selecting filters

> provide a tick box for filters

Anupam-dagar:
@CosmicCoder96 All parts of this issue has been completed and merged. This issue can be closed now.

PR #3144, opened 2019-06-14
feat: Add CFS filter to sidebar filters in explore route.

Anupam-dagar:
Reference: #3098

This adds a new search filter to sidebar of explore route.

- Add a new filter accordion with `open` and `closed` options in sidebar.
- Add a query for cfs.

#### Checklist

Screenshot from 2019-06-18 22-49-22

### DIFF
--- a/app/components/explore/side-bar.js
+++ b/app/components/explore/side-bar.js
@@ -14,7 +14,7 @@ export default Component.extend({
   showFilters   : false,
 
   hideClearFilters: computed('category', 'sub_category', 'event_type', 'startDate', 'endDate', 'location', 'ticket_type', 'cfs', function() {
-    return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location || this.ticket_type || this.cfs !== null);
+    return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location || this.ticket_type || this.cfs);
   }),
 
   showAllCategories: computed('category', 'sub_category', function() {
--- a/app/components/explore/side-bar.js
+++ b/app/components/explore/side-bar.js
@@ -13,7 +13,7 @@ export default Component.extend({
   customEndDate : null,
   showFilters   : false,
 
-  hideClearFilters: computed('category', 'sub_category', 'event_type', 'startDate', 'endDate', 'location', 'ticket_type', function() {
+  hideClearFilters: computed('category', 'sub_category', 'event_type', 'startDate', 'endDate', 'location', 'ticket_type', 'cfs', function() {
     return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location || this.ticket_type || this.cfs !== null);
   }),
 
@@ -62,7 +62,11 @@ export default Component.extend({
     },
 
     selectEventCfs(cfs) {
-      this.set('cfs', cfs === this.cfs ? null : (cfs === 'open' ? true : false));
+      console.log('incoming param cfs and current this.cfs', cfs, this.cfs);
+      console.log('typeof cfs and typeof this.cfs', typeof(cfs), typeof(this.cfs));
+      this.set('cfs', cfs === this.cfs ? null : cfs);
+      console.log('value of param cfs are this.set and value of this.cfs are this.set', cfs, this.cfs);
+      console.log('typeof cfs and typeof this.cfs after this.set', typeof(cfs), typeof(this.cfs));
     },
 
     selectDateFilter(dateType) {
--- a/app/components/explore/side-bar.js
+++ b/app/components/explore/side-bar.js
@@ -14,7 +14,7 @@ export default Component.extend({
   showFilters   : false,
 
   hideClearFilters: computed('category', 'sub_category', 'event_type', 'startDate', 'endDate', 'location', 'ticket_type', function() {
-    return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location || this.ticket_type !== null);
+    return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location || this.ticket_type || this.cfs !== null);
   }),
 
   showAllCategories: computed('category', 'sub_category', function() {
@@ -59,6 +59,10 @@ export default Component.extend({
       }
 
       this.send('selectDateFilter', 'custom_dates');
+    },
+
+    selectEventCfs(cfs) {
+      this.set('cfs', cfs === this.cfs ? null : (cfs === 'open' ? true : false));
     },
 
     selectDateFilter(dateType) {
--- a/app/components/explore/side-bar.js
+++ b/app/components/explore/side-bar.js
@@ -62,7 +62,7 @@ export default Component.extend({
     },
 
     selectEventCfs(cfs) {
-      this.set('cfs', cfs.toString() === this.cfs ? null : cfs);
+      this.set('cfs', cfs === this.cfs ? null : cfs);
     },
 
     selectDateFilter(dateType) {
--- a/app/components/explore/side-bar.js
+++ b/app/components/explore/side-bar.js
@@ -62,11 +62,7 @@ export default Component.extend({
     },
 
     selectEventCfs(cfs) {
-      console.log('incoming param cfs and current this.cfs', cfs, this.cfs);
-      console.log('typeof cfs and typeof this.cfs', typeof(cfs), typeof(this.cfs));
-      this.set('cfs', cfs === this.cfs ? null : cfs);
-      console.log('value of param cfs are this.set and value of this.cfs are this.set', cfs, this.cfs);
-      console.log('typeof cfs and typeof this.cfs after this.set', typeof(cfs), typeof(this.cfs));
+      this.set('cfs', cfs.toString() === this.cfs ? null : cfs);
     },
 
     selectDateFilter(dateType) {
@@ -136,7 +132,6 @@ export default Component.extend({
     },
     clearFilterTypes() {
       this.set('event_type', null);
-
     },
 
     clearFilters() {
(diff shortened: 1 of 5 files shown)


Anupam-dagar:
I am not quiite sure if we can implement `not available` since the db field is either true or false i.e we either have it open or closed.
Also @uds5501 @shreyanshdwivedi @CosmicCoder96 can you please check it locally. For me, the events with cfs closed were also appearing in the filter of `open` cfs. I expect the query to be correct.
Please review.

app/components/explore/side-bar.js
```diff
@@ -44,6 +44,10 @@ export default Component.extend({
       this.send('selectDateFilter', 'custom_dates');
     },
 
+    selectEventCfs(cfs) {
```
mrsaicharan1: Just out of doubt, What is this property being used for ?

kushthedude:
@Anupam-dagar can you please open a sub issue related to it ?

app/routes/explore.js
```diff
@@ -65,6 +65,18 @@ export default Route.extend({
       });
     }
 
+    if (params.cfs) {
+      let cfsVal = false;
+      if (params.cfs === 'open') {
```
CosmicCoder96: this entire block can be reduced to a single line:
```js
let cfsVal = (params.cfs === 'open' ? true: false);

File: app/components/explore/side-bar.js (hunk 1 above)
Comment (Anupam-dagar) [reply app/components/explore/side-bar.js]: to set `cfs` variable to either `open` or `closed`. Via this, events are filtered,

**Comment by Anupam-dagar** [comment]
> @Anupam-dagar can you please open a sub issue related to it ?

I referenced the issue in the PR as well as commit, I don't know if it will be correct to create a sub issue for it.

**Comment by kushthedude** [comment]
> > @Anupam-dagar can you please open a sub issue related to it ?
> 
> I referenced the issue in the PR as well as commit, I don't know if it will be correct to create a sub issue for it.

Quoted by @CosmicCoder96 
`
 Please don't write 'partially fixes' in front of an issue number, it will still close the whole issue. If you are partially fixing a issue, then the complete issue is a parent issue, and you are fixin a sub issue, create a sub issue, and write fixes # in the PR for it.
`

**Comment by Anupam-dagar** [comment]
> > > @Anupam-dagar can you please open a sub issue related to it ?
> > 
> > 
> > I referenced the issue in the PR as well as commit, I don't know if it will be correct to create a sub issue for it.
> 
> Quoted by @CosmicCoder96
> `Please don't write 'partially fixes' in front of an issue number, it will still close the whole issue. If you are partially fixing a issue, then the complete issue is a parent issue, and you are fixin a sub issue, create a sub issue, and write fixes # in the PR for it.`

I have added referenced, writing reference doesn't close the issue, it just adds a reference there. Partially fixes closes it because it has the term fixes.

File: app/routes/explore.js (hunk 2 above)
Comment (Anupam-dagar) [reply app/routes/explore.js]: @CosmicCoder96 Requested changes completed.

File: app/templates/explore.hbs [hunk 3]
```diff
@@ -35,6 +35,12 @@
           <a role="button" {{action 'clearFilter' 'end_date'}}><i class="icon close"></i></a>
         </div>
       {{/if}}
+      {{#if filters.cfs}}
+        <div class="ui mini label">
+          {{if (eq filters.cfs 'true') 'Open' 'Closed'}}
```
CosmicCoder96: no need to explicitly compare with true.

Anupam-dagar: doing `{{if filters.cfs 'Open' 'Closed'}}` results in only `Open` being set, even on selecting `Closed`.

niranjan94: @Anupam-dagar then it's probably because `filters.cfs` is a string and not a boolean. Ensure it is a boolean.

Anupam-dagar: @niranjan94 I am setting it as boolean only. `this.set('cfs', cfs === this.cfs ? null : (cfs === 'open' ? true : false));`
Also if `filters.cfs` is boolean, then shouldn't the `if` block must not get executed when `closed` is selected since it will become `if false`.

niranjan94: @Anupam-dagar check the data type in the model definition (if one exists). Also check if it is being assigned as a non-boolean anywhere else. Because the only explanation for the behaviour you talk about is if it is a non-boolean.

Anupam-dagar: @niranjan94 I refactored my code, it was being set as boolean but I interpreted it wrong in the router. A little weird thing is happening. Please take a look at the screenshot.
fossasia

Before the first click, no filter is set for cfs. I click on `Open`, the cfs value passed in action as parameter is `true`. Since `cfs` variable for request isn't set, `this.cfs` is `null`. The type here is boolean and Object as expected. I set the variable `cfs` to the param value. Now the types are boolean and boolean as expected.

Now I do the second click on the same filter to unset it, before the `this.set` line, the cfs param and `this.cfs` are both `true` but the types are `boolean` and `string` which should have been `boolean` and `boolean` as `this.cfs` was set to a `boolean` which can be seen from the previous console log. Now after the `this.set` statement, again the cfs is set to `true` which should have been `null` but since the type mismatched it is set to `true`. Finally the types now are both `boolean`.

For the third click, the type for `this.cfs` is boolean and hence I can unset the filter. I am not able to understand how the type is getting changed from `boolean` to `string`.

Also I am still unsure for this code block
```
{{#if filters.cfs}}
        <div class="ui mini label">
          {{if filters.cfs 'Open' 'Closed'}}
          <a role="button" {{action 'clearFilter' 'cfs'}}><i class="icon close"></i></a>
        </div>
{{/if}}
```
I tried after setting to boolean too. I am concerned about the case when `filters.cfs` is `false`, in that case will the `if` block get executed since the value is false. If it doesn't get executed then we won't be able to display the 'Closed' label.

Anupam-dagar: @shreyanshdwivedi @uds5501 @mrsaicharan1 can you please take a look here.

Anupam-dagar:
I have made the changes. I have considered `true` and `false` as string data.

niranjan94:
> I have made the changes. I have considered true and false as string data.

@Anupam-dagar this is the wrong fix. It should remain a string _only_ in the query. Any other usage in code should be a boolean. You should do the necessary conversions for it.

Anupam-dagar:
@niranjan94 I took a look at the changes I made. I got to know that in the routes for sending a request, I need to send `true` or `false` as a string and not as a boolean. Sending it as a boolean doesn't show the correct result.

niranjan94:
@Anupam-dagar see https://guides.emberjs.com/v3.10.0/routing/query-params/#toc_default-values-and-deserialization 

> Query param values are cast to the same datatype as the default value. The same also applies to boolean default values.

Anupam-dagar:
@niranjan94 I tried by setting a default value, it didn't seem to work as in that case I had to make a decision to what type of events I should show by default.
I got rid of the boolean datatype. All the operations now happen within string datatype. The `true` and `false` boolean for query param is decided on the basis of cfs type `Open` or `Closed`.

app/routes/explore.js
```diff
@@ -76,6 +76,14 @@ export default Route.extend({
         val  : `%${params.location}%`
       });
     }
+    if (params.cfs) {
+      let cfsVal = params.cfs === 'Open' ? true : false;
```
niranjan94: Use lower case `open`/`close`

app/routes/explore.js
```diff
@@ -76,6 +76,14 @@ export default Route.extend({
...
+      filterOptions.push({
+        name : 'is_sessions_speakers_enabled',
+        op   : 'eq',
+        val  : cfsVal
```
niranjan94: This can just be `params.cfs === 'open'`

Anupam-dagar:
@niranjan94 This is ready for another review. I guess it's complete now.

app/routes/explore.js
kushthedude: @Anupam-dagar You missed this one!

Anupam-dagar: @kushthedude this has also been fixed, check in changes section.

app/components/explore/side-bar.js
```diff
@@ -13,8 +13,8 @@ export default Component.extend({
   customEndDate : null,
   showFilters   : false,
 
-  hideClearFilters: computed('category', 'sub_category', 'event_type', 'startDate', 'endDate', 'location', 'ticket_type', function() {
-    return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location || this.ticket_type !== null);
+  hideClearFilters: computed('category', 'sub_category', 'event_type', 'startDate', 'endDate', 'location', 'ticket_type', 'cfs', function() {
+    return !(this.category || this.sub_category || this.event_type || this.startDate || this.endDate || this.location || this.ticket_type || this.cfs !== null);
```
CosmicCoder96: if you are passing a string to query param cfs, you can just use `|| this.cfs` here?

app/routes/explore.js
```diff
@@ -76,6 +76,14 @@ export default Route.extend({
         val  : `%${params.location}%`
       });
     }
+    if (params.cfs) {
+      let cfsVal = params.cfs === 'open';
+      filterOptions.push({
+        name : 'is_sessions_speakers_enabled',
+        op   : 'eq',
+        val  : cfsVal
```
CosmicCoder96:
```js
val: params.cfs === 'open'
```

app/components/explore/side-bar.js
Anupam-dagar: the other data type are also string, hence i added it in this way.

CosmicCoder96: this.cfs !== null is true if cfs is not null, false if it is null.
so you can just write this.cfs, no need for the explicit comparison to null.

Anupam-dagar:
@CosmicCoder96 @niranjan94 Updated pull request with the requested changes. Please review.

PR #3150, opened 2019-06-16
feat: Populate searchableLocationName with event location:

### LINKED ISSUE #3149: Populate searchableLocationName while saving event locations
Currently `searchableLocationName` of event model is not used, hence it's not able to refer to events by the city. 

Ideally, it should contain the name of the city.

Sub-issue of #2970

CosmicCoder96:
Fixes #3149 

Ensures that `searchableLocationName` is populated, when event location is being saved. The data from this field will be later used to list the most common cities.

### DIFF
--- a/app/components/widgets/forms/location-input.js
+++ b/app/components/widgets/forms/location-input.js
@@ -17,8 +17,17 @@ export default Component.extend({
     return values(this.address).join(' ').trim();
   }),
 
-  placeNameChanger: observer('combinedAddress', function() {
-    this.set('placeName', this.combinedAddress);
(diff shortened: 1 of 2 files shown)


prateekj117:
@CosmicCoder96 This solution would work for new locations, what about the already saved locations?

CosmicCoder96:
All legacy data already has searchable locations defined in database. Which
leaves 10-20 new events without it. We’ll live with it. Or we can just save
their location again  through the wizard it will reflect it then.

mariobehling:
Yes, we can just re-run the wizard. Thanks. Let's merge this.

PR #3121, opened 2019-06-12
feat: company data possible to be added without ticking 'Send tax invoice'

shreyanshdwivedi:
Fixes #3115 

#### Short description of what this resolves:
The tax invoice data of a registered company can only be added if the field "Send Tax Invoice.." is ticked. 

#### Changes proposed in this pull request:
 - Made it possible to enter the data even if the field is not ticked.

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The acceptance, integration, unit tests and linter pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #3162, opened 2019-06-20
Revert "feat: company data possible to be added without ticking 'Send tax invoice'"

kushthedude:
Reverts fossasia/open-event-frontend#3121

Reason To Revert : 
- Try to type into any of the tax fields which are after `Tax ID`, as soon as you type any letter, An overflow is created.
- UI has been depreciated.

Find a way to tackle the UI OVERFLOW, created by those fields and create new PR, As of now reverting it seems the best possible solution

#### Screenshot : 
image

### DIFF
--- a/app/templates/components/modals/tax-info-modal.hbs
+++ b/app/templates/components/modals/tax-info-modal.hbs
@@ -44,30 +44,32 @@
         </label>
       </div>
     </div>
-    <div class="field">
-      <label class="required">{{t 'Registered Company'}}</label>
-      {{input type='text' id='tax_invoice_company' value=tax.registeredCompany}}
-    </div>
-    <div class="field">
-      <label class="required">{{t 'Business Address'}}</label>
-      {{textarea rows=3 id='tax_invoice_address' value=tax.address}}
-    </div>
-    <div class="field">
-      <label class="required">{{t 'City'}}</label>
-      {{input type='text' id='tax_invoice_city' value=tax.city}}
-    </div>
-    <div class="field">
-      <label class="required">{{t 'State'}}</label>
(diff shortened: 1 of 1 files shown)


kushthedude:
@shreyanshdwivedi Please try this on local

kushthedude:
@CosmicCoder96 @niranjan94 @iamareebjamal Please revert the following, 
Video Link : 
https://drive.google.com/file/d/1QsxEsHpnKw4_QZGiO11Scus4azUT8mof/view?usp=sharing

kushthedude:
@CosmicCoder96 no this pr was not the reason , its due to the ui modal and in my previous pr i encountered the following bug and i resolved it using y-scrollable property ,but then @mariobehling said to avoid internal scroll in modals and due to this overflow is happening

shreyanshdwivedi:
>@CosmicCoder96 no this pr was not the reason
 
@kushthedude then why are you reverting PR? Any specific reason?

kushthedude:
> > @CosmicCoder96 no this pr was not the reason
> 
> @kushthedude then why are you reverting PR? Any specific reason?

@shreyanshdwivedi The following error did bother me a lot , After trying many things, The cause of error was `UI Modal` after some specific height it creates error and one more thing , You can not even alter the height of that modal, So until we find a better solution we should not make this change, An Overflow that too on every click is too irritating .
But still, if you can find some solution for it, @shreyanshdwivedi I would request you to push it.

shreyanshdwivedi:
@kushthedude suppose you revert this PR and then user checks the checkbox. He'll still face the issue right?
I'll suggest to find a better fix to the problem you stated

niranjan94:
@kushthedude reverting is not the solution. Please fix the bug instead.

PR #3163, opened 2019-06-20
Use latest version (Master branch)  for ember-router-scroll

### LINKED ISSUE #275: Misbehaving vertical menu at event detail page
I'm submitting a
- [x] bug report
- [ ] feature request

**Current behavior:**
Menu items - **Info**, **Tickets**, **Speakers** are not configured correctly for clicks. On clicks it just scrolls the content with some fixed length independent of which item is clicked

**Expected behavior:**
On clicks content should get scrolled to the respective section

**Steps to reproduce:**
Click on the one of them (**Info**, **Tickets**, **Speakers**) repetitively. It will just scroll the content.

CosmicCoder96:
Fixes #275 

Uses the latest version (master branch), since the last release was in 2018, a newer one hasn't been released yet.

Please open more specific follow up issues if some buggy scroll behavior is still present.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "ember-qunit": "^4.4.1",
     "ember-resolver": "^5.0.0",
     "ember-route-action-helper": "^2.0.7",
-    "ember-router-scroll": "1.1.0",
+    "ember-router-scroll": "DockYard/ember-router-scroll#695d894",
     "ember-scroll-to": "^0.6.5",
     "ember-simple-auth": "^1.8.2",
     "ember-simple-auth-token": "^3.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6249,9 +6249,9 @@ ember-router-generator@^1.2.3:
   dependencies:
     recast "^0.11.3"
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -100,7 +100,7 @@
     "ember-qunit": "^4.4.1",
     "ember-resolver": "^5.0.0",
     "ember-route-action-helper": "^2.0.7",
-    "ember-router-scroll": "1.1.0",
+    "ember-router-scroll": "DockYard/ember-router-scroll",
```
niranjan94: Use a specific commit `DockYard/ember-router-scroll#695d894`

kushthedude: @niranjan94 If we will use specific commits, Do we need to change it after every commit in the dependancy? Rather specifying Git Repo seems a pretty good option ?

niranjan94: @kushthedude specifying commits lets us have control over the dependency. If you specify a repo+branch, then an accidental rogue commit on the dependency repo can and will break our builds. 

And yes. we will need to change it it periodically to a newer commit after verifying locally that the commits don't cause an issue. But this is only until the package publishes a version to npm.

PR #3169, opened 2019-06-22
Ember Table set up & Manage events table

CosmicCoder96:
@niranjan94 I am in the midst of implementing this, I need feedback on certain conventions.

1 > Is this the right way to reopen and initialize the table?

2> If not, is there any reference for it other than API docs.

3> Actions. What would be a good way to pass them, while keeping the layout general still.

4> Please ignore pagination boiler-plate for now.

5 > Pagination | Should it be inside tfoot? Or outside the table

6 > Search bar | Should it be in thead ? Or outside the table tag
Preview ~so far~:

image


app/templates/components/tables/events/list.hbs
```diff
@@ -0,0 +1,16 @@
+{{#ember-table class="ui table" as |t|}}
```
niranjan94: No need to pass in the class `ui table` here

package.json
```diff
@@ -143,7 +143,8 @@
     "@bower_components/raven-js": "getsentry/raven-js#~3.3",
     "@bower_components/semantic-ui-calendar": "mdehoog/semantic-ui-calendar#^0.0.8",
     "@bower_components/tinyColorPicker": "PitPik/tinyColorPicker#^1.1.1",
-    "@bower_components/wysihtml": "Voog/wysihtml#^0.5.5"
+    "@bower_components/wysihtml": "Voog/wysihtml#^0.5.5",
+    "ember-table": "addepar/ember-table#0aa5637"
```
niranjan94: Move this to devDependencies

niranjan94:
> I am in the midst of implementing this, I need feedback on certain conventions.
> 
> 1 > Is this the right way to reopen and initialize the table?
> 
> 2> If not, is there any reference for it other than API docs.
> 
> 3> Actions. What would be a good way to pass them, while keeping the layout general still.
> 
> 4> Please ignore pagination boiler-plate for now.
> 
> 5 > Pagination | Should it be inside tfoot? Or outside the table
> 
> 6 > Search bar | Should it be in thead ? Or outside the table tag
> Preview so far:
> 

1. What you're doing is fine.
3. Which actions are we talking about ?
4. Sure. Ignored.
5. Outside the table.
6. Outside the table.

CosmicCoder96:
> Which actions are we talking about?

@niranjan94 Actions which are defined inside the controller but are triggered from buttons inside individual cell components. We were using closure actions, so they were passed from the route template to the table. Components directly made use of them. However, they didn't need to be explicitly passed to the cell component, passing them, to the table was enough. It's not true in this case. Passing all the actions to each column seems like overkill.

app/templates/components/tables/events/list.hbs
```diff
@@ -0,0 +1,16 @@
...
+  {{t.head columns=columns}}
```
niranjan94: We don't want to put all of this boilerplate code for every table we want to render. So move this out to a component. And then call that component here with the rows, columns.

niranjan94:
For the actions ...

Let me try something out locally and get back to you.

CosmicCoder96:
Yeah, exactly I had this here for the moment as I was experimenting with
actions. This would be moved to a base component

CosmicCoder96:
@niranjan94 For pagination, should we follow a queryParams approach? Wherein I update the query param with `refreshModel` true so that new data gets loaded into a table?

CosmicCoder96:
@niranjan94 I have completed the pagination functionality, both logic and template, please review that portion. I will work on sorting, search next.

app/components/tables/utilities/page-size-input.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@ember/component';
+import { A } from '@ember/array';
+
+export default Component.extend({
```
niranjan94: Lets start using ES6 classes + decorators for all new resources. That's the way ember is moving towards.

app/components/tables/utilities/page-size-input.js
```diff
@@ -0,0 +1,6 @@
+import Component from '@ember/component';
+import { A } from '@ember/array';
+
+export default class extends Component {
+  pageSizes = A([10, 25, 50, 100, 250, 'All']);
```
niranjan94: `A` not needed. Stick to native arrays whenever possible.

app/components/tables/utilities/search-box.js
```diff
@@ -0,0 +1,4 @@
+import Component from '@ember/component';
+
+export default Component.extend({
```
niranjan94: ES6 class

app/controllers/events/list.js
```diff
@@ -1,86 +1,125 @@
...
+        speakers : row,
+        tickets  : row,
+        url      : row
+      });
+    });
+    return rows;
+  }
+
+  @action
+  doAction(eventName) {
```
niranjan94: Can remove

CosmicCoder96:
@niranjan94 Done with all the modules, please review.

app/components/tables/headers/sort.js
```diff
@@ -0,0 +1,29 @@
...
+        return 'caret up';
+      } else {
+        return 'caret down';
+      }
+    }
+    return 'sort';
+  }
+
+
+  didInsertElement() {
```
niranjan94: Call the super method.

app/components/tables/headers/sort.js
```diff
@@ -0,0 +1,29 @@
...
+    if (this.sorts && this.sorts[0] && this.sorts[0].valuePath === this.column.valuePath) {
+      this.set('sortBy', this.sorts[0].valuePath);
```
niranjan94: use `setProperties`

app/controllers/events/list.js
```diff
@@ -1,86 +1,129 @@
...
+  }
+
+  @action
+  editEvent(id) {
+    this.transitionToRoute('events.view.edit.basic-details', id);
+  }
+
+  @action
+  openDeleteEventModal(id, name) {
+    this.set('isEventDeleteModalOpen', true);
```
niranjan94: use `setProperties`

app/controllers/events/list.js
```diff
@@ -1,86 +1,129 @@
...
+    this.set('eventName', name);
+    this.set('eventId', id);
+  }
+
+  @action
+  async deleteEvent() {
+    this.set('isLoading', true);
+
+    try {
+      const event = await this.store.findRecord('event', this.eventId, { backgroundReload: false });
```
niranjan94: Use peekRecord. I think the event would already be available in the store

CosmicCoder96:
@niranjan94 Thanks for the review, I have made the changes.

app/controllers/events/list.js
```diff
@@ -1,86 +1,131 @@
...
+      eventId                : id
+    });
+  }
+
+  @action
+  async deleteEvent() {
+    this.set('isLoading', true);
+
+    try {
+      const event = await this.store.peekRecord('event', this.eventId, { backgroundReload: false });
```
niranjan94: No await needed for this.

app/controllers/events/list.js
```diff
@@ -1,86 +1,131 @@
...
+      event.destroyRecord();
```
niranjan94: `await` this

CosmicCoder96:
@niranjan94 Made the chages.

kushthedude:
@CosmicCoder96 Can't run it on local, No request is getting sent to the server

niranjan94:
> Can't run it on local, No request is getting sent to the server

@kushthedude Could you give more information ? Error logs, console logs etc ...

CosmicCoder96:
@kushthedude You need to run yarn, packages have been added.

kushthedude:
> @kushthedude You need to run yarn, packages have been added.

@CosmicCoder96 @niranjan94 Yarn Already-up-to-date ,
image

There are no server logs. 
just this error on console 
```js
Uncaught Error: Could not find module `ember-table/components/ember-table/component` imported from `open-event-frontend/extensions/ember-table/component`
```

CosmicCoder96:
@kushthedude it's working fine on netlify PR deploy link, and locally.

kushthedude:
@CosmicCoder96 all the network request is failing on TRAVIS, Is it also accidental?

CosmicCoder96:
@kushthedude the travis failed requests are not related to this PR, might have to do with the new role though, I noticed similar test fails on another PR this morning.

kushthedude:
@CosmicCoder96, Body Over-Flow and Moment support, Are this pending implementation or you will fix them in this only?

app/extensions/ember-table/template.hbs
```diff
@@ -0,0 +1,8 @@
+<table class="ui table">
```
kushthedude: @CosmicCoder96 Can you please also make it unstackable, As it will loose headers in `Mobile View`?

CosmicCoder96:
@kushthedude added the stackable class, what do you mean by moment support?

kushthedude:
It's showing `No Dates Available`

> @kushthedude added the stackable class, what do you mean by moment support?

CosmicCoder96:
@kushthedude Thanks, fixed it.

kushthedude:
@CosmicCoder96 And one more the Text Overflow issue?
image

CosmicCoder96:
@kushthedude wrapped it in a container similar to previous tables, some degree of overflow in internal columns still exists, which can be tackled later, but majority of it is eliminated now.

app/extensions/ember-table/template.hbs
```diff
@@ -0,0 +1,10 @@
+<div class="ui basic less padding segment x-scrollable">
```
kushthedude: @CosmicCoder96 Change it to this, Zero Overflow ! 😄 
```js
<div class="ui segment column sixteen wide inner-table-wrapper x-scrollable">
  <table class="ui very basic unstackable table">
    {{yield (hash
              api=api
              head=(component "ember-thead" api=api)
              body=(component "ember-tbody" api=api)
              foot=(component "ember-tfoot" api=api)
            )}}
  </table>
</div>
```

CosmicCoder96: doesn't fix the core of the problem. It's the non-reducing text. @kushthedude The non basic segment makes it worse.
image

niranjan94: @CosmicCoder96 for x scroll within the table, use the functionality provided by ember-table

kushthedude: @CosmicCoder96 I know it, But previously we handled it by illusions, The very basic table just removes all the lines and the outer segment which is wrapped acts like table border.
And it looks like this now 
image

app/components/tables/headers/sort.js
```diff
@@ -0,0 +1,35 @@
...
+      } else {
+        return 'caret down';
+      }
+    }
+    return 'sort';
+  }
+
+
+  didInsertElement() {
+    this._super(...arguments);
```
niranjan94: Should be 

```js
super.didInsertElement(...arguments)
````

File: app/routes/events/list.js [hunk 17]
```diff
@@ -11,17 +13,21 @@ export default Route.extend({
       case 'past':
         return this.l10n.t('Past');
     }
-  },
+  }
+
   beforeModel(transition) {
     this._super(...arguments);
```
niranjan94: Should be 

```js
super.beforeModel(...arguments)
````

CosmicCoder96:
@niranjan94 Updated.

app/extensions/ember-table/template.hbs
CosmicCoder96: @niranjan94 checking it out.

CosmicCoder96:
@niranjan94 @kushthedude  Actually, `fixed line` class of semantic ui coupled with ember-tables width control properties gives a really decent result. I have pushed the changes, please check it out.

kushthedude:
> @niranjan94 @kushthedude Actually, `fixed single line` class of semantic ui coupled with ember-tables width control properties gives a really decent result. I have pushed the changes, please check it out.

@CosmicCoder96 I can't see the link now, Table is fixed showing an incomplete link
image

CosmicCoder96:
@kushthedude I don't want to make the table `basic` and even then, the ember-table properties are interfering with it, and the column contents are getting squashed into each other.

CosmicCoder96:
@niranjan94 I am confused about the end goal here, do we want a table which never resizes with screen size? and is scrollable?

CosmicCoder96:
@niranjan94 From what I gather, if we don't want to use custom css classes, we want this:https://github.com/Semantic-Org/Semantic-UI/issues/2901 
For the individual columns of the table to not be resizable with screen size, the x-scrollable method is not working with ember-tables, please let me know how we should proceed.

kushthedude:
> @kushthedude I don't want to make the table `basic` and even then, the ember-table properties are interfering with it, and the column contents are getting squashed into each other. Also, innerTableWrapper was defined in ember-model-table theme, that class has no impact here.

I got that, Even the columns are disappearing and collapsing for mobile view, Let's find some new way

kushthedude:
@CosmicCoder96 Isn't there some attribute like `row-span` or `column-span` in the ember-table as we used to have in `ember-models table`?

PR #3178, opened 2019-06-23
fix: Increasing Column Width of Price Section

### LINKED ISSUE #3174: Event Wizard : Donation Ticket Price is not visible due to lack of space
**Describe the bug**

Donation Ticket Price is not visible due to lack of space
**To Reproduce**
1. Go to 'Create Event'
2. Click on 'Donation Ticket'
3. Scroll down to 'Price'

**Expected behaviour**

Price should be visible
**Screenshots**

image

 - OS: MacOS
 - Browser Chrome
 - Version 24

kushthedude:
Fixes #3174 

- The following at least displays ticket price properly in full width of the screen.
- The following is still not a valid solution if width< 786px, For this, we have to make two fields rather than accommodating both inputs in single `Two Fields Div` , But it would break design that's why didn't follow it

image

#### Checklist

### DIFF
--- a/app/templates/components/forms/wizard/basic-details-step.hbs
+++ b/app/templates/components/forms/wizard/basic-details-step.hbs
@@ -159,10 +159,10 @@
         {{#if data.event.tickets}}
           <div class="ui stackable celled five column grid ticket-header">
             <div class="row weight-600">
-              <div class="six wide column">
+              <div class="five wide column">
                 <label class="required">{{t 'Ticket Name'}}</label>
               </div>
-              <div class="two wide column">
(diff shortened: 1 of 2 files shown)


kushthedude:
> > The following is still not a valid solution if width< 786px, For this, we have to make two fields rather than accommodating both inputs in single Two Fields Div , But it would break design that's why didn't follow it
> 
> @kushthedude are you planning to tackle this anyhow?

I had in mind to make them two fields rather than a single field inline , But the requires range is less as after 786 px , Our app will shrink down to mobile view , What you say @shreyanshdwivedi

PR #3228, opened 2019-07-02
enh: Moving Order_Expiry_Time into Admin/Settings

### LINKED ISSUE #3161: Move registration time limit to admin settings and set standard to 15 minutes
Organizers should not have an option to set a time limit on the registration. Please move the time limit to https://eventyay.com/admin/settings and set the standard to 15 minutes. Apply the time limit for all events.

* Keep it out of wizard step 2
* Keep it out of event order forms e.g. here: https://eventyay.com/events/a315169f/tickets/order-form

58038155-d3a44580-7b4c-11e9-8b24-89fc7c5dfd7c

kushthedude:
Fixes #3161 

- Moved all the `Order_Expiry_Time` into `Admin Dashboard`

#### Checklist


app/controllers/admin/settings/index.js
```diff
@@ -8,6 +8,7 @@ export default Controller.extend({
       settings.save()
         .then(() => {
           this.notify.success(this.l10n.t('Settings have been saved successfully.'));
+          this.send('refreshRoute');
```
CosmicCoder96: No, don't do this.

app/templates/components/forms/admin/settings/system/order-expiry-form.hbs
```diff
@@ -0,0 +1,12 @@
+<h3 class="ui header">
+  {{t 'Order Expiry Time'}}
+  <div class="sub header">
+    {{t 'Set registration time limit for your order'}}
+  </div>
+</h3>
+<div class="field">
+  <label class="required">{{t 'REGISTRATION TIME LIMIT'}}</label>
```
CosmicCoder96: don't make it in block letters.

app/templates/components/forms/admin/settings/system/order-expiry-form.hbs
```diff
@@ -0,0 +1,12 @@
...
+  <div class="{{unless device.isMobile 'two wide'}} field">
```
CosmicCoder96: use semantic ui classes to achieve this.

app/templates/components/forms/admin/settings/system/order-expiry-form.hbs
```diff
@@ -0,0 +1,12 @@
+<h3 class="ui header">
+  {{t 'Order Expiry Time'}}
+  <div class="sub header">
+    {{t 'Set registration time limit for your order'}}
```
CosmicCoder96: add (in Minutes)

app/templates/orders/new.hbs
```diff
@@ -28,7 +28,7 @@
     </div>
     <div class="row">
       <div class="ten wide column">
-        {{forms/orders/order-form save='save' data=model.order fields=model.form}}
+        {{forms/orders/order-form save='save' data=model.order fields=model.form settings=settings}}
```
CosmicCoder96: where is setting defined, how does this work?

kushthedude: @CosmicCoder96 It seems to be working, I tried defining `model.setting` over there which was of no good, What would be the ideal way?

CosmicCoder96: @kushthedude it cannot work unless setting model is binded. What is the source? If there is none, then it cannot work by definition, verify it.

kushthedude: I tried using model.setting to bind it but to my surprise it made the counter fixed to 60 min and when I used the settings everything was working as expected with just a refresh. If its not correct , Can you please some other method?

CosmicCoder96: @kushthedude I am not questioning the method, I am asking you where the `settings` object is coming from if it is not defined in the model. Why would `model.setting` ever work if it is not defined in the route model?

kushthedude: @CosmicCoder96 Thanks, Made the appropriate changes.

app/controllers/admin/settings/index.js
kushthedude: Changed it

app/routes/orders/new.js
```diff
@@ -35,7 +35,8 @@ export default Route.extend({
       form  : await eventDetails.query('customForms', {
         'page[size]' : 50,
         sort         : 'id'
-      })
+      }),
+      settings: await this.store.queryRecord('setting', {})
```
niranjan94: @kushthedude why is this needed ? We already have the settings service globally injected into all components/controllers/routes ... You can directly use `settings` from within the component without passing it to it.

kushthedude: @niranjan94 , when I tried to use injected settings , it required a refresh to change the value to latest updated expiry time which needed to define a new action refreshRoute, thats why I injected settings again and it is working fine now

niranjan94: @kushthedude then you'll have to make changes to settings service to ensure it stays updated on changes. As far as settings are concerned, the settings service should be the single point of truth. 

Please refrain from falling back to shortcuts instead of resolving the actual problem.

app/components/forms/admin/settings/system/order-expiry-form.js
```diff
@@ -0,0 +1,4 @@
+import Component from '@ember/component';
+
+export default Component.extend({
```
niranjan94: Lets start moving to ES6 style classes for all new resources. 
https://github.com/fossasia/open-event-frontend/pull/3169/files

kushthedude:
@niranjan94 @CosmicCoder96 Please Review !

app/components/forms/orders/order-form.js
```diff
@@ -37,8 +37,8 @@ export default Component.extend(FormMixin, {
   }),
   sameAsBuyer: false,
 
-  getRemainingTime: computed('data', function() {
-    let orderExpiryTime = this.get('data.event.orderExpiryTime');
+  getRemainingTime: computed('settings', function() {
```
niranjan94: The computed property should be observing everything you are using. `settings.orderExpiryTime`, `data.createdAt` and `data.identifier`.

Please pay more attention to detail

niranjan94: And what is this computer property returning ? I see no return statement at all.

kushthedude: @niranjan94 , Same but it was implemented earlier this way, It's just used to calculate remaining time and this calls timer in it to calculate, But still no return it has

kushthedude: @niranjan94 I tried to return getRemaining but alas it was of no use due to the re-rendering of a rendered value, That's why I think this.set is used to overcome this challenge

niranjan94: This is very incorrect usage of a computed property. And a computed property should also never be reset by something else. Please refactor this correctly.

shreyanshdwivedi:
@kushthedude any update on this?

kushthedude:
I don't know what went wrong, All of my opened PR's are orphan, Will open a new one with a response to it

kushthedude:
Closed in favour of #3283
--

PR #3198, opened 2019-06-26
feat: introduce new columns for session ratings & switch to ember table

shreyanshdwivedi:
Fixes #3195 

#### Short description of what this resolves:
Adds columns - `| Rating | Avg Rating | No. of ratings |`

#### Changes proposed in this pull request:
 - Modified `updateRating` function
 - Introduced 2 new columns - `Avg Rating` and `No. of ratings`

Screenshot_2019-08-01 Confirmed Sessions AddOwnerEvent Events Open Event

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The acceptance, integration, unit tests and linter pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #3246, opened 2019-07-05
feat: switch to ember-table for all remaining tables [Admin Area]

CosmicCoder96:
Fixes #3169 

Replaces all the ember-model tables inside the admin area with ember-tables.

### DIFF
--- a/app/controllers/admin/events/list.js
+++ b/app/controllers/admin/events/list.js
@@ -101,9 +101,9 @@ export default class extends Controller.extend(EmberTableControllerMixin) {
    }
 
   @action
-  moveToDetails(id) {
-    this.transitionToRoute('events.view', id);
-  }
+   moveToDetails(id) {
+     this.transitionToRoute('events.view', id);
+   }
 
   @action
   editEvent(id) {
--- a/app/controllers/admin/events/list.js
+++ b/app/controllers/admin/events/list.js
@@ -1,129 +1,169 @@
 import Controller from '@ember/controller';
-import { computed } from '@ember/object';
-export default Controller.extend({
-  hasRestorePrivileges: computed('authManager.currentUser.isAdmin', function() {
-    return this.get('authManager.currentUser.isSuperAdmin') || this.get('authManager.currentUser.isAdmin');
-  }),
-  columns: [
-    {
-      propertyName : 'name',
-      template     : 'components/ui-table/cell/cell-event',
-      title        : 'Name'
-    },
-    {
-      propertyName : 'startsAt',
-      template     : 'components/ui-table/cell/cell-simple-date',
-      dateFormat   : 'MMMM DD, YYYY - hh:mm A',
-      title        : 'Starts At'
-    },
-    {
-      propertyName : 'endsAt',
-      template     : 'components/ui-table/cell/cell-simple-date',
-      dateFormat   : 'MMMM DD, YYYY - hh:mm A',
-      title        : 'Ends At'
-    },
-    {
-      propertyName : 'state',
-      template     : 'components/ui-table/cell/cell-event-state',
-      title        : 'State'
-    },
-    {
-      propertyName   : 'roles',
-      template       : 'components/ui-table/cell/cell-roles',
-      title          : 'Roles',
-      disableSorting : true
-    },
-    {
-      propertyName   : 'eventStatisticsGeneral.sessions',
-      template       : 'components/ui-table/cell/cell-sessions',
-      title          : 'Sessions',
-      disableSorting : true
-    },
-    {
-      propertyName   : 'eventStatisticsGeneral.speakers',
-      template       : 'components/ui-table/cell/cell-speakers-dashboard',
-      title          : 'Speakers',
-      disableSorting : true
-    },
-    {
-      propertyName   : 'tickets',
-      template       : 'components/ui-table/cell/cell-tickets',
-      title          : 'Tickets',
-      disableSorting : true
-    },
-    {
-      propertyName   : 'url',
-      template       : 'components/ui-table/cell/cell-link',
-      title          : 'Public URL',
-      disableSorting : true
-    },
-    {
-      propertyName   : 'is-featured',
-      template       : 'components/ui-table/cell/admin/event-is-featured',
-      title          : 'Featured Event',
-      disableSorting : false
-    },
-    {
-      propertyName   : 'deletedAt',
-      template       : 'components/ui-table/cell/cell-buttons',
-      title          : 'Actions',
-      disableSorting : true
-    }
-  ],
-  actions: {
-    moveToDetails(id) {
-      this.transitionToRoute('events.view', id);
-    },
-    editEvent(id) {
-      this.transitionToRoute('events.view.edit.basic-details', id);
-    },
-    openDeleteEventModal(id, name) {
-      this.set('isEventDeleteModalOpen', true);
-      this.set('confirmName', '');
-      this.set('eventName', name);
-      this.set('eventId', id);
-    },
-    deleteEvent() {
-      this.set('isLoading', true);
-      this.store.findRecord('event', this.eventId, { backgroundReload: false }).then(function(event) {
-        event.destroyRecord();
-      })
-        .then(() => {
-          this.set('isLoading', false);
-          this.notify.success(this.l10n.t('Event has been deleted successfully.'));
-          this.send('refreshRoute');
-        })
-        .catch(() => {
-          this.set('isLoading', false);
-          this.notify.error(this.l10n.t('An unexpected error has occurred.'));
-        });
-      this.set('isEventDeleteModalOpen', false);
-    },
-    restoreEvent(event) {
-      this.set('isLoading', true);
-      event.set('deletedAt', null);
-      event.save({ adapterOptions: { getTrashed: true } })
-        .then(() => {
-          this.notify.success(this.l10n.t('Event has been restored successfully.'));
-          this.send('refreshRoute');
-        })
-        .catch(e => {
-          this.notify.error(this.l10n.t('An unexpected error has occurred.'));
-          console.warn(e);
-        })
-        .finally(() => {
-          this.set('isLoading', false);
-        });
-    },
-    toggleFeatured(event) {
-      event.toggleProperty('isFeatured');
-      event.save()
-        .then(() => {
-          this.notify.success(this.l10n.t('Event details modified successfully'));
-        })
-        .catch(() => {
-          this.notify.error(this.l10n.t('An unexpected error has occurred.'));
-        });
-    }
+import { computed, action } from '@ember/object';
+import { or } from '@ember/object/computed';
+import EmberTableControllerMixin from 'open-event-frontend/mixins/ember-table-controller';
+
+
+export default class extends Controller.extend(EmberTableControllerMixin) {
+
+   @or('authManager.currentUser.isSuperAdmin', 'authManager.currentUser.isAdmin') hasRestorePrivileges;
+
+   @computed()
+   get columns() {
+     return [
+       {
+         name            : 'Name',
+         valuePath       : 'id',
+         extraValuePaths : ['logoUrl', 'identifier', 'deletedAt', 'name'],
+         isSortable      : true,
+         headerComponent : 'tables/headers/sort',
+         cellComponent   : 'ui-table/cell/cell-event',
+         options         : {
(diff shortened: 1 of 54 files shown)


app/controllers/admin/sessions/list.js
```diff
@@ -1,71 +1,103 @@
...
+      }
+    ];
+  }
+
+  @computed('model.data.[]', 'model.data.@each.event')
+  get rows() {
+    const rows = [];
+    this.model.data.map(row => {
+      rows.pushObject({
+        name        : row.get('event.name'),
```
CosmicCoder96: @niranjan94 
Events which share same session names:
image

app/controllers/admin/events/list.js
```diff
@@ -1,129 +1,178 @@
...
+         }
+       }
+     ];
+   }
+
+  @computed('model.data.[]')
+   get rows() {
+     const rows = [];
+     this.model.data.map(row => {
+       rows.pushObject({
```
niranjan94: Why push to rows ? You can directly map the data and return

app/controllers/admin/events/list.js
```diff
@@ -1,129 +1,178 @@
...
+           toggleFeatured: this.toggleFeatured.bind(this)
+         }
+       }
+     ];
+   }
+
+  @computed('model.data.[]')
+   get rows() {
+     const rows = [];
+     this.model.data.map(row => {
```
niranjan94: Just return the result of map directly. No separate row array required.

app/controllers/admin/events/list.js
```diff
@@ -1,129 +1,178 @@
...
+         width         : 40,
+         actions       : {
+           toggleFeatured: this.toggleFeatured.bind(this)
+         }
+       }
+     ];
+   }
+
+  @computed('model.data.[]')
+   get rows() {
```
niranjan94: Also, what is the purpose of this ? Can't you use the model data directly ?

CosmicCoder96: you mean instead of defininf rows, pass rows=model.data in the template ? @niranjan94

CosmicCoder96:
@niranjan94 It's ready for another review.

app/controllers/admin/events/list.js
niranjan94: Yep.

app/controllers/admin/reports/system-logs/activity-logs.js
```diff
@@ -1,23 +1,45 @@
...
+
+      },
+      {
+        name      : 'Actions',
+        valuePath : 'action'
+      }];
+  }
+
+  @computed('model.[]')
+  get rows() {
```
niranjan94: Use model directly. One less computed property. One less iteration of the dataset.

app/controllers/admin/reports/system-logs/mail-logs.js
```diff
@@ -1,30 +1,53 @@
...
+      {
+        name          : 'Message',
+        valuePath     : 'message',
+        cellComponent : 'ui-table/cell/admin/reports/system-logs/mail-logs/cell-mail-message'
+      }
+    ];
+  }
+
+  @computed('model.[]')
+  get rows() {
```
niranjan94: Use model directly. One less computed property. One less iteration of the dataset.

app/controllers/admin/reports/system-logs/notification-logs.js
```diff
@@ -1,31 +1,49 @@
...
+      {
+        name          : 'Message',
+        valuePath     : 'message',
+        cellComponent : 'ui-table/cell/admin/reports/system-logs/notification-logs/cell-sanitize'
+      }
+    ];
+  }
+
+  @computed('model.[]')
+  get rows() {
```
niranjan94: Use model directly. One less computed property. One less iteration of the dataset.

app/controllers/admin/sessions/list.js
```diff
@@ -1,71 +1,101 @@
...
+          deleteSession : this.deleteSession.bind(this),
+          editSession   : this.editSession.bind(this),
+          viewSession   : this.viewSession.bind(this)
+        }
+      }
+    ];
+  }
+
+  @computed('model.[]', 'model.@each.event')
+  get rows() {
```
niranjan94: Use model directly. One less computed property. One less iteration of the dataset.

app/controllers/admin/users/list.js
```diff
@@ -1,108 +1,134 @@
...
+          dateFormat: 'MMMM DD, YYYY - hh:mm A'
+        }
+
+
+      }
+    ];
+  }
+
+  @computed('model.[]')
+  get rows() {
```
niranjan94: Use model directly. One less computed property. One less iteration of the dataset.

app/controllers/admin/users/view/events/list.js
```diff
@@ -1,82 +1,133 @@
...
+        name          : 'Public URL',
+        valuePath     : 'url',
+        cellComponent : 'ui-table/cell/cell-link',
+        isSortable    : false
+      }
+    ];
+  }
+
+  @computed('model.[]')
+  get rows() {
```
niranjan94: Use model directly. One less computed property. One less iteration of the dataset.

CosmicCoder96:
> Use model directly. One less computed property. One less iteration of the dataset.

@niranjan94 How exactly will that work?
model has the format:

image

where as the mapped array has the format:
image


Ember table expects the formatted mapped array?

niranjan94:
The models have getters for the properties already. So, even though you cannot see the property when you do a console.log, accessing them via for example, `model.email` would work ... And hence would work fine with ember tables too ... Just make sure the `valuePath` is correct for each column

niranjan94:
ember tables uses `get` internally ... So `model.get()` would work fine

CosmicCoder96:
@niranjan94 So I tried with a single column table,
my column definition schema being:
get columns() {
    return [
      {
        name      : 'Email',
        valuePath : 'email' 
     }
    ];
  }

passed rows=model in the template, model is identical to the one in the screenshot above. (user list)

It throws this exception:

```
collapse-tree.js:514 Uncaught (in promise) TypeError: Ember.get(...).some is not a function
    at Class.<anonymous> (collapse-tree.js:514)
    at ComputedProperty.get (metal.js:3200)
    at Object._get2 [as get] (metal.js:1614)
    at Class.<anonymous> (collapse-tree.js:609)
    at ComputedProperty.get (metal.js:3200)
    at _get2 (metal.js:1614)
    at _getPath (metal.js:1646)
    at Object._get2 [as get] (metal.js:1598)
    at Class.<anonymous> (collapse-tree.js:851)
    at ComputedProperty.get (metal.js:3200)
```
Collapsee tree.js:514:
```js
return !Ember.get(this, 'value.children').some(function (child) {
        return Ember.isArray(Ember.get(child, 'children'));
      });
```

What else needs to change?

niranjan94:
@CosmicCoder96 for admin/users/list the model is 

```js
    return this.store.query('user', {
      include      : 'events',
      get_trashed  : true,
      filter       : filterOptions,
      'page[size]' : 10
    });
```

So, you should use `model` not `model.data`

CosmicCoder96:
@niranjan94 I am using model.

niranjan94:
@CosmicCoder96 lemme check this locally ...

niranjan94:
@CosmicCoder96 okay got it. 

[`this.store.query`](https://api.emberjs.com/ember-data/3.10/classes/DS.Store/methods/query?anchor=query) returns a [`RecordArray`](https://api.emberjs.com/ember-data/release/classes/DS.RecordArray). Which is an [`ArrayProxy`](http://api.emberjs.com/ember/release/classes/ArrayProxy) that tries to act like a native javascript array, but fails sometimes. And this is one of those cases where ember-tables attempts to use rows as a native array but is not able to because it isn't. 

To fix it, do something like,

```js
async model(params) {
  return (await this.store.query('user', queryString)).toArray();
}
```

Instead of something like,

```js
model(params) {
  return this.store.query('user', queryString);
}
```

That way, you can use the model directly in the table without having an computed property in between.

_**Bonus!**_

To make this less verbose, you can write a wrapper function that does this for you. 

```js
/**
 * Convert the given ArrayProxy or Promise<ArrayProxy> to a native Array.
 * 
 * @param promise
 * @return {Promise<Array>}
 */
async function asArray(promise) {
  const resolved = await promise;
  if (resolved.toArray) {
    return resolved.toArray();
  }
  return resolved;
}
```

And use it as,

```js
model(params) {
  return asArray(this.store.query('user', queryString));
}
```

app/controllers/admin/users/list.js
```diff
@@ -1,108 +1,116 @@
...
+      {
+        name          : 'Status',
+        valuePath     : 'status',
+        cellComponent : 'ui-table/cell/admin/users/cell-status'
+
+
+      },
+      {
+        name          : 'System Roles',
+        valuePath     : null,
```
CosmicCoder96: to pass the entire record @niranjan94

app/mixins/ember-table-route.js
```diff
@@ -43,6 +45,26 @@ export default Mixin.create({
...
+   * @return {Promise<Array>}
+   */
+  async asArray(promise) {
+    console.log(promise);
+    const resolved = await promise;
+    console.log(resolved);
+    console.log(resolved.toArray());
+    if (resolved.toArray) {
+      return {
+        data     : resolved.toArray(),
```
CosmicCoder96: did this, as resolved array does not contain any metadata. @niranjan94

app/templates/components/tables/default.hbs
```diff
@@ -33,7 +33,7 @@
           {{#r.cell as |cell column row|}}
             {{#if column.cellComponent}}
               {{#component column.cellComponent
-                           record=(get row column.valuePath)
+                           record=(if column.valuePath (get row column.valuePath) row)
```
CosmicCoder96: pass entire record if valuePath is null.

CosmicCoder96:
@niranjan94 I have just modified the `admin/users` table which uses the new method, is the approach okay, if yes, I will proceed to replicate it elsewhere.

app/controllers/admin/users/list.js
niranjan94: Why entire record ? Couldn't you pass only the properties you need ?

CosmicCoder96: @niranjan94 Okay, I can provide around 10 extra valuePaths for the properties it requires, but for the column which has buttons, I do need the entire record in the cell component because it's action uses record.destroyRecord(). Is that okay, or the action needs to be refactored?

niranjan94: > I can provide around 10 extra valuePaths for the properties it requires

I think that's okay. That way we explicitly show what all data that cell requires.

> I do need the entire record in the cell component because it's action uses

Not at all. You only need an `id`. With that ID you can just do a peekRecord in the action to get the full record from the store. This will be a simple, fast synchronous operation. 

--

Pass in just the data your cells require. Nothing more.

CosmicCoder96:
@niranjan94 This is ready for review, I have used extraValue paths for all properties.

app/controllers/admin/sessions/list.js
```diff
@@ -1,71 +1,87 @@
...
-          this.set('isLoading', false);
-        });
-    },
-    editSession(session_id, event_id) {
-      this.transitionToRoute('events.view.sessions.edit', event_id, session_id);
-    },
-    viewSession(id) {
-      this.transitionToRoute('my-sessions.view', id);
+import { computed, action } from '@ember/object';
+import EmberTableControllerMixin from 'open-event-frontend/mixins/ember-table-controller';
```
shreyanshdwivedi: @CosmicCoder96 actually I'm already refactoring this file in my PR https://github.com/fossasia/open-event-frontend/pull/3198 where I'm also introducing few session rating columns and actions.
Should I continue with refactoring in that PR or just update it to ember-model-table and complete the feature only in that PR and you can then refactore it here?

app/routes/events/view/sessions/list.js
```diff
@@ -1,6 +1,7 @@
 import Route from '@ember/routing/route';
-
-export default Route.extend({
+import { action } from '@ember/object';
```
shreyanshdwivedi: @CosmicCoder96 actually I'm already refactoring this file in my PR #3198 where I'm also introducing few session rating columns and actions. :sweat_smile: 
I'll introduce any changes if missing there.

PR #3264, opened 2019-07-11
feat: make billing fields mandatory for paid orders

### LINKED ISSUE #3231: Make billing information mandatory for paid orders

CosmicCoder96:
Fixes #3231


mariobehling:
This should not have been done. I think we discussed this. This should be an option for organizers to make it mandatory, but it should not be mandatory for all events.

PR #3423, opened 2019-08-15
fix: avoid multiple success message in accounts

### LINKED ISSUE #3279: Account Section : Billing Info updation gives multiple notification on success or error
#### Description: 

Fill up the `Billing Info` in the `Accounts Section` and submit, See multiple notifications for a single click

**Expected behaviour**

Single notification for the single key press

**Screenshots**

Screenshot 2019-07-14 at 11 59 56 AM

CosmicCoder96:
Fixes #3279


kushthedude:
@CosmicCoder96 Isnt it like going backwards? We were moving towards ES6 syntax?

CosmicCoder96:
In the long term yes, form mixin needs to be rewritten, then we can come
back to this. It triggers validations twice with es6 sysntax. It is
difficult to say at this point if it’s due to a problem with the mixin, or
a bug in Ember itself since they just released it.

PR #3432, opened 2019-08-17
fix: event settings visible to admin and owner only

### LINKED ISSUE #3354: Event Dashboard: Show `Settings` only to `Event Owners`

Currently, Event Role of every kind is able to see the following settings which should not be the case.

Hide the tab for the non-admin users.

**Additional context**

I will work on it

shreyanshdwivedi:
Fixes #3354 

Currently, Event Role of every kind is able to see the following settings which should not be the case

- show `Settings` as an option only to admin or owner
- if a non-admin and non-owner user tries to access `Settings`, redirect them to event.view

#### Checklist


app/routes/events/view/settings.js
```diff
@@ -11,5 +11,13 @@ export default Route.extend({
       roleInvites : await eventDetails.query('roleInvites', {}),
       roles       : await this.store.findAll('role')
     };
+  },
+  afterModel(model) {
```
kushthedude: Is this block necessary? @shreyanshdwivedi

mrsaicharan1: @kushthedude Yes. An `afterModel` hook is used for such transitions and this is why the PR was actually proposed. Reading the docs would be helpful https://guides.emberjs.com/release/routing/redirection/

CosmicCoder96: you need to do this in beforeModel, in afterModel the data you are trying to protect is already loaded

shreyanshdwivedi: @CosmicCoder96 actually there is no sensitive data shown. The Settings has only option to transfer event which this PR blocks to see for the non-admin non-owner user.
Do we still need to shift this to beforeModel?

mrsaicharan1: @shreyanshdwivedi Yeah, so I thought that you required the owner from the model. But as that isn't needed and you can directly use this.authManager.user, beforeModel would do

shreyanshdwivedi: Actually it's better to use beforeModel bcoz no need to hit server if we don't need to. I'll shift it to beforeModel

shreyanshdwivedi:
@CosmicCoder96 @mrsaicharan1 please review

app/routes/events/view/settings.js
```diff
@@ -6,6 +6,10 @@ export default Route.extend({
   },
   async model() {
     let eventDetails = this.modelFor('events.view');
+    let { currentUser } = this.authManager;
+    if (!(currentUser.isAnAdmin || eventDetails.owner.get('email') === currentUser.email)) {
+      this.transitionTo('events.view');
```
mrsaicharan1: Shouldn't this be going inside `beforeModel`

shreyanshdwivedi:
@CosmicCoder96 please reveiew

shreyanshdwivedi:
@CosmicCoder96 please review this.

app/routes/events/view/settings.js
```diff
@@ -4,6 +4,12 @@ export default Route.extend({
   titleToken() {
     return this.l10n.t('Settings');
   },
+  beforeModel() {
+    let { currentUser } = this.authManager;
```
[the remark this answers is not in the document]
iamareebjamal: This is equivalent to `let currentUser = this.authManager.currentuser`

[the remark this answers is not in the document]
shreyanshdwivedi: @CosmicCoder96 please review this

PR #3433, opened 2019-08-17
fix: cfs date toggle for new events

### LINKED ISSUE #3431: validation issue while enabling CFS
**Describe the bug**

While enabling CFS, if one of the field is empty and we click outside, it gives the expected validation error but this error won't go even on filling up the correct values. Also the error message is wrong for CFS end time.

**To Reproduce**
1. Go to '/events/event_id/edit/sessions-speakers'
2. Enable session & speakers
3. Empty the end date value

**Expected behaviour**

Error should go away once it has been resolved. Also the correct error message should be shown.
**Screenshots**

Screenshot from 2019-08-17 16-30-35

Screenshot from 2019-08-17 16-34-55

CosmicCoder96:
Fixes #3431 
Fixes CFS toggle for new events

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -16,6 +16,7 @@ export const fixFilterQuery = query  => {
   if (query.hasOwnProperty('filter')) {
     query.filter = JSON.stringify(query.filter);
   }
+
   return query;
 };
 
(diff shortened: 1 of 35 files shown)


kushthedude:
@CosmicCoder96 Events created after this change are not facing such CFS issues, But old events still get the validation error event after this change

CosmicCoder96:
@kushthedude interesting, so every old event has this error?

PR #3452, opened 2019-08-20
[WIP] fix: form submission on enter key press.

Anupam-dagar:
Fixes: #2362

Currently the event creation form gets submitted on enter key press.

The default behaviour for any form input on enter key press is to submit the form. Even though preventDefault is the default behaviour of ember action and it is explicitly specified too, it doesn't seem to work here. I tried bubbles=false too. But that also doesn't seem to fix the issue. At last individually doing the preventDefault worked. This might not be the correct solution and hence I need reviews on it.
- Add event.preventDefault() to avoid form submission.

#### Checklist

### DIFF
--- a/app/components/widgets/forms/location-input.js
+++ b/app/components/widgets/forms/location-input.js
@@ -30,6 +30,12 @@ export default Component.extend({
 
   }),
 
+  keyDown(event) {
+    if (event.keyCode === 13) {
+      event.preventDefault();
+    }
+  },
+
   actions: {
     showAddressView(show = true) {
       this.set('addressViewIsShown', show);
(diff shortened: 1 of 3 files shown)


mrsaicharan1:
@Anupam-dagar Are you sure that this is not breaking the default behavior? Can you please check?

Anupam-dagar:
I tried 2 ways but they doesn't seem to work. I tried implementing `keydown` method in location input component itself, it detects on every key input in the location input and if the input is enter, then preventDefault executes, but still the form was getting submitted. Then I tried doing the prevent default on key down on auto complete input, that also executes but the preventDefault fails to stop the submission.

kushthedude:
Stale ?!

PR #3455, opened 2019-08-21
fix: Saving stripe client test id

### LINKED ISSUE #3454: Stripe Client ID is not being saved
https://open-event-fe.netlify.com/admin/settings/payment-gateway

Stripe Test Client ID is not getting saved, hence cannot enable stripe

kushthedude:
Fixes #3454

### DIFF
--- a/app/models/setting.js
+++ b/app/models/setting.js
@@ -50,7 +50,7 @@ export default ModelBase.extend({
   stripeClientId             : attr('string'),
   stripeSecretKey            : attr('string'),
   stripePublishableKey       : attr('string'),
-  stripeTestCliendId         : attr('string'),
+  stripeTestClientId         : attr('string'),
   stripeTestSecretKey        : attr('string'),
   stripeTestPublishableKey   : attr('string'),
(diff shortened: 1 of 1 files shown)


kushthedude:
@shreyanshdwivedi @uds5501 Please review

app/models/setting.js
```diff
@@ -50,7 +50,7 @@ export default ModelBase.extend({
   stripeClientId             : attr('string'),
   stripeSecretKey            : attr('string'),
   stripePublishableKey       : attr('string'),
-  stripeTestCliendId         : attr('string'),
```
iamareebjamal: Why this passed in review?

mrsaicharan1: @iamareebjamal Typo mistake.

kushthedude: > Why this passed in review?

Might the PR be of large size, Thats why we would have missed noticing the error.

CosmicCoder96: Why does anything pass in review indeed 😢

PR #3525, opened 2019-09-17
updating documentation typos

SundaramDubey:
updating documentation  "ember serve" to "ember server" for better readability.
Screenshot from 2019-09-17 22-18-03
I think it will give a better understanding who is just getting started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The Open Event Frontend can be easily deployed on a variety of platforms. Detail
 
 ## Running / Development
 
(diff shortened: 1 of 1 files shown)


shreyanshdwivedi:
@SundaramDubey, first of all, you should learn about better practices. You are expected to open an issue first and then work on it if it is really required.
As far as this PR is concerned, the documentation is correct and `ember server` work as well and I don't think to change it to `server` will increase any kind of understanding for newbies.

PR #3488, opened 2019-08-28
fix: Enable sorting in admin users table 

kushthedude:
fixes #3487 

#### Screenshot :
Screenshot 2019-08-29 at 6 06 29 AM

PR #3552, opened 2019-10-11
fix: event sort by name to actually sort by name

### LINKED ISSUE #3499: Event Dashboard: Events are getting sorted by `ID` not their name.
**Describe the bug**

In `Manage Events` and `Admin/Events` , Events are getting sorted by `ID` not their `Name`.

**To Reproduce**

1. Go to 'Manage Events or Admin/Events'
2. Sort using `Name`
3. See error

**Expected behaviour**

Sorting should be done by `Name`

**Screenshots**

Screenshot 2019-08-31 at 10 27 13 AM

JoaoJesus94:
Fixes #3499 
This PR will resolve #3499 

* Event list will be actually sorted by name and not by id

image

#### Checklist

Not Applicable


app/components/tables/headers/sort.js
```diff
@@ -21,7 +21,7 @@ export default class extends Component {
     super.didInsertElement(...arguments);
     if (this.sorts && this.sorts[0] && this.sorts[0].valuePath === this.column.valuePath) {
       this.setProperties({
-        sortBy  : kebabCase(this.sorts[0].valuePath), // Ensures field names are server compatible with sort
+        sortBy  : this.sorts[0].valuePath === 'identifier' ? 'name' : kebabCase(this.sorts[0].valuePath), // Ensures field names are server compatible with sort
```
kushthedude: Why? Event is not only table using the sorting

JoaoJesus94: yes, you're right. The thing is, when I checked I didn't find any other sortable field with valuePath of identifier, so, that would work for that case. But I just fount other one

JoaoJesus94: What I can do is at `ember-table-route.js` like this

image

because, params will only have a `event_state` if it comes from `events/` url.

Or

I can pass a new argument with the text `event` to `applySortFilters` at `routes/events/list.js` and verify it at `applySortFilters`

kushthedude: @JoaoJesus94 You can see #3488, Similar kind of change is involved in following PR. Use it as a reference and solve the following. Also changing `ember-table-route` will affect every ember-table we are using in the web-app, and if you will pass a new argument in `routes/events/list.js` it will fail as the sort filters is getting called from `ember-table-route`

JoaoJesus94:
> You are on the right track, You also need to change the actions which are binded with the record.

What actions are you talking about, it seems that all actions are working as expected. There's just one test failing, regarding the event table

Issue #3555, opened 2019-10-15
Invalid requests from frontend


iamareebjamal:
This is a parent issue describing some invalid queries generated from frontend which raise errors on the server

These issues should be fixed on priority. What's stranger is that all URLs are http, not https (maybe because being terminated by cloudflare)

Secondly, all of the URLs are from unauthenticated users

```
Only '=', '!=', 'is_()', 'isnot()', 'is_distinct_from()', 'isnot_distinct_from()' operators can be used with None/True/False
```

1. http://api.eventyay.com/v1/events/1107/attendees?fields%5Bevent%5D=id&fields%5Bticket%5D=id&filter=%5B%7B%22name%22%3A%22pdf-url%22%2C%22op%22%3A%22ilike%22%2C%22val%22%3Anull%7D%5D&include=order%2Cticket%2Cevent&page%5Bnumber%5D=1&page%5Bsize%5D=20

param | value
-- | --
fields[event] | id
fields[ticket] | id
filter | [{"name":"pdf-url","op":"ilike","val":null}]
include | order,ticket,event
page[number] | 1
page[size] | 20

2. http://api.eventyay.com/v1/events/1107/attendees?fields%5Bevent%5D=id&fields%5Bticket%5D=id&filter=%5B%7B%22name%22%3A%22pdf-url%22%2C%22op%22%3A%22ge%22%2C%22val%22%3Anull%7D%5D&include=order%2Cticket%2Cevent&page%5Bnumber%5D=1&page%5Bsize%5D=20

param | value
-- | --
fields[event] | id
fields[ticket] | id
filter | [{"name":"pdf-url","op":"ge","val":null}]
include | order,ticket,event
page[number] | 1
page[size] | 20

3. http://api.eventyay.com/v1/events/1107/attendees?fields%5Bevent%5D=id&fields%5Bticket%5D=id&filter=%5B%7B%22name%22%3A%22pdf-url%22%2C%22op%22%3A%22notilike%22%2C%22val%22%3Anull%7D%5D&include=order%2Cticket%2Cevent&page%5Bnumber%5D=1&page%5Bsize%5D=20

param | value
-- | --
fields[event] | id
fields[ticket] | id
filter | [{"name":"pdf-url","op":"notilike","val":null}]
include | order,ticket,event
page[number] | 1
page[size] | 20

[...]

PR #3557, opened 2019-10-15
fix: Hotfix for non-rendering paymentCurrencies

### LINKED ISSUE #3556: Public Ticket List: `Unability of rendering of Currency Symbol`
So the following issue is due to the `Unability of rendering of Currency Symbol`

_Originally posted by @kushthedude in https://github.com/fossasia/open-event-frontend/issues/3555#issuecomment-542169132_

kushthedude:
Fixes #3556

### DIFF
--- a/app/templates/components/public/ticket-list.hbs
+++ b/app/templates/components/public/ticket-list.hbs
@@ -1,3 +1,4 @@
+{{!-- Removed `Currency Symbol` helper from the `paymentCurrency` since some currency are not getting rendered. Hotfix until a permanent solution found --}}
 <form class="ui form" {{action on='submit' preventDefault=true}}>
   <table class="ui unstackable very basic compact table" style="margin-bottom: 0">
     <thead class="full-width">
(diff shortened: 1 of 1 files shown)


iamareebjamal:
I can't access the deployment. Add a before/after screenshot?

kushthedude:
After : 
Screenshot 2019-10-16 at 3 22 34 AM

Also @iamareebjamal since I am not sure which currency was used in the following event. It wont be solved until a new release is made.

iamareebjamal:
You can change the API_URL locally to api.eventyay.com and see if it fixes the issue

kushthedude:
> You can change the API_URL locally to api.eventyay.com and see if it fixes the issue

Issue Solved :
Screenshot 2019-10-16 at 4 15 55 AM

Thanks for this nice little trick

kushthedude:
But what's more interesting is that we don't have `IDR` in the `paymentsDictionary` then how was he able to enter IDR into the dropdown?

kushthedude:
@iamareebjamal

iamareebjamal:
Again, most probably because JS validations mean nothing?

kushthedude:
@iamareebjamal Please merge. Also there is a need for new release for FE after this commit.

iamareebjamal:
It's not that big a bug to warrant a release. We need to fix #3555 before

PR #3562, opened 2019-10-16
FIX #2544: Target name gets highlighted

### LINKED ISSUE #2544: When scrolling to a target , The target name should get highlighted .
**Describe the bug**

In the sidebar when user clicks on Scrollable objects and button doesn't highlights their heading

**To Reproduce**

1. View any event
2. Click on *Scrollable object like Tickets or Getting Here* .
3. They dont get highlighted and info is highlighted rather.

**Expected behaviour**

They should be highlighted as 'info'
**Screenshot**

image

 - OS: MacOS Mojave
 - Browser: Chrome
 - Version: 73.0.3

**Additional context**
I am on it

geetchoudhary:
Fixes #2544 

In the sidebar when the user clicks on scrollable links the correct `{{scroll-to}}` gets highlighted.

#### Video:
mp4gifgif

#### Checklist

### DIFF
--- a/app/components/public/side-menu.js
+++ b/app/components/public/side-menu.js
@@ -3,11 +3,17 @@ import { computed } from '@ember/object';
 import moment from 'moment';
 
 export default Component.extend({
+  selectedIndex: undefined,
   async didInsertElement() {
     this._super(...arguments);
     let speakersCall = await this.get('event.speakersCall');
     this.set('shouldShowCallforSpeakers',
       speakersCall && speakersCall.announcement && (speakersCall.privacy === 'public'));
+  },
+  actions: {
+    changeSelectedIndex(index) {
+      this.set('selectedIndex', index);
+    }
   },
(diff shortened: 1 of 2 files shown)


geetchoudhary:
@kushthedude , PTAL. Thanks :smile:

app/components/public/side-menu.js
```diff
@@ -3,12 +3,18 @@ import { computed } from '@ember/object';
 import moment from 'moment';
 
 export default Component.extend({
+  selectedIndex: undefined,
```
kushthedude: Can you make it `index`?

geetchoudhary: We can do this but having  
```
actions: {
    changeIndex(index) {
      this.set('Index', index);
    }
  }
```
will be very confusing for future issues, if any.

CosmicCoder96:
Stale, also we are approaching it the wrong way, maybe scroll to should be replaced with something that works, no point of using a library, and then manually writing an action, which the library was supposed to take itself.

PR #3586, opened 2019-10-31
fix: Increase page[size] to render all form fields.

### LINKED ISSUE #3585: Call for Speakers: Does not show all form fields in Speaker and Session page
In the wizard it is possible to enable/disable form fields for speakers and sessions. The form does not actually show the fields that are enabled in the speakers and session form.

Here is the wizard form. A lot of form fields are enabled, e.g. Github. But: They dont show up in the form of the speaker.

Screenshot from 2019-10-31 11-19-33

Here is the speaker form. Many of the above enabled form fields don't show up.
Screenshot from 2019-10-31 11-22-24

The session form is also missing a lot of form elements.
Screenshot from 2019-10-31 11-22-59

prateekj117:
Fixes #3585 

In total, all custom fields related to an event  (speaker custom fields + session custom fields + attendee custom fields) can have a size of 55, while we only fetch 50, which causes us to lose some fields rendering from server.

- Changed it to 0, to render all custom fields associated with that event.

#### Checklist

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/app/routes/public/cfs/new-speaker.js
+++ b/app/routes/public/cfs/new-speaker.js
@@ -16,7 +16,7 @@ export default Route.extend({
       event : eventDetails,
       forms : await eventDetails.query('customForms', {
         sort         : 'id',
-        'page[size]' : 50
+        'page[size]' : 70
       }),
       speaker: await this.store.createRecord('speaker', {
         email    : currentUser.email,
--- a/app/routes/public/cfs/new-speaker.js
+++ b/app/routes/public/cfs/new-speaker.js
@@ -16,7 +16,7 @@ export default Route.extend({
       event : eventDetails,
       forms : await eventDetails.query('customForms', {
         sort         : 'id',
-        'page[size]' : 70
+        'page[size]' : 0
       }),
       speaker: await this.store.createRecord('speaker', {
         email    : currentUser.email,


prateekj117:
@mariobehling @iamareebjamal Please review.

iamareebjamal:
Change it to 0

kushthedude:
Provide a screenshot without jumbled fields and all fields rendering

prateekj117:
@iamareebjamal Done.

prateekj117:
@kushthedude Check gitter.

iamareebjamal:
It was an urgent issue and a small fix. @kushthedude You already have a lot of issues assigned.

Also, why page[size] 0 doesn't make any sense? I have suggested that.

How is that a temporary fix? What else can be a "correct" solution? The issue simply is that jsonapi limits the results by page size, so it needs to be set to 0 which means don't limit the results

iamareebjamal:
And yes, make the form y-scrollable @prateekj117 if its size is greater than the viewport

kushthedude:
> How is that a temporary fix? What else can be a "correct" solution? The issue simply is that jsonapi limits the results by page size, so it needs to be set to 0 which means don't limit the results

I didnt know about that. Also setting `page[size]:0` would break the inconsistency and if the issue is only because of the insufficient page[size], We can make the form scrollable or wrap it up in a container and make the container scrollable as we have done in tables.
Also if `page[size]:0` is the only solution presently, lets go with it

Also @iamareebjamal you will have to manually fix the order of fields in `session form` as it was done earlier by @CosmicCoder96 .

iamareebjamal:
> setting page[size]:0 would break the inconsistency

Why?

> We can make the form scrollable or wrap it up in a container and make the container scrollable as we have done in tables.

If you do that, and keep page[size] = 50, all fields won't be fetched

> you will have to manually fix the order of fields in session form

Can you please paste the expected order of the fields? And why is it not present in the API/frontend?

prateekj117:
Before:
Screenshot from 2019-11-01 13-31-43

After:
Screenshot from 2019-11-01 13-21-30

iamareebjamal:
Session Form screenshots

prateekj117:
@iamareebjamal Not missing any field in session form.

iamareebjamal:
Issue says that there are missing elements in it

prateekj117:
@iamareebjamal In case of sessions, I see all fields in the form as selected in the screenshots provided by @mariobehling Sir. 
It might be the case that the 5-6 custom fields which were not given by API had `session` type in case of a given event.

PR #3610, opened 2019-11-08
fix: show "more call for speaker" button when appropriate

### LINKED ISSUE #3309: Index Page: Add a condition for display of `Show More Call of Speakers`

No, Displaying of the following `hyperlink` should be when the number of `CFS` is greater than some specific integer

Add a Logical Condition check for the following.

#### Screenshot: 

Since only 1 CFS is there, The page still shows `Show More Call of Speaker`
image

maze-runnar:
Fixes #3309 
**changes it propose**
when there are less than or equal to three the "show more call for speaker " button will not be visible . 
Screenshot from 2019-11-08 09-44-02

#### Checklist

- [ ]  I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -27,11 +27,13 @@
       {{/each}}
     </div>
     <div class="ui hidden divider"></div>
-    <div class="ui centered grid">
-      <div class="row">
-        {{#link-to "explore" (query-params cfs='open') class='ui blue button'}}{{t 'Show more calls for speakers'}}{{/link-to}}
+    {{#if (gt callForSpeakersEvents 3)}}
+      <div class="ui centered grid">
+        <div class="row">
+          {{#link-to "explore" (query-params cfs='open') class='ui blue button'}}{{t 'Show more calls for speakers'}}{{/link-to}}
+        </div>
       </div>
-    </div>
+    {{/if}}
   {{/if}}
 
   {{modals/event-share-modal isOpen=isShareModalOpen event=eventToShare}}
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -27,7 +27,7 @@
       {{/each}}
     </div>
     <div class="ui hidden divider"></div>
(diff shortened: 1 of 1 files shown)


app/templates/index.hbs
```diff
@@ -27,11 +27,13 @@
       {{/each}}
     </div>
     <div class="ui hidden divider"></div>
-    <div class="ui centered grid">
-      <div class="row">
-        {{#link-to "explore" (query-params cfs='open') class='ui blue button'}}{{t 'Show more calls for speakers'}}{{/link-to}}
+    {{#if (gt callForSpeakersEvents 3)}}
```
kushthedude: It's not the correct way `callForSpeakersEvents` returns the queried events not there number. Also, the comparison works here due to the `Ember Arith Helper`. I think a new property `CFSeventsnumber` should be used for comparison, I am also not sure if adding new property will be optimal or not. Hence, I will wait for the review of other mentors.

maze-runnar: @kushthedude  i understand what you are trying to say . So i changed the comparison here from  {{#if (gt callForSpeakersEvents 3)}} to  {{#if (gt callForSpeakersEvents.length 3)}}
now  here is no reason to say it false in my opinion .

maze-runnar: Please check it .

maze-runnar: @uds5501 can you please check it again that it is also working fine ?

maze-runnar:
@kushthedude  can you please check it again , if it working fine ?

mariobehling:
Thanks a lot for this PR. Please help to get some important issues out of the way in this project. We have labeled these issues here: https://github.com/fossasia/open-event-frontend/issues?q=is%3Aissue+is%3Aopen+label%3A%22Priority%3A+High%22 Our goal is to focus on these issues.

maze-runnar:
Ok @mariobehling i will trt to solve these issues .

PR #3615, opened 2019-11-09
fix: max order greater than min order in ticketing system

### LINKED ISSUE #3521: Frontend check for min order and max order in ticketing system
**Describe the bug**
Their should be a frontend side check of minimum order and maximum order in ticketing system while creating any event

**To Reproduce**
1. Go to 'create an event'
2. fill the info
3. Add a ticketing system 
4. Minimum order value has no check of less than 1 and max order doesn't has check of greater than min order

**Expected behaviour**
it should be corrected

**Screenshots**
image

 - OS: [MacOS]
 - Browser [ safari]

**Additional context**
I am taking this issue

codedsun:
Fixes #3521 

Added frontend check of minOrder and max order value in frontend

-
-
-

#### Checklist

### DIFF
--- a/app/components/forms/wizard/basic-details-step.js
+++ b/app/components/forms/wizard/basic-details-step.js
@@ -276,6 +276,10 @@ export default Component.extend(FormMixin, EventWizardMixin, {
             {
               type   : 'number',
               prompt : this.l10n.t('Invalid number')
+            },
+            {
+              type   : 'integer[1..]',
+              prompt : this.l10n.t('Invalid number')
             }
           ]
         },
--- a/app/mixins/event-wizard.js
+++ b/app/mixins/event-wizard.js
@@ -59,7 +59,14 @@ export default Mixin.create(MutableArray, CustomFormMixin, {
       }
     }
     const numberOfTickets = data.tickets ? data.tickets.length : 0;
-    if (event.name && event.locationName && event.startsAtDate && event.endsAtDate && numberOfTickets > 0) {
+    let minMaxValueOfTicketValid = true;
+    for (const ticket of data.tickets ? data.tickets.toArray() : []) {
+      if (!(parseInt(ticket.get('maxOrder')) >= parseInt(ticket.get('minOrder')))) {
+        minMaxValueOfTicketValid = false;
+        break;
+      }
+    }
+    if (event.name && event.locationName && event.startsAtDate && event.endsAtDate && numberOfTickets > 0 && minMaxValueOfTicketValid) {
       await event.save();
 
       for (const ticket of data.tickets ? data.tickets.toArray() : []) {
@@ -125,6 +132,9 @@ export default Mixin.create(MutableArray, CustomFormMixin, {
       }
       if (numberOfTickets === 0) {
         errorObject.errors.push({ 'detail': 'Tickets are required for publishing event' });
+      }
+      if (!minMaxValueOfTicketValid) {
+        errorObject.errors.push({ 'detail': 'Min Order value should be less than equal to Max Order' });
       }
       throw (errorObject);
     }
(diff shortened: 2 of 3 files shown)


codedsun:
@kushthedude @shreyanshdwivedi Please check

codedsun:
@iamareebjamal It has been implemented on server, also need to be implemented on client side

codedsun:
@CosmicCoder96 @kushthedude Please review

codedsun:
@kushthedude @iamareebjamal chck this

app/components/forms/wizard/basic-details-step.js
```diff
@@ -276,6 +276,10 @@ export default Component.extend(FormMixin, EventWizardMixin, {
             {
               type   : 'number',
               prompt : this.l10n.t('Invalid number')
+            },
+            {
```
kushthedude: Not required

app/mixins/event-wizard.js
```diff
@@ -59,7 +59,14 @@ export default Mixin.create(MutableArray, CustomFormMixin, {
...
     const numberOfTickets = data.tickets ? data.tickets.length : 0;
-    if (event.name && event.locationName && event.startsAtDate && event.endsAtDate && numberOfTickets > 0) {
+    let minMaxValueOfTicketValid = true;
+    for (const ticket of data.tickets ? data.tickets.toArray() : []) {
+      if (!(parseInt(ticket.get('maxOrder')) >= parseInt(ticket.get('minOrder')))) {
+        minMaxValueOfTicketValid = false;
+        break;
+      }
+    }
+    if (event.name && event.locationName && event.startsAtDate && event.endsAtDate && numberOfTickets > 0 && minMaxValueOfTicketValid) {
```
kushthedude: Not required in this way , as server will throw error error too for the same. If you can go in way like Semsntic UI validations, then the PR would make sense, as I will have to click submit button to fire up this check

codedsun: > I think the best place to add the check will be where we are getting data to create the event. Add a check just before hitting server. There are many checks implemented similarly.
Validation rules will not be helpful as you are using two different entities here.

It's mentioned in the issue

codedsun: Also, why required to hit the server, when the client side data is provided wrong? It makes sense that always the ticketMaxOrder will be greater than equal to ticketMinOrder We can add the check frontend side

app/components/forms/wizard/basic-details-step.js
codedsun: @kushthedude why?

app/mixins/event-wizard.js
codedsun: @kushthedude

codedsun:
@kushthedude @iamareebjamal Check this

kushthedude:
> I think the best place to add the check will be where we are getting data to create the event. Add a check just before hitting server. There are many checks implemented similarly.

If that's the case, then I don't think we need this check. Since we already have this check on the server. Also, This check only works when data is sent to the server but we indeed already have a similar check for this on the server. If this check can be implemented using `Semantic UI Validation` as has been implemented in case of donation ticket then it would be great to move forward.

kushthedude:
> > I think the best place to add the check will be where we are getting data to create the event. Add a check just before hitting server. There are many checks implemented similarly.
> 
> If that's the case, then I don't think we need this check. Since we already have this check on the server. Also, This check only works when data is sent to the server but we indeed already have a similar check for this on the server. If this check can be implemented using `Semantic UI Validation` as has been implemented in case of donation ticket then it would be great to move forward.

@codedsun

kushthedude:
Reopen when updated .

PR #3637, opened 2019-11-17
fix: Adding scroll to Add event topic in Admin to avoid overflow

### LINKED ISSUE #2689: Add scroll to Add event topic in Admin to avoid overflow.
**Describe the bug**

Currently if we keep adding event topics the div keeps adding it to the bottom. 
**To Reproduce**
1. Go to 'Admin'
2. Click on 'Content'
3. Scroll down to 'Events'

**Expected behaviour**

The div length should be fixed and it should be scrollable after it.
**Screenshots**

Screenshot 2019-04-16 at 8 09 26 PM

**Additional context**

I am working on it.

maze-runnar:
Fixes #2689 
Screenshot from 2019-11-17 12-19-37

#### Checklist


maze-runnar:
I can also use instead of inline css but i was confused that i should make a whole new file or include my class in any previosly placed css file, @kushthedude ?

app/templates/admin/content/events.hbs
```diff
@@ -1,7 +1,7 @@
 <div class="ui stackable grid">
   <div class="row">
     <div class="five wide column">
-      <div class="ui center aligned fluid segment">
+      <div class="ui center aligned fluid segment ">
         <div class="ui grid container">
```
snitin315: Remove the extra space here.

app/templates/admin/content/events.hbs
```diff
@@ -15,31 +15,33 @@
...
-                        {{#ui-popup content=(t 'Edit') class='ui icon button' click=(action 'openModalFor' eventType)}}
-                          <i class="edit icon"></i>
-                        {{/ui-popup}}
-                        {{#ui-popup content=(t 'Delete') class='ui icon button' click=(action (confirm (t 'Are you sure you would like to delete this event type?') (action 'deleteEventProperty' eventType)))}}
-                          <i class="trash outline icon"></i>
-                        {{/ui-popup}}
+        <div  class = "y-scrollable" style = "max-height: 768px;overflow-x: hidden;">
+          <table class="ui celled table">
+            <tbody>
+              {{#each model.eventTypes as |eventType|}}
```
snitin315: Avoid custom css, and I guess using the `y-scrollable` class in in `<table>` should work fine something like this `<table class = "ui celled table y-scrollable">` . Have you tried it?

maze-runnar: > Avoid custom css, and I guess using the `y-scrollable` class in in `<table>` should work fine something like this `<table class = "ui celled table y-scrollable">` . Have you tried it?

I have tried many ways , working on it around 3 days to find a menthid that would not use custom css but max-height is the problem . without using max-height how i am supposed to fix height above which table should be scrollable.

app/templates/admin/content/events.hbs
maze-runnar: Ok .

kushthedude:
I think adding `y-scrollable` to the table class should do the work 😄

maze-runnar:
yes it is working for scrollable but still need an extra div to set max height .
If you can please suggest  alternate of that too :smile:

app/templates/admin/content/events.hbs
```diff
@@ -57,31 +59,33 @@
...
-                    </div>
-                    <div class="eight wide column right aligned">
-                      <div class="ui horizontal compact basic buttons">
-                        {{#ui-popup content=(t 'Edit') class='ui icon button' click=(action 'openModalFor' eventTopic)}}
-                          <i class="edit icon"></i>
-                        {{/ui-popup}}
-                        {{#ui-popup content=(t 'Delete') class='ui icon button' click=(action (confirm (t 'Are you sure you would like to delete this event topic?') (action 'deleteEventProperty' eventTopic)))}}
-                          <i class="trash outline icon"></i>
-                        {{/ui-popup}}
+        <div style = "max-height: 768px;overflow-x: hidden;">
```
kushthedude: We strictly prohibit the use of custom styling or css in `hbs`, You will rarely see any custom css in the web-app. Please try to use `semantic ui classes`

maze-runnar:
I didn't find any class for setting max height , can you please suggest me any ?

kushthedude:
> I didn't find any class for setting max height , can you please suggest me any ?

Why you need max height to be setted ?

maze-runnar:
> > I didn't find any class for setting max height , can you please suggest me any ?
> 
> Why you need max height to be setted ?

above which scrolling will happen .

kushthedude:
@uds5501 Please see if we need this change as it is using custom css and we will anyway migrate it to ember table ?

app/templates/admin/content/events.hbs
uds5501: @kushthedude @maze-runnar This is fine for hotfix but I have one doubt, would it not mess up with mobile views?

maze-runnar: i think, it would not -
Screenshot from 2019-11-12 19-14-10
Screenshot from 2019-11-12 19-17-21


> @kushthedude @maze-runnar This is fine for hotfix but I have one doubt, would it not mess up with mobile views?

app/templates/admin/content/events.hbs
```diff
@@ -15,31 +15,33 @@
...
-                    </div>
-                    <div class="eight wide column right aligned">
-                      <div class="ui horizontal compact basic buttons">
-                        {{#ui-popup content=(t 'Edit') class='ui icon button' click=(action 'openModalFor' eventType)}}
-                          <i class="edit icon"></i>
-                        {{/ui-popup}}
-                        {{#ui-popup content=(t 'Delete') class='ui icon button' click=(action (confirm (t 'Are you sure you would like to delete this event type?') (action 'deleteEventProperty' eventType)))}}
-                          <i class="trash outline icon"></i>
-                        {{/ui-popup}}
+        <div style = "max-height: 768px;overflow-x: hidden;">
```
CosmicCoder96: No inline css please.

PR #3641, opened 2019-11-19
fix: making delay in table search when user inputs

### LINKED ISSUE #3532: Ember tables: Add a delay for search query network request trigger
Currently, each keystroke triggers a new query while searching on the table as the query param is directly bound to it, instead update the query param after a certain delay of 1 or 2 seconds.

maze-runnar:
Fixes #3532 
changes it proposed :
when the string enter in the input is greater than  a certain length a query triggers .
#### Checklist


app/mixins/ember-table-route.js
```diff
@@ -22,12 +22,14 @@ export default Mixin.create({
 
   applySearchFilters(options, params, searchField) {
     searchField = kebabCase(searchField);
-    if (params.search) {
-      options.pushObject({
-        name : searchField,
-        op   : 'ilike',
-        val  : `%${params.search}%`
-      });
+    if(params.search.length > 2){
```
prateekj117: @maze-runnar Rather than this, I would suggest to rather maybe trigger a query when someone presses enter. @iamareebjamal Your views on this?

iamareebjamal: Debouce will work, but yeah, we can make it explicit for now because even with debouce, there'll be a lot of redundant queries

kushthedude: This is very explicit, Also `FOSSASIA` is of 8 length, So 7 query will still be triggered. Also, This should not be the case where we are explicitly restricting query search of length one.
Rather fixing debounce should be our best option

iamareebjamal: > Also FOSSASIA is of 8 length, So 7 query will still be triggered

No queries will be triggered until enter is pressed

maze-runnar: > @maze-runnar Rather than this, I would suggest to rather maybe trigger a query when someone presses enter. @iamareebjamal Your views on this?

@prateekj117 i am stuck at one place can you please help me ?
```
filterOptions = this.applySearchFilters(filterOptions, params, searchField);
    let queryString = {
      get_trashed    : true,
      include        : 'event,speakers',
      filter         : filterOptions,
      'page[size]'   : params.per_page || 10,
      'page[number]' : params.page || 1
    };
    queryString = this.applySortFilters(queryString, params);

```

here how can i apply event  "/app/routes/admin/sessions/list.js " in the above portion of code .
filterOption is where the "applysearchFilter" triggers .
i want to use -

handleSearch(event) {
  if(event.keyCode === 13) {
filterOptions = this.applySearchFilters(filterOptions, params, searchField);
}
}
and in searchbox.hbs file 
<div class="ui small icon input">
  {{input type="text" value=searchQuery placeholder="Search ..."  key-up=(action 'handleSearch' )}}
  <i class="search icon"></i>
</div>

maze-runnar: if number of triggers after length of two is problem for **if(params.search.length > 2){**
then can change it to `
if(params.search.length > 2 && params.search.length %3 ===0){`
this will handle all searches from one place change other wise we have to change in every file where search-box is used as "admin/sessions/list" and "http://localhost:4200/events/live" and many other ..

iamareebjamal: Already taken care of https://github.com/fossasia/open-event-frontend/pull/3646/files

maze-runnar:
Ok

Issue #3642, opened 2019-11-19
Admin Search allows for server DOS


iamareebjamal:
Another problem with providing a lot of power with JSON:API is that it comes with a lot of responsibility on client side. A client can easily self DOS server with queries which it thinks are innocuous. This is what admin search does now.

First of all, it sends search filter events on every keystroke, an issue for which is already opened #3532 

But that is not the main issue. Let's say I search for fossasia. This is the query it'll send to the server:

https://api.eventyay.com/v1/events?filter=[{"name":"name","op":"ilike","val":"%fossas%"}]&get_trashed=true&include=tickets,sessions,speakers,owner,organizers,coorganizers,track-organizers,registrars,moderators&page[number]=1&page[size]=10"

See any problem? It requests the tickets, the **sessions, the speakers**, organizers,coorganizers,track-organizers,registrars,moderators of the event with the event itself. Just to render this:

![d444e951-378d-4f78-bc2f-e65c85d48f94](https://user-images.githubusercontent.com/3874064/69191666-ed990200-0b48-11ea-9b3e-80c3a0006f6f.png)

Why anybody thought of this as a good idea is beyond me. This is already so bad that I was shocked. But OK, at least the page size is set to 10. How much data it could possibly fetch? Right?

This is the response:

![Screenshot from 2019-11-20 03-47-27](https://user-images.githubusercontent.com/3874064/69191461-8713e400-0b48-11ea-869c-c46e391185bb.png)
![Screenshot from 2019-11-20 04-13-23](https://user-images.githubusercontent.com/3874064/69193247-5635ae00-0b4c-11ea-8049-e3b12299b98d.png)
![Screenshot from 2019-11-20 04-14-02](https://user-images.githubusercontent.com/3874064/69193254-59c93500-0b4c-11ea-9beb-a1c9f01ce4b3.png)

700 nodes in included! **700**. The response is 400 KB and the request takes 30 seconds to complete. **30 seconds** It's a relief there are not a lot of events with a lot of speakers and sessions, or you could cook and eat instant ramen before the request completes. 

But it's just one request, right? 
Because of this request, every other request, requests of even 70 bytes are slowed down because the server is pummeled. 

So what is happening? Apparently, JSON:API spec doesn't talk about restricting included items in the payload. Or the library used in server is poorly written, but basically, the results are limited to 10, but included items contain every single ticket, session and speaker in all the filtered 10 events!!!
Every . single . one
[...]

PR #3646, opened 2019-11-20
feat: Add debounce on search field and remove included resources

### LINKED ISSUE #3532: Ember tables: Add a delay for search query network request trigger

iamareebjamal:
Fixes #3532 and Partial fix for #3642

### DIFF
--- a/app/controllers/admin/events/list.js
+++ b/app/controllers/admin/events/list.js
@@ -76,11 +76,6 @@ export default class extends Controller.extend(EmberTableControllerMixin) {
          width         : 90
        },
        {
-         name          : 'Tickets',
-         valuePath     : 'tickets',
-         cellComponent : 'ui-table/cell/cell-tickets'
-       },
-       {
          name          : 'Public URL',
          valuePath     : 'url',
          cellComponent : 'ui-table/cell/cell-link',
--- a/app/routes/admin/events/list.js
+++ b/app/routes/admin/events/list.js
@@ -117,10 +117,10 @@ export default class extends Route.extend(EmberTableRouteMixin) {
     }
 
     filterOptions = this.applySearchFilters(filterOptions, params, searchField);
-
+    
     let queryString = {
       get_trashed    : true,
-      include        : 'tickets,sessions,speakers,owner,organizers,coorganizers,track-organizers,registrars,moderators',
+      include        : 'general-statistics,owner,organizers,coorganizers,track-organizers,registrars,moderators',
       filter         : filterOptions,
       'page[size]'   : params.per_page || 10,
       'page[number]' : params.page || 1
--- a/app/routes/admin/events/list.js
+++ b/app/routes/admin/events/list.js
@@ -120,7 +120,7 @@ export default class extends Route.extend(EmberTableRouteMixin) {
     
     let queryString = {
       get_trashed    : true,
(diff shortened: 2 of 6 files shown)


app/routes/admin/events/list.js
```diff
@@ -117,10 +117,10 @@ export default class extends Route.extend(EmberTableRouteMixin) {
     }
 
     filterOptions = this.applySearchFilters(filterOptions, params, searchField);
-
+    
     let queryString = {
       get_trashed    : true,
-      include        : 'tickets,sessions,speakers,owner,organizers,coorganizers,track-organizers,registrars,moderators',
+      include        : 'general-statistics,owner,organizers,coorganizers,track-organizers,registrars,moderators',
```
kushthedude: @iamareebjamal You will need to make this change in `routes/events/list` too

app/routes/admin/events/list.js
```diff
@@ -117,10 +117,10 @@ export default class extends Route.extend(EmberTableRouteMixin) {
     }
 
     filterOptions = this.applySearchFilters(filterOptions, params, searchField);
-
+
```
kushthedude: Trailing space

app/routes/admin/events/list.js
kushthedude: Also, Is the `general-statistics` needed ?

iamareebjamal: > You will need to make this change in routes/events/list too

Why?

> Also, Is the general-statistics needed ?

Yes, sessions and speaker counts are shown using general statistics

app/routes/admin/events/list.js
iamareebjamal:
```suggestion

```

iamareebjamal:
Removed general statistics for now because it is custom endpoint on server and not linked correctly. Tried but failed to link it to event correctly, so, will handle later/

iamareebjamal:
@prateekj117 @kushthedude Please review

app/routes/admin/events/list.js
kushthedude: > Yes, sessions and speaker counts are shown using general statistics

Speaker and Session Statistics are getting shown without using the `general-statistics' ?

app/controllers/admin/events/list.js
```diff
@@ -75,11 +75,6 @@ export default class extends Controller.extend(EmberTableControllerMixin) {
          cellComponent : 'ui-table/cell/cell-speakers-dashboard',
          width         : 90
        },
-       {
```
kushthedude: @iamareebjamal I think we should comment it out, As `ticket statistics` have played a crucial role in our tables. We will need it once the DOS issues are resolved?

iamareebjamal: This is a git repository, not a college project in a pen drive that something which is needed in the future needs to be commented out instead of deleting. It has the entire history of changes in the repo, any deleted thing can be recovered.

> We will need it once the DOS issues are resolved

Then it'll add the DOS issue again. When is it a good idea to fetch unbounded one to many relationships of a list of objects? Answer - Never
If it'll need to be added, it'll be completely reimplemented

kushthedude:
It looks good then.

Issue #3582, opened 2019-10-30
Wrong Event Time on public page


mariobehling:
The FOSSASIA page shows the event starts at 4pm. The event time in the wizard is: Start 9AM. I am using a browser in Germany.

Expected: Event time should show correct local time.
![Screenshot from 2019-10-30 12-35-15](https://user-images.githubusercontent.com/1583873/67855112-1f9ee180-fb12-11e9-838d-e483ca401950.png)

Wizard time of the same event:
![Screenshot from 2019-10-30 12-38-25](https://user-images.githubusercontent.com/1583873/67855207-45c48180-fb12-11e9-8d38-b247fb4c06a4.png)

maze-runnar:
i want to work on this sir.

kushthedude:
> i want to work on this sir.

@maze-runnar You already have issue assigned to you, Please complete those before.

iamareebjamal:
@kushthedude Update?

DaKeiser:
@kushthedude Can I take this issue ? I guess https://github.com/fossasia/open-event-frontend/pull/3581 would solve the issue temporarily by mentioning the reference of the timezone. Meanwhile I will look into this issue.

iamareebjamal:
No, that won't fix the issue. Wrong time is sent on the server, so it is stored incorrectly in the DB

kushthedude:
The issue is not related to different locations, server is sending wrong
time. Need to debug from server.

iamareebjamal:
server can't send wrong time. It doesn't process time at all. It returns whatever is stored in DB, so there is either a bug in event creation/edit form or event POST method (unlikely)

kushthedude:
I will check the creation process and data getting sent to db and get back
to you.

mariobehling:
Accessing the site from different timezones shows a different event time. It seems the system takes the time from browser through IP. And there is a conflict in the logic somewhere as the event time should be displayed according to the event time zone instead of the user's time (or the browser).

Same problem with timezone applies to ticket sales time. If I enter ticket sales end 23.59 for the event. Suddenly the live site will display it will end 7 hours later if I am in another time zone.

To keep things simple for now expected is:

Don't adjust according to any timezone of the user or browser. Only display the timezone marked in the wizard step 1 -> event time.

ImShashikant:
Hi @mariobehling, may I work on this issue?

kushthedude:
@mariobehling I have tried to access the same from different timezones,
[...]

PR #3652, opened 2019-11-23
fix: Changing the date format and fixing at other places

kushthedude:
Refers #3582


iamareebjamal:
Let's take this in a minor release after some testing

kushthedude:
> Let's take this in a minor release after some testing

After the session and speaker timezone issues are fixed, We can take this in account

kushthedude:
Will open a new PR with resolving the issues in Session Timings too.

PR #3654, opened 2019-11-24
feat: Billing Information Enhancements and Updations

### LINKED ISSUE #3518: Wizard Step 1/Order Page: Make Billing Info either Mandatory or Optional and change fields
The billing info is always collected even if the organizer does not want to collect this information. Also there are some discrepancies of the fields and collection data. Therefore please change the following:
1. In Wizard step 1: Add a button "Require billing info from ticket buyer." to enable the organizer to make this optional
Screenshot from 2019-09-10 20-24-29
2. In the order page do the following
  a) Only make billing info required if the organizer chose "Require billing info.." (in Wizard). If billing info is required always show the fields. If the organizer did not require it, show it as an option using a tick box. -> So, if the organizer does not require it, the user can still add the billing info optionally. The user should tick a box in that case to display the billing option.
  b) In the billing info section itself add field to enter a "Name". Make this field required.
  c) Make "Company" optional.

Screenshot from 2019-09-10 20-27-58

kushthedude:
Fixes #3518

### DIFF
--- a/app/components/forms/orders/order-form.js
+++ b/app/components/forms/orders/order-form.js
@@ -34,9 +34,18 @@ export default Component.extend(FormMixin, {
       this.data.set('paymentMode', 'free');
       return false;
     }
+
     return true;
   }),
   sameAsBuyer: false,
+
+  isBillingInfoMandatory: computed('event', function() {
+    if (this.event.isBillingInfoMandatory) {
+      return true;
+    } else {
+      return false;
+    }
+  }),
 
   getRemainingTime: computed('settings', function() {
     let orderExpiryTime = this.get('settings.orderExpiryTime');
@@ -380,12 +389,12 @@ export default Component.extend(FormMixin, {
             }
           ]
         },
-        company: {
-          identifier : 'company',
+        taxBusinessInfo: {
+          identifier : 'taxBusinessInfo',
           rules      : [
             {
               type   : 'empty',
-              prompt : this.l10n.t('Please enter your company')
+              prompt : this.l10n.t('Please enter your TAX ID or Business ID')
             }
           ]
         },
--- a/app/components/forms/orders/order-form.js
+++ b/app/components/forms/orders/order-form.js
@@ -38,8 +38,12 @@ export default Component.extend(FormMixin, {
   }),
   sameAsBuyer: false,
 
-  isBillingInfoMandatory: computed('event', function() {
-    return this.event.isBillingInfoMandatory;
+  isBillingInfoNeeded: computed('event', 'data.isBillingEnabled', function() {
+    return this.event.isBillingInfoMandatory || this.data.isBillingEnabled;
+  }),
+
+  isBillingNotMandatory: computed('event', function() {
+    return !this.event.isBillingInfoMandatory;
   }),
 
   getRemainingTime: computed('settings', function() {
@@ -389,7 +393,7 @@ export default Component.extend(FormMixin, {
           rules      : [
             {
               type   : 'empty',
-              prompt : this.l10n.t('Please enter your TAX ID or Business ID')
+              prompt : this.l10n.t('Please enter your Tax ID or Business ID')
             }
           ]
         },
--- a/app/components/forms/orders/order-form.js
+++ b/app/components/forms/orders/order-form.js
@@ -42,10 +42,6 @@ export default Component.extend(FormMixin, {
     return this.event.isBillingInfoMandatory || this.data.isBillingEnabled;
   }),
 
-  isBillingNotMandatory: computed('event', function() {
-    return !this.event.isBillingInfoMandatory;
-  }),
-
   getRemainingTime: computed('settings', function() {
     let orderExpiryTime = this.get('settings.orderExpiryTime');
     let willExpireAt = this.get('data.createdAt').add(orderExpiryTime, 'minutes');
--- a/app/components/forms/orders/order-form.js
+++ b/app/components/forms/orders/order-form.js
@@ -34,17 +34,12 @@ export default Component.extend(FormMixin, {
       this.data.set('paymentMode', 'free');
       return false;
     }
-
     return true;
   }),
   sameAsBuyer: false,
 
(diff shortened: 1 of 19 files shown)


kushthedude:
@iamareebjamal Please check, To be merged after: https://github.com/fossasia/open-event-server/pull/6447

app/components/forms/orders/order-form.js
```diff
@@ -34,10 +34,19 @@ export default Component.extend(FormMixin, {
...
     return true;
   }),
   sameAsBuyer: false,
 
+  isBillingInfoMandatory: computed('event', function() {
+    if (this.event.isBillingInfoMandatory) {
+      return true;
+    } else {
+      return false;
+    }
```
iamareebjamal: Or... `return this.event.isBillingInfoMandatory`

app/components/forms/orders/order-form.js
```diff
@@ -34,10 +34,19 @@ export default Component.extend(FormMixin, {
       this.data.set('paymentMode', 'free');
       return false;
     }
+
```
iamareebjamal: Revert

app/templates/components/forms/orders/order-form.hbs
```diff
@@ -81,52 +81,103 @@
...
-              {{#each countries as |country|}}
-                <div data-value="{{map-value mapper country.name}}" class="item">
-                  <i class="{{lowercase country.code}} flag"></i>
-                  {{country.name}}
-                </div>
-              {{/each}}
-            </div>
-          {{/ui-dropdown}}
-        </div>
+        {{#if isBillingInfoMandatory}}
```
iamareebjamal: This should instead be toggled by order amount?

kushthedude: I have checked the `if PaidOrder` above.

iamareebjamal: Updated comment. Check

app/templates/components/forms/orders/order-form.hbs
```diff
@@ -81,52 +81,103 @@
...
+                  {{#each countries as |country|}}   
+                    <div data-value="{{map-value mapper country.name}}" class="item">    
+                      <i class="{{lowercase country.code}} flag"></i>    
+                      {{country.name}}   
+                    </div>   
+                  {{/each}}    
+                </div>   
+              {{/ui-dropdown}}   
+            </div>   
+          {{/if}}
```
iamareebjamal: That's duplication. Otherwise known as 'Mother of Bugs' 

Remove any duplication

app/templates/components/forms/orders/order-form.hbs
kushthedude: > This should not be shown in free orders in any case. In paid orders, it should be shown unconditionally if billing info is mandatory. And if it is not, then it should be togglable by is_billing_enabled checkbox in order

Its the same happening here, Please check the else part. In the else there is another checkbox to toggle `is_billing_info`

iamareebjamal: https://stackoverflow.com/questions/4226284/how-to-convince-a-colleague-that-code-duplication-is-bad

app/templates/components/forms/orders/order-form.hbs
kushthedude: Will remove the duplicated code by tonight

kushthedude:
@iamareebjamal Removed duplications , Please give it a look

app/templates/components/forms/orders/order-form.hbs
iamareebjamal: There's still repetition

kushthedude:
Where?

app/templates/components/forms/orders/order-form.hbs
kushthedude: Since we need to render two seperate checkboxes, Altogether I have removed the duplication of forms. Also, We need to toggle the value of `is_billing_enabled` in other `else` part, hence I dont think it would be duplication. 😕

iamareebjamal: Will give you a solution

iamareebjamal:
```
if not billing_info_mandatory:
    show_checkbox
if billing_info_mandatory or is_billing_enabled:
    show billing form
```

kushthedude:
@iamareebjamal UPdated.

app/components/forms/orders/order-form.js
```diff
@@ -38,6 +38,14 @@ export default Component.extend(FormMixin, {
   }),
   sameAsBuyer: false,
 
+  isBillingInfoNeeded: computed('event', 'data.isBillingEnabled', function() {
+    return this.event.isBillingInfoMandatory || this.data.isBillingEnabled;
+  }),
+
+  isBillingNotMandatory: computed('event', function() {
+    return !this.event.isBillingInfoMandatory;
+  }),
```
iamareebjamal: Why can't you just use not in the template

app/templates/components/widgets/forms/billing-info.hbs
```diff
@@ -0,0 +1,41 @@
+<div class="field">
```
iamareebjamal: Can we move it back to the template now?

app/components/forms/orders/order-form.js
kushthedude: For or and not, I will need to use ember arithmentic helper add-on which cost more then the js checking

iamareebjamal: Not talking about or. I'm talking about `not`

app/templates/components/widgets/forms/billing-info.hbs
kushthedude: It looks more organised since its a just a tool which is needed for the `order-form`, I dont see any issue in moving it to a widget?

app/components/forms/orders/order-form.js
kushthedude: done the not

PR #3676, opened 2019-12-01
wip: Redirect after login

### LINKED ISSUE #2396: Redirect page after the login/signup on Call for speakers
**Edit** - See #2370 for relevant discussion
- @iamareebjamal 

**Describe the bug**

When an unauthenticated/unregistered user login/register to submit a session, he/she gets redirected to the homepage instead of the event that he/she was.

**To Reproduce**
1. Go to 'An event with open Call for Speakers section'
2. Click on 'Submit Proposal'
3. Login or Signup using an account
4. After this action the page redirects to homepage

**Expected behavior**

The page should redirect to the view of the "Call for speakers" section on the specified event.

**Desktop & Smartphones (please complete the following information):**
 - OS: Any OS
 - Browser: Any browser

iamareebjamal:
Fixes #2396

### DIFF
--- a/app/components/account/application-section.js
+++ b/app/components/account/application-section.js
@@ -23,14 +23,8 @@ export default Component.extend({
                 this.session
                   .authenticate(authenticator, credentials)
                   .then(async() => {
-                    const tokenPayload = this.authManager.getTokenPayload();
-                    if (tokenPayload) {
-                      this.authManager.persistCurrentUser(
-                        await this.store.findRecord('user', tokenPayload.identity)
-                      );
-                      this.set('data', this.get('authManager.currentUser'));
-                    }
-
+                    await this.authManager.loadUser();
+                    this.set('data', this.get('authManager.currentUser'));
                     this.set('isLoading', false);
                   });
               });
--- a/app/components/forms/login-form.js
+++ b/app/components/forms/login-form.js
@@ -52,13 +52,6 @@ export default class extends Component.extend(FormMixin) {
       });
       try {
         await this.session.authenticate(authenticator, credentials);
-        const tokenPayload = this.authManager.getTokenPayload();
-        if (tokenPayload) {
-          this.authManager.persistCurrentUser(
-            await this.store.findRecord('user', tokenPayload.identity)
-          );
-
-        }
       } catch (e) {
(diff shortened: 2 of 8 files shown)


iamareebjamal:
@prateekj117 Please test that the issue has been resolved or not and everything else changed elsewhere is working as expected. Test in fastboot mode as well

iamareebjamal:
@kushthedude Please test

kushthedude:
> @kushthedude Please test

Didn't get time for the complete test, A few key points:
1.) The redirection is not working in the case for the new users, If a new user signs up using the `CFS Modal` he should also get redirected to the same page.
2.) Also, the redirection is only working in case of `CFS Proposal` , There are other places like `create event` too where we would need this such behaviour.

iamareebjamal:
Second point was never handled in the first place. That's why it's not handled here as well. Redirection right now only works for the last page user was at when sent to login

Taking a look at first point, however, it doesn't seem likely it'll happen because first the user is redirected to login. At this point, there is context that user should be redirected to cfs after logging in. But then, for creatng a new account, it's again redirected to a new route. Then after registering, it is again redirected to login route. The context can be lost at any point here. The implementation of redirection is brittle and there are multiple issues open in ember-simple-auth for this. Much better solution is to use redirect URI parameter, but we are using ember and not native web platform, and ember takes the control from us

kushthedude:
What if we redirect to user to the cfs page rather than login after
registering?

iamareebjamal:
OK, and then again redirect to login when asked? Because the user is not logged in when registered

kushthedude:
Cant we log him in and redirect from the registration page only?
Or if we only want the solution for the CFS Page, We can follow the
approach which we are using during the order creation ?

iamareebjamal:
That approach is even buggier and most probably broke with this change. That's why the WIP status

kushthedude:
> That approach is even buggier and most probably broke with this change. That's why the WIP status

What were the bugs in the following ?

iamareebjamal:
Unreliability and duplication

prateekj117:
@iamareebjamal Tested. LGTM.

iamareebjamal:
Fixed by #3786

Issue #3418, opened 2019-08-15
Speaker Form : Show `Email Override` checkbox only to Admin.


kushthedude:
Currently, the `Checkbox` is getting shown to the speaker itself which is not the working expected.
Only show the following to `Super Admin` or `Admin`.

shreyanshdwivedi:
Workin on it? If yes, please assign yourself

CosmicCoder96:
show it to event organizer, coorganizer, owner as well.

kushthedude:
Duplicate of #3506

PR #3701, opened 2019-12-14
fix: Email Override possible for all users

### LINKED ISSUE #3506: Call for Speakers Signup Form: "Don't require this email" possible for everyone
On the Call for Speakers Signup Form, it should only possible for organizers to use "Don't require this email". Right now every speaker can just use this option.

Screenshot from 2019-09-04 19-23-19

kushthedude:
Fixes #3506

### DIFF
--- a/app/routes/public.js
+++ b/app/routes/public.js
@@ -8,7 +8,7 @@ export default Route.extend({
 
   model(params) {
     return this.store.findRecord('event', params.event_id, {
-      include: 'social-links,event-copyright,speakers-call,tax'
+      include: 'social-links,event-copyright,speakers-call,tax,owner,organizers'
     });
   }
 });
--- a/app/components/forms/session-speaker-form.js
+++ b/app/components/forms/session-speaker-form.js
@@ -501,6 +501,15 @@ export default Component.extend(FormMixin, {
     return this.speakers.length === 1;
   }),
 
+  isUserOwnerOrAdmin: computed('event.owner.email', 'authManager.currentUser', function() {
+    return (this.authManager.currentUser.isAnAdmin || (this.event.owner.get('email') === this.authManager.currentUser.get('email')));
+  }),
+
+  isUserOrganiser: computed('event.organizers', 'authManager.currentUser', function() {
+    let { currentUser } = this.authManager;
+    return this.event.organizers.includes(currentUser);
+  }),
+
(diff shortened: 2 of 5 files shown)


app/routes/public.js
```diff
@@ -8,7 +8,7 @@ export default Route.extend({
 
   model(params) {
     return this.store.findRecord('event', params.event_id, {
-      include: 'social-links,event-copyright,speakers-call,tax'
+      include: 'social-links,event-copyright,speakers-call,tax,owner'
```
iamareebjamal: Why including when you're already awaiting below. Only include things which are immediately visible in the page and defer things which can be lazy loaded

iamareebjamal:
Also, not just the owner, but also organizers should be given this option?

app/routes/public.js
kushthedude: @iamareebjamal This include states that the things which I will be using for the `public` route. If I wont define this here then the `edit-speaker` route's `eventDetails` which is the child route of `public` wont be able to fetch the owner.

kushthedude: If not this, I will have to make another query in the `edit-speaker` and `new-speaker` page which henceforth will increase the loadtime.

kushthedude:
> Also, not just the owner, but also organizers should be given this option?

Cool, I will the change the corresponding changes for same 👍 , But @iamareebjamal I don't think it should be displayed to `co-organisers` too ? https://github.com/fossasia/open-event-frontend/issues/3418#issuecomment-521713319

app/routes/public.js
iamareebjamal: > I will have to make another query in the edit-speaker and new-speaker page which henceforth will increase the loadtime.

It won't. This will surely increase the load time of public route without any reason

iamareebjamal:
Issue mentions only coorganizers

app/routes/public.js
kushthedude: I will optimise it once I push changes for the organiser visibility.

kushthedude: Done!

kushthedude:
@mariobehling @iamareebjamal Please merge, Tested for `Owner, Organiser, System Admin`. No one except them can see this option now.

kushthedude:
@iamareebjamal Please review & merge

iamareebjamal:
@prateekj117 Please test this

PR #3793, opened 2020-01-11
fix: Sessions were restricted to limit 20

### LINKED ISSUE #3788: When Organizer adds New Speaker, Sessions are Missing from Option "Choose from an existing session"
When an organizer adds new speaker, sessions are missing from the option "Choose from an existing session". Seems like the limit is the display of 20 sessions.

Screenshot from 2020-01-11 08-26-43

kushthedude:
Fixes #3788

### DIFF
--- a/app/routes/events/view/speakers/create.js
+++ b/app/routes/events/view/speakers/create.js
@@ -17,6 +17,7 @@ export default Route.extend({
         creator : this.get('authManager.currentUser')
       }),
       sessions: await eventDetails.query('sessions', {
+        'page[size]': 0
       }),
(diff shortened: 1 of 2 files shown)


iamareebjamal:
Open an issue to make it a search option as well, because there may be 100s of sessions and loading them all could be very inefficient

kushthedude:
> Open an issue to make it a search option as well, because there may be 100s of sessions and loading them all could be very inefficient

Like Load Sessions based on search keyword ?

iamareebjamal:
yes

Issue #3739, opened 2019-12-29
Organizer unable to add new speakers into new sessions 


edenyay:
**Describe the bug**
I failed to add speaker's info into the new session I've just added for the the FOSSASIA Summit 2020, even with organizer role. Email and other info disappeared, couldn't upload photo either.
![Screen Shot 2019-12-29 at 19 37 48](https://user-images.githubusercontent.com/32028823/71557031-fbf50a80-2a72-11ea-9edc-e6b639e8b745.png)

**Expected behavior**
Be able to add new sessions and speakers again 

**Additional details (please complete the following information):**
 - OS: [e.g. MacOS, 10.14.6]
- Browser: Chrome

iamareebjamal:
@kushthedude Is this duplicate of #3731 ?

kushthedude:
> @kushthedude Is this duplicate of #3731 ?

No, Also its been quite a time I have not used the following functionality. I will look into it today.

kushthedude:
@iamareebjamal Please dont make release, let me  test it more with more number of fields.

edenyay:
problem re-exists since yesterday: organizers can add new sessions but cannot speakers details.

kushthedude:
@edenyay Solved with a new release made.
Please re-open if it occurs again on production.

PR #3798, opened 2020-01-13
fix: URL Fields are segmented without protocol

kushthedude:
Refers #3739 
- One of the reason for submission error can be links getting submitted to server without the protocol of website .


kushthedude:
@iamareebjamal

PR #3805, opened 2020-01-14
fix: workaround for clearable property of dropdown

kushthedude:
- Once the session was selected it was not able to change it or make it null, this may cause a lot of issue if organiser want to add new session after looking into list of all sessions.


kushthedude:
@iamareebjamal We can drop a release after this!

kushthedude:
Need to update in semantic-ui-ember

Issue #3650, opened 2019-11-23
Speakers Page also DOSes the server


iamareebjamal:
https://eventyay.com/e/3689bc98/speakers (DO NOT OPEN!!! It will crash the server)

When accessing speakers page, the frontend sends query to acess all sessions of all speakers in separate queries, then all tracks of all all those sessions, and then all microlocations of all those tracks. All of this happens without include, so there are 1000s of requests handled by the server within seconds.

If there are 200 speakers and in aggregate all of them have 300 sessions, then 300 tracks and 300 microlocations, then 900 requests will be sent to server just to show the speaker names and picture. This is where include should be used instead of sending 900 requests to the server, but is still not needed. 

If you need to show the sessions of speakers, then request them when someone clicks on the speaker name. Not just load all sessions and tracks and microlocations redundantly. 

Frontend has been disassociated from reality and from what it loads on each page, please analyze the resources fetched and reduce the queries to the server. Just look at network tab and ensure than no more than 5 requests (obviously not a hard rule) are sent to the server and the payload is within 100 KB (which is still a LOT).

Related: #3642

kushthedude:
We don't also need to load tracks and micro-location on this page. Only the speaker and their related session should be enough. I will take a look at it

iamareebjamal:
No, even they shouldn't be loaded

iamareebjamal:
So, there was a bug in docker itself. docker-proxy in docker 18 stopped listening on port 8080 (server) when there were a lot of requests, even though it continued to listen on 4200 (frontend) for some reason. So even when the worker restarted and was listening inside the container, the host connection on port 8080 got refused because docker-proxy stopped listening to 8080. Upgrading to docker 19 fixed the issue, and server won't throw 502 now, but worker still gets overloaded due to this and restarts, so this needs to be fixed

kushthedude:
@iamareebjamal This may soon be a  frequent problem, As there are going to be a huge number of session/speakers for Summit-2020.
Proposed solutions:
- Implement Pagination in the Speakers/Session page instead of loading each of them.
- Implement an infinityLoader for the Speakers Page. And remove the extra information from the speaker column like `Micro location`, `Session Summary` as these properties can also be accessed from the Session page.
- Redesign the Speaker/Session Page and merging them into one where we can enlist the session with their corresponding speakers in spite of having two tabs for each of them.

iamareebjamal:
[...]

PR #3807, opened 2020-01-15
fix: using include in public speaker query

kushthedude:
Refers #3650

- Using include to fetch sessions correponding to speakers.
- Using FilterOptions to define filter rather than explicitly defining them in query object

### DIFF
--- a/app/routes/public/speakers.js
+++ b/app/routes/public/speakers.js
@@ -3,7 +3,7 @@ import Route from '@ember/routing/route';
 export default Route.extend({
   async model() {
     const eventDetails = this.modelFor('public');
-    let filterOptions = [
+    const filterOptions = [
       {
         or: [
           {
--- a/app/routes/public/speakers.js
+++ b/app/routes/public/speakers.js
@@ -3,34 +3,36 @@ import Route from '@ember/routing/route';
 export default Route.extend({
   async model() {
     const eventDetails = this.modelFor('public');
+    let filterOptions = [
+      {
+        or: [
+          {
+            name : 'sessions',
+            op   : 'any',
+            val  : {
+              name : 'state',
+              op   : 'eq',
+              val  : 'accepted'
+            }
+          },
+          {
+            name : 'sessions',
+            op   : 'any',
+            val  : {
+              name : 'state',
+              op   : 'eq',
+              val  : 'confirmed'
+            }
+          }
+        ]
+      }
+    ];
     return {
       event    : eventDetails,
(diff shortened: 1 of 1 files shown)


iamareebjamal:
Don't understand how it fixes anything?

kushthedude:
> If there are 200 speakers and in aggregate all of them have 300 sessions, then 300 tracks and 300 microlocations, then 900 requests will be sent to server just to show the speaker names and picture. This is where include should be used instead of sending 900 requests to the server, but is still not needed.

iamareebjamal:
Summarize the changes in the PR

kushthedude:
> Summarize the changes in the PR

Done

iamareebjamal:
> Using FilterOptions to define filter rather than explicitly defining them in query object

Yeah, so this confused me. Why?

kushthedude:
For ease of reference and understanding, should I revert this change of let
it be?

app/routes/public/speakers.js
```diff
@@ -3,34 +3,36 @@ import Route from '@ember/routing/route';
 export default Route.extend({
   async model() {
     const eventDetails = this.modelFor('public');
+    let filterOptions = [
```
iamareebjamal: Don't `let` it be. It should be const

iamareebjamal:
This increases load time of queries from 700 ms to 2 seconds. This is not the solution when all speakers are loaded. More than 200 speakers can be loaded in 1 go. 

Use pagination on speakers page and only load sessions when speaker detail is clicked on. And there include track and microlocation with session. Not sure where microlocation is being used now

kushthedude:
> This increases load time of queries from 700 ms to 2 seconds. This is not the solution when all speakers are loaded. More than 200 speakers can be loaded in 1 go.
> 
> Use pagination on speakers page and only load sessions when speaker detail is clicked on. And there include track and microlocation with session. Not sure where microlocation is being used now

When the UI Accordion is expanded, Session is shown with it's microlocation.

iamareebjamal:
Can't see microlocation in any single session in FOSSASIA Summit 2020. Please give an example

https://eventyay.com/e/a74665c3/speakers

kushthedude:
> some time ago, we discussed the idea of lazy loading speakers too -- if an event like FOSSASIA summit iself has a 100 speakers, loading all 100 images in one go does not make sense as a user either can't see them until he scrolls.

@CosmicCoder96 should infinityLoader be implemented ?

kushthedude:
This needs further review and changes.

PR #3822, opened 2020-01-18
chore: Bump node/yarn version & removing obsolete script

kushthedude:
- We had to revert back to node 8 due to some failing tests which is not the case right now
- We have netlify as our deployer and we no longer need surge script for deployment


kushthedude:
Not to be merged until next release. 
Need thorough testing before making into production.

iamareebjamal:
Update docker image

package.json
```diff
@@ -131,8 +131,8 @@
     "xgettext-template": "^3.4.2"
   },
   "engines": {
-    "node": "8.* || >= 10.*",
-    "yarn": ">= 1.0.0"
+    "node": ">= 10.*",
```
iamareebjamal: Why not 12

kushthedude: @iamareebjamal I am unable to make netlify utilise the latest version as it is using the cached version of node which is 10 used in recent PR builds

kushthedude: @iamareebjamal Ping!

Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN node scripts/l10n.js generate && \
 ##
 ##
 
-FROM node:10-alpine
+FROM node:12.14.1-alpine3.9
```
iamareebjamal: 12-alpine doesn't work?

kushthedude: It does work.

iamareebjamal:
netlify deploy is failing

kushthedude:
Itb6is because netlify is using cached version of last deployment which is
node10 . Hence it is failing.
Any solution for this?

kushthedude:
@iamareebjamal Please review & merge

PR #3840, opened 2020-01-23
fix: Add omise cdn in CSP

### LINKED ISSUE #3839: Omise Checkout is not accessible due to CSP
Omise checkout script is blocked by CSP Policy.

kushthedude:
Fixes #3839

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -154,6 +154,7 @@ module.exports = function(environment) {
         'https://platform.twitter.com',
         'https://cdn.syndication.twimg.com',
         'http://127.0.0.1:5000',
+        'cdn.omise.co/omise.js',
         'cdn.ravenjs.com'
       ],
       'font-src': [


iamareebjamal:
CSP needs to change a lot. There are other violations as well. Besides, CSP statically lists heroku and local apps as allowed. This should not be the case, it should extract origin from API_BASE_URL and use that to dynamically generate CSP like in case of SENTRY_DSN

kushthedude:
> CSP needs to change a lot. There are other violations as well. Besides, CSP statically lists heroku and local apps as allowed. This should not be the case, it should extract origin from API_BASE_URL and use that to dynamically generate CSP like in case of SENTRY_DSN

@iamareebjamal What about this https://github.com/rwjblue/ember-cli-content-security-policy

iamareebjamal:
Aren't we already using a dependency? How does this help?

PR #3856, opened 2020-01-26
fix: Implement infinityLoader for event speakers

kushthedude:
Refers #3650

### DIFF
--- a/app/routes/public/speakers.js
+++ b/app/routes/public/speakers.js
@@ -3,36 +3,40 @@ import Route from '@ember/routing/route';
 export default Route.extend({
   async model() {
     const eventDetails = this.modelFor('public');
-    return {
-      event    : eventDetails,
-      speakers : await eventDetails.query('speakers', {
-        filter: [
+    let filterOptions = [];
+    filterOptions = [
+      {
+        or: [
           {
-            or: [
-              {
-                name : 'sessions',
-                op   : 'any',
-                val  : {
-                  name : 'state',
-                  op   : 'eq',
-                  val  : 'accepted'
-                }
-              },
-              {
-                name : 'sessions',
-                op   : 'any',
-                val  : {
-                  name : 'state',
-                  op   : 'eq',
-                  val  : 'confirmed'
-                }
-              }
-            ]
+            name : 'sessions',
+            op   : 'any',
+            val  : {
+              name : 'state',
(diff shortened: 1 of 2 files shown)


kushthedude:
@iamareebjamal Let's include this in the latest release, I tested it on Summit 2020 results are really astonishing:
Before this: Speaker page loads in 20+ Seconds
After this: Speaker page loads in < 8 Seconds 

Also this 8 seconds will further be reduced to mere 3-4 seconds once the settings are removed from the global injection.

app/routes/public/speakers.js
```diff
@@ -3,36 +3,40 @@ import Route from '@ember/routing/route';
 export default Route.extend({
   async model() {
     const eventDetails = this.modelFor('public');
-    return {
-      event    : eventDetails,
-      speakers : await eventDetails.query('speakers', {
-        filter: [
+    let filterOptions = [];
```
iamareebjamal: Why is it let?

kushthedude: Changing to const !

kushthedude: @iamareebjamal Done

kushthedude: Changing to const will fail the ESLintrule as filterOption may tend to change , hence let should be used

iamareebjamal: > Changing to const will fail the ESLintrule

Then that ESLintrule is stupid. Nowhere in this code filterOptions is changed, neither can this reference ever be changed

Change it to const

kushthedude:
Should I disable the rule or disable the rule in this file only?

iamareebjamal:
What's the rule identifier?

kushthedude:
> What's the rule identifier?

```sh
1095            7:5 - 'filterOptions' is constant. (no-const-assign)
```

iamareebjamal:
That's triggered because you change to const and reassign in the next line. How is that ever going to work? :man_facepalming: const means no changing the value, do the declaration and assignment on the same line and change let to const

kushthedude:
> That's triggered because you change to const and reassign in the next line. How is that ever going to work? 🤦‍♂ const means no changing the value, do the declaration and assignment on the same line and change let to const
> 
> https://eslint.org/docs/rules/no-const-assign

Done 😅

Issue #3704, opened 2019-12-17
Remove/Replace abandoned add-ons getting used in our repository.


kushthedude:
Currently, There are many abandoned add-ons getting used in our app.

Some of which are causing trouble while reading the `browser logs` due to the depreciated usage of codes.

There are other alternatives easily available for such addons.
Some of the abandoned add-ons are : ember-g-map, ember-cli-geo

This need not be done immediately but will be a longterm benefit.

mariobehling:
Thanks! A number of add-ons have been removed. If there are any specific add-ons left over, please open a dedicated issue.

PR #3857, opened 2020-01-26
chore: Remove ember-cli-geo

iamareebjamal:
For #3704


kushthedude:
@iamareebjamal What about it's code usage in the repo, Should that also be removed ?

iamareebjamal:
It wasn't used anywhere. It was just added in 1 unrelated PR and forgotten about

PR #3868, opened 2020-01-26
fix: Improving the infinityLoader implemented

kushthedude:
Refers #3650


iamareebjamal:
https://www.dictionary.com/browse/improvise

app/templates/public/speakers.hbs
```diff
@@ -4,11 +4,6 @@
       {{public/speaker-item speaker=speaker}}
     </div>
   {{/each}}
-  {{#infinity-loader infinityModel=model}}
-    <div class="center aligned five wide column">
-      <div class="ui loading very padded basic segment">
-      </div>
-    </div>
-    {{infintyModel.reachedInfinity}}
+  {{#infinity-loader infinityModel=model triggerOffset=40 eventDebounce=30}}
```
iamareebjamal: Why did you remove the loader?

kushthedude: I just removed the spinner, infinityLoader is still there. Removing the spinner reduced the waiting on api host.

iamareebjamal: That doesn't make any sense. Please add it back

app/templates/public/speakers.hbs
```diff
@@ -4,11 +4,11 @@
...
-  {{#infinity-loader infinityModel=model}}
-    <div class="center aligned five wide column">
-      <div class="ui loading very padded basic segment">
-      </div>
-    </div>
+  {{#infinity-loader infinityModel=model triggerOffset=40 eventDebounce=30}}
+    <div class="center aligned five wide column">		
+      <div class="ui loading very padded basic segment">		
+      </div>		
+    </div>
```
iamareebjamal: Trailing spaces

kushthedude: Done

iamareebjamal:
Still the same problem. Pushes the list up instead of down. Tested in Firefox and Chrome. Happens if you are absolutely at the end of the list. Also, have to scroll up and down multiple times before it starts to load stuff. Merging for now, but it is very annoying experience

app/templates/public/speakers.hbs
```diff
@@ -4,7 +4,7 @@
       {{public/speaker-item speaker=speaker}}
     </div>
   {{/each}}
-  {{#infinity-loader infinityModel=model}}
+  {{#infinity-loader infinityModel=model triggerOffset=40 eventDebounce=30}}
```
iamareebjamal: Increase it more, still only loads speakers when I am absolutely at the bottom

iamareebjamal:
Also, load at least 12 or 15 speakers

app/templates/public/speakers.hbs
kushthedude: Done

iamareebjamal:
Well the sticking problem is still there and it only loads items when I am at absolute bottom, so threshold is not working as well. Will see tomorrow what can be done

kushthedude:
> Well the sticking problem is still there and it only loads items when I am at absolute bottom, so threshold is not working as well. Will see tomorrow what can be done

It is a wait of <1 sec TBH. Also, I think micro-location should not be fetched, As we are already displaying micro-location in the session card. What do you suggest?
@iamareebjamal

kushthedude:
> Well the sticking problem is still there and it only loads items when I am at absolute bottom, so threshold is not working as well. Will see tomorrow what can be done

It is a wait of <1 sec TBH. Also, I think micro-location should not be fetched, As we are already displaying micro-location in the session card. What do you suggest?
@iamareebjamal

iamareebjamal:
Let's take microlocation change in next release. About this PR, it doesn't change the experience for me unfortunately

kushthedude:
> Let's take microlocation change in next release. About this PR, it doesn't change the experience for me unfortunately

I tested it just now, It is atleast better than the previous option without offset which was loading 6 speaker at a time 😕 .

app/templates/public/speakers.hbs
```diff
@@ -4,7 +4,7 @@
       {{public/speaker-item speaker=speaker}}
     </div>
   {{/each}}
-  {{#infinity-loader infinityModel=model}}
+  {{#infinity-loader infinityModel=model triggerOffset=400 eventDebounce=50}}
```
iamareebjamal: Let's remove the changes here then

kushthedude: Sure!
Have we removed the codacy PR check?

iamareebjamal: Yes

iamareebjamal:
```suggestion
  {{#infinity-loader infinityModel=model}}
```

iamareebjamal:
@kushthedude Release is being made. Do you want this included?

kushthedude:
I suggest including it but I am afk right now, can you please push the required changes, if any?

PR #3870, opened 2020-01-26
chore: Add fastboot server in package.json

iamareebjamal:


kushthedude:
Please add safe-eval too

iamareebjamal:
Not needed

PR #3943, opened 2020-02-01
chore: removing template-inline-precompile

kushthedude:
ember-cli-htmlbars-inline-precompile is not needed anymore with dependencies upgrades.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "ember-cli-head": "^0.4.1",
     "ember-cli-html-minifier": "^1.1.0",
     "ember-cli-htmlbars": "^4.2.2",
-    "ember-cli-htmlbars-inline-precompile": "^2.1.0",
     "ember-cli-ifa": "^0.7.0",
     "ember-cli-inject-live-reload": "^2.0.2",
     "ember-cli-mirage": "^0.4.15",


iamareebjamal:
@kushthedude No need to merge dev into branches. You're just wasting CI minutes that way, a new build with no changes will be created. The only need for merges in when there are conflicts, DB migrations or interdependent dependencies

kushthedude:
> @kushthedude No need to merge dev into branches. You're just wasting CI minutes that way, a new build with no changes will be created. The only need for merges in when there are conflicts, DB migrations or interdependent dependencies

Then why there is need for update branch button, Also isn't target branch needs to be in sync with HEAD branch before development in this option ?

iamareebjamal:
> Then why there is need for update branch button

Because of this - The only need for merges in when there are conflicts, DB migrations or interdependent dependencies

There's an option to merge without CI passing, should that be done as well? I wish there was an option to restrict branch updates to only admin

> Also isn't target branch needs to be in sync with HEAD branch before development in this option ?

No

kushthedude:
> No

So you can merge a PR whose head is not in sync with development 😕 ?

iamareebjamal:
Yes

kushthedude:
Might be an admin only feature then😅

PR #3294, opened 2019-07-19
fix: Fix export events module(URL fetch logic)

mrsaicharan1:
EventExport
Fixes #3293 

#### Short description of what this resolves:
Fixes event exports with the help of loader service

Related Server fix:

#### Changes proposed in this pull request:

- Implemented an action to fix downloads & file serving
- Modified component template logic

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [ ] The acceptance, integration, unit tests and linter pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #4042, opened 2020-02-10
fix: event URL generated by frontend

kushthedude:
Fixes https://github.com/fossasia/open-event-server/issues/6833

### DIFF
--- a/app/components/events/view/export/download-zip.js
+++ b/app/components/events/view/export/download-zip.js
@@ -1,27 +1,5 @@
 import Component from '@ember/component';
-import { action } from '@ember/object';
 
 export default class extends Component {
-  @action
-  async exportEventDownload(eventDownloadUrl) {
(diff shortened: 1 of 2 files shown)


mrsaicharan1:
How is this the solution?

kushthedude:
This action was adding the extra server url in the event download url.

mrsaicharan1:
If there's an extra param, why would you change the whole logic? There's a precise workflow in place. Please check it out on the server and FE. 
I don't think this is the correct fix.

kushthedude:
The URL fetch logic was added in this #3294, Indeed the cause for invalid `localhost` URL specified in the issue might be empty fields for API URL & Frontend URL in Settings.

iamareebjamal:
https://github.com/fossasia/open-event-frontend/pull/4042#issuecomment-584160690

Issue #2602, opened 2019-04-12
Location Field Should be implemented like name .


kushthedude:
**Is your feature request related to a problem? Please describe.**

The Location which is a required field is implemented wrongly in Basic detail form , As name is required when user clicks submit , It scrolls back to name ,Similar should be implemented for Location .

**Describe the solution you'd like**

Implement Location field similar to name field

**Additional context**

I am working on it

kushthedude:
@ritikamotwani @CosmicCoder96 I observed when we dont fill location button is absent to submit , I propose to implement location field like name like when its empty and we try to submit , We get redirected to the field with a tooltip

ritikamotwani:
No, it's not required. Discussions regarding the location field are ongoing. First of all it is not a required field if you want to save your event and not publish it. And also we intend to hold online events which won't require a location. So any kind of refactoring would be out of scope for this issue. Hence closing it.

kushthedude:
No problem will close the pr too

PR #4059, opened 2020-02-11
fix: remove location check from event-wizard mixin

### LINKED ISSUE #4058: Basic Details Step: can not save event as draft without filling location field.
**Describe the bug**

Self Explainatory.

**Expected behaviour**

Event should be saved as draft without location.

**Additional context**

On it.

snitin315:
Fixes #4058 

Now Events can be saved as draft without filling the location field.

### DIFF
--- a/app/mixins/event-wizard.js
+++ b/app/mixins/event-wizard.js
@@ -117,9 +117,6 @@ export default Mixin.create(MutableArray, CustomFormMixin, {
       if (event.name === undefined || event.name === '') {
         errorObject.errors.push({ 'detail': 'Event name has not been provided' });
       }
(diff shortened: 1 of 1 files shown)


iamareebjamal:
And can the event be published without location?

snitin315:
> And can the event be published without location?

@iamareebjamal No.

kushthedude:
> And can the event be published without location?

Event should not be published without location, But it should be able to get saved as a draft if location is not mandatory according to https://github.com/fossasia/open-event-frontend/issues/2602#issuecomment-482779381. This FE check was forcing the client to give location before saving as draft too.

PR #4093, opened 2020-02-15
feat: Parallelize application requests

iamareebjamal:
From 10 seconds to 5 seconds of initial page load

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -30,10 +30,11 @@ export default Route.extend(ApplicationRouteMixin, {
   },
 
   async model() {
-    let notifications = [];
+
+    let notificationsPromise = Promise.resolve([]);
     if (this.get('session.isAuthenticated')) {
       try {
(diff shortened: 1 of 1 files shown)


app/routes/application.js
```diff
@@ -30,10 +30,11 @@ export default Route.extend(ApplicationRouteMixin, {
   },
 
   async model() {
-    let notifications = [];
+
```
iamareebjamal:
```suggestion
```

app/routes/application.js
```diff
@@ -49,17 +49,29 @@ export default Route.extend(ApplicationRouteMixin, {
       }
     }
 
+    const pagesPromise = this.store.query('page', {
+      sort: 'index'
+    });
+
+    const settingsPromise = this.store.queryRecord('setting', {});
```
kushthedude: Was this required? Settings are already injected in the application 🤔

iamareebjamal: OK, then use it here

PR #4130, opened 2020-02-22
feat: Restoring surge deployments for PR

kushthedude:
Fixes #

### DIFF
--- a/scripts/pr-deploy.sh
+++ b/scripts/pr-deploy.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+    echo "Not a PR. Skipping surge deployment"
+    exit 0
+fi
+
+ember build production
+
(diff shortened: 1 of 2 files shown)


scripts/pr-deploy.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+    echo "Not a PR. Skipping surge deployment"
+    exit 0
+fi
+
+ember build production
```
iamareebjamal: This is already run

kushthedude:
@iamareebjamal Surge deployments are working but the open-event-commenter app is not working which is used to comment the surge link on the PR

iamareebjamal:
Use danger - https://danger.systems/

kushthedude:
> Use danger - https://danger.systems/

Looking into it

scripts/pr-deploy.sh
kushthedude: Ember Build Production is needed for succesfull surge deployment. I am thinking to put deploy preview in another job other than tests.

PR #4298, opened 2020-04-05
chore: Convert components to angle bracket syntax

iamareebjamal:


iamareebjamal:
If you're spending 4x time on solving framework compatibility issues and bugs as compared to fixing bugs and adding features in your own project, there is something seriously wrong in that framework :(((

iamareebjamal:
Spent 2 hours chasing useless errors and indecipherable stacktraces

kushthedude:
> Spent 2 hours chasing useless errors and indecipherable stacktraces

But alas, Build is still failing 🏗

iamareebjamal:
It never was successful. This is the thing with ember. I'll try something, it'll fail. I'll curse and vent and come back a week later to be detective again in a crime where I myself am the murderer and the murdered and solve it and break something again and repeat

iamareebjamal:
So bye for another week and I'll fix it and break something again probably

snitin315:
Maybe its time to create a new branch and start migrating the project to react bit by bit.
not an easy task :disappointed:

iamareebjamal:
Let's check by upgrading to octane first and see if it works out. None of it is easy. Neither migrating to react/vue

iamareebjamal:
I feel like it is some kind of joke where in one place, it is said to upgrade to latest ember and someone else says stay classic
https://twitter.com/mwpastore/status/1246980602412351489

kushthedude:
Octane is new, but ember has still not dropped support for classic format.
Anyway sooner or later, it will drop support and move away from classic
forever. So we would need to upgrade anyway, only advantage we can get
later is some new and efficient codemods which may not require a lot of
labour work perhaps.

iamareebjamal:
The weird thing about anyone who is defending ember here is that no one among them is creating PRs to move the project further. I am. And it is exhausting debugging this stuff. So, maybe if @kushthedude can move the project ahead by fixing the PR and handling other required changes to get the project to latest standard. Or should we wait for it to be deprecated?

kushthedude:
> The weird thing about anyone who is defending ember here is that no one among them is creating PRs to move the project further. I am. And it is exhausting debugging this stuff. So, maybe if @kushthedude can move the project ahead by fixing the PR and handling other required changes to get the project to latest standard. Or should we wait for it to be deprecated?

I was just explaining that guy's point of view 😬

kushthedude:
> The weird thing about anyone who is defending ember here is that no one among them is creating PRs to move the project further. I am. And it is exhausting debugging this stuff. So, maybe if @kushthedude can move the project ahead by fixing the PR and handling other required changes to get the project to latest standard. Or should we wait for it to be deprecated?

I was just explaining that guy's point of view. Also, I was not the one to choose Ember for the project.

iamareebjamal:
The migration we are doing here is not octane, it is ES6 classes and angle bracket syntax. And it is not even problem with it, at least ember has codemods. But the debugging is too difficult that it took me an hour to get which component was throwing an error and the error was completely unrelated as well. It originated from ember composable helpers `sort by` which would not even be needed if ember did not use hbs.

1. Debugging is terribly hard
2. HBS is atrocious to work with and needs 3rd party packages to get basic stuff
3. Packages are not maintained (issue open for 6 months in repo)

snitin315:
Ember is a nightmare

jaydgruber:
Hi folks. Please take a look at #4301, which hopefully should provide a template for a more incremental path forward

Also, from personal experience, when I see a team really struggling with and fighting with Ember, and venting some frustration like this, it is because they are cut off from the community.

The majority of Ember developers are not putting their problems and answers up on StackOverflow as you might be used to from other ecosystems. It can be frustrating to google a problem and not find any answers. The best way to find help is to drop into the [discord](https://discordapp.com/invite/emberjs) and, I suggest the `#topic-octane-migration` channel. I think you will find a very welcoming and helpful community there

iamareebjamal:
Finally! Fingers crossed!!!

PR #4301, opened 2020-04-06
chore: Angle Brackets Codemod on app/template/accounts/

jaydgruber:
### What
- removes ember-cli-template-lint in favor of using ember-template-lint directly
  - ref: [ember-cli-linter deprecations](https://github.com/ember-template-lint/ember-cli-template-lint#-deprecated-), which moves to simplify lint calling styles and make them more similar across js & hbs lint
  - a [major bump](https://github.com/ember-template-lint/ember-template-lint/blob/master/CHANGELOG.md#v210-2020-02-28) of ember-template-lint is required to use the `overrides` configuration, which will be helpful to make incremental progress on codemods
- runs the [ember-angle-brackets-codemod](https://github.com/ember-codemods/ember-angle-brackets-codemod) on `app/template/accounts/`
- enables the template-lint rule [no-curly-component-invocation](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-curly-component-invocation.md) on the same directory to prevent backsliding

### Why
- incremental progress on conversion to angle bracket syntax
- suggestions to resolve some of the frustration expressed on #4298 

### Suggestions for a path forward
- ask for help earlier before getting frustrated! Join the community over in [Discord](https://discordapp.com/invite/emberjs)
  - `#topic-octane-migration` `#e-template-lint` channels may be of interest
- leverage existing tools!
  - most codemods have a paired lint rule in [eslint-plugin-ember](https://github.com/ember-cli/eslint-plugin-ember#ember-octane) or [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint)
- find ways to break the work down into smaller chunks. [pull request recommendations](https://opensource.com/article/18/6/anatomy-perfect-pull-request) suggest ~250 **lines** of code for the average code review to reduce burden on the reviewer. 250 **files** is too much!
  - Personally, I have found that codemods are ok to be a bit bigger in size than your normal code review, since they are just making a repetitive change over and over. Feedback from my team tells me ~50 files changed is a reasonable limit for codemod PRs, to resepect their time
  - Help your reviewers out with good PR descriptions. 95% of a codemod PR is going to be the same change. If you have to make any additional manual changes, call them out clearly in the PR description so that reviewers can pay extra attention to those. If you find yourself writing more than ~5-10 types of manual change in a codemod PR description... maybe a good time to split off into the next incremental PR
- momentum is important. find ways to keep making progress each day. If you find yourself stuck, instead of going in circles on the same problem:
[...]

PR #4323, opened 2020-04-16
chore: Change to angle bracket syntax for sort_by

iamareebjamal:
{{sort_by}} was being treated as a composable helper
and thus all builds like #4301 and mainly #4298 were breaking
Some changes were reverted in #4322, but changing to
{{this.sort_by}} makes it unambiguous and also octane
compatible


iamareebjamal:
This is extremely repetitive

PR #4371, opened 2020-05-09
fix: Update in local installation file

ankit1011-coder:
I was update the installation link of Watchman page. because that is not found in old file.


iamareebjamal:
Should have been chore, this was not a bug fix

kushthedude:
Oops😅

PR #4538, opened 2020-07-08
chore: Finally remove @open-event/theme

### LINKED ISSUE #4526: Start Page not scrollable on mobile
The start page is not scrollable on Android mobile v10 chrome.

This happens once the event search bar loads.

Screenshot_20200704_163111_com android chrome

iamareebjamal:
Fixes #4526 

Since we have reverted to rounded corners and undone any changes to default semantic theme, there is no use of depending on fork of semantic UI theme as it is difficult to maintain and causes random undebuggable bugs like the one linked.

To be fixed before merged:

- [ ] Modals are not showing up correctly

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -43,14 +43,6 @@ module.exports = function(defaults) {
     sourcemaps: {
       enabled: true
     },
-    SemanticUI: {
-      source: {
-        css        : 'node_modules/@open-event/theme/dist',
-        javascript : 'node_modules/@open-event/theme/dist',
-        images     : 'node_modules/@open-event/theme/dist/themes/default/assets/images',
-        fonts      : 'node_modules/@open-event/theme/dist/themes/default/assets/fonts'
-      }
-    },
     autoImport: {
       webpack: {
         externals : { jquery: 'jQuery' },
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -18,10 +18,10 @@ module.exports = function(defaults) {
     },
     SemanticUI: {
       source: {
-        css        : 'node_modules/fomantic-ui/dist',
-        javascript : 'node_modules/fomantic-ui/dist',
-        images     : 'node_modules/fomantic-ui/dist/themes/default/assets/images',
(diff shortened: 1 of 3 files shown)


iamareebjamal:
Still having issues with fomantic UI for modals. Will diagnose more what can be done

kushthedude:
@iamareebjamal Won't we want to update the open-event theme to use fomantic & then refer it here like it was done in case of semantic-ui?

iamareebjamal:
Why would we depend on an indirection and maintain another package when we aren't changing anything? And then update 2 packages when there is an update to fomantic UI and make a release in open even theme?

kushthedude:
> Why would we depend on an indirection and maintain another package when we aren't changing anything? And then update 2 packages when there is an update to fomantic UI and make a release in open even theme?

Makes sense, but I thought we would want it to maintain the open-event ecosystem.

iamareebjamal:
Why? Any reason?

kushthedude:
> Why? Any reason?

I thought we would want to maintain the open-event ecosystem including the design system too, so next if I someone would need fomantic-ember they can see the open-event-theme.

iamareebjamal:
There is no open-event theme. It is exactly same as default semantic UI theme. Just rounded corners were removed and now have been restored

iamareebjamal:
We can't use fomantic UI unless https://github.com/fomantic/Fomantic-UI/issues/811 is resolved

iamareebjamal:
This is the reason for scroll block https://github.com/Semantic-Org/Semantic-UI/issues/6449

iamareebjamal:
This commit fixes the issue https://github.com/fomantic/Fomantic-UI/pull/273/files but the latest code https://raw.githubusercontent.com/fomantic/Fomantic-UI/master/src/definitions/modules/modal.js causes this issue https://github.com/fomantic/Fomantic-UI/issues/1590 which is not reproducible in a minimal example

So, merging for now. Will find proper solution after more debugging

iamareebjamal:
If no solution is found, then we'll patch our own code with the PR so that both scrolling issue and modals dismissal is fixed

PR #4601, opened 2020-07-21
fix: Filter attendee form from all custom forms

### LINKED ISSUE #4597: Attendee Form: Editing the form sometimes suddenly shows field of organizer custom forms
Attendee Form: Editing the form sometimes suddenly shows field of organizer custom forms, e.g. compare https://eventyay.com/events/a2b1a68e/edit/attendee

Screenshot from 2020-07-21 16-46-10

iamareebjamal:
Fixes #4597

Ember Data strikes again. There is a custom function `query` used everywhere throughout the project which filters out the items from server. But ember data returns all the items from local in memory DB because how cool it is, and this custom query method does not take it into account. Hence, once all custom form fields are loaded in memory, it returns all the items rather than just the attendee custom form

### DIFF
--- a/app/routes/events/view/edit/attendee.js
+++ b/app/routes/events/view/edit/attendee.js
@@ -18,11 +18,11 @@ export default class AttendeeRoute extends Route.extend(CustomFormMixin) {
     const event = this.modelFor('events.view');
     const data = {
       event,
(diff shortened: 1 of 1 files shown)


mariobehling:
OMG! Thanks for the fix!

PR #4705, opened 2020-08-04
fix: Session times in public session list

### LINKED ISSUE #4238: Public Schedule: Times of sessions are displayed wrongly
The session times are displayed wrongly even though I am in the same time zone of the event.

Screenshot from 2020-03-18 10-30-03

Haider8:
Fixes #4238 

Session time was wrong

Show session time after converting to that event's timezone

#### Checklist


iamareebjamal:
@Haider8 Travis Build is failing

Haider8:
@iamareebjamal Couldn't fix, need help. It says `Cannot read property 'get' of undefined`

iamareebjamal:
See the code you have added and the test which is failing. Should be clear why the test fails

Haider8:
I am using this line to get the event timezone, `this.session.event.get('timezone')`.

iamareebjamal:
Yes, and see the test which is failing

iamareebjamal:
Still failing

app/components/public/session-item.js
```diff
@@ -15,4 +16,9 @@ export default class SessionItem extends Component {
       this.set('hideImage', false);
     }
   }
+
+  @computed('session.event', 'session.startsAt')
+  get sessionStartsAtInEventTimezone() {
+    return moment.tz(this.session.startsAt, this.session.event.get('timezone'));
```
iamareebjamal, line 22: What happens if you use `session.event.timezone`?

Haider8: It tells that since `session.event` is a proxy, it will be better that I use `.get()` rather than directly accessing that property.

iamareebjamal: Then wrap event in EmberObject in the test

iamareebjamal:
Like this https://github.com/fossasia/open-event-frontend/blob/c1572e612397709182437b2858798b57c8bb76cd/tests/integration/components/order-card-test.js#L11

app/components/public/session-item.js
Haider8: Like this ?
```js
`event: EmberObject.create({ timezone: 'Asia/Kolkata' })
```

iamareebjamal: Yes

PR #4783, opened 2020-08-17
fix: Changing tests from assert.ok to specific test cases

### LINKED ISSUE #4689: Change tests from assert.ok to specific test cases
assert.ok is not fit for any checking except those returning booleans, it was copy-pasted without checking actual use cases. When a test fails, it is important to know what was expected and what was actual value, but assert.ok tells you:

Expected: true
Actual: false

🤷 

Replace assert.ok with specific methods like `assert.dom(element).includesText` which will tell what the actual value and expected value was if a test fails

maze-runnar:
Fixes #4689

#### Checklist


maze-runnar:
will reopen when it will come in scope.

PR #4845, opened 2020-08-23
fix: Standardize color utils and speaker session UI, timezone

iamareebjamal:
- [x] Refactor all color utils in one place
- [x] Adjust threshold for switching colors
- [x] Deduplicate color helpers
- [x] Make session list UI in speakers aligned with global UI
- [x] Correct timezone being used in session list in speakers

Fixes #4846

Before:
image
image

After:
image
image


mariobehling:
Very good attention to detail. Thanks!

PR #555, opened 2017-07-25
Adding user awareness text if no data available in my sessions

harshitagupta30:
#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The acceptance, integration, unit tests and linter pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

#### Short description of what this resolves:
Adding user awareness text if no data available for my-sessions route.

Deployment link: https://gentle-taiga-30801.herokuapp.com/my-sessions
Fixes #396 

@fossasia/open-event-frontend Please Review. Thanks.

Issue #333, opened 2017-06-23
Remove add icon from discount code page under tickets to maintain uniformity


harshitagupta30:
**I'm submitting a ...**  (check one with "x")
- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, instead ask your query in out Gitter channel at https://gitter.im/fossasia/open-event-frontend

**Current behavior:**
Everywhere in the app for creating any item, we are using only text starting with "create" but in discount code page under the tickets tab, there is an additional add icon along with the text on "Create a discount" button.
 
![selection_073](https://user-images.githubusercontent.com/12701036/27492149-c1e2c1be-5862-11e7-8cc5-40a43fe55995.png)
 
**Expected behavior:**
Remove the add icon from the "Create a discount" button to maintain consistency

**Steps to reproduce:**
Check "Create a discount" button under discount codes in tickets tab.

sumedh123:
@harshitagupta30 Please close this one.

PR #4901, opened 2020-08-28
fix: Enhance filter opacity and color

### LINKED ISSUE #4899: Public Event Page: Enhance filter feature
Go to session tab on this public page https://eventyay.com/e/8fa7fd14/sessions?track=Projects%20%26%20People and select a filter 

**Filter mode is not so visible**
After I selected a filter, after few seconds I already forgot that the filter was on and which filter I have selected. Highlighted grey color is not catchy to human eyes. The clear filter icon looks too prominent and negative. If we must use an icon, I suggest something like [this](https://commons.wikimedia.org/wiki/File:Toicon-icon-lines-and-angles-filter.svg 
)

filter

**Proposed solution: Simple fix**
Change the highlighted color to a stronger color 

OR 

**Alternative solution**
I think we can follow this example from Sched. Users can see in the central, which filter is on and how to clear it

Screenshot from 2020-08-28 06-24-18

maze-runnar:
Fixes #4899 

Screenshot from 2020-08-28 10-56-41

#### Checklist

### DIFF
--- a/app/styles/components/extras.scss
+++ b/app/styles/components/extras.scss
@@ -3,10 +3,10 @@ a.link-item {
   display: table;
 
   &.active {
-    background-color: rgba(0, 0, 0, .05);
+    background-color: rgba(0, 0, 0, .1);
   }
 
   &:hover {
-    background-color: rgba(0, 0, 0, .025);
+    background-color: rgba(0, 0, 0, .1);
   }
 }
--- a/app/styles/components/extras.scss
+++ b/app/styles/components/extras.scss
@@ -7,6 +7,6 @@ a.link-item {
   }
 
   &:hover {
-    background-color: rgba(0, 0, 0, .1);
+    background-color: rgba(0, 0, 0, .05);
   }
 }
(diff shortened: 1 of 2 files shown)


app/styles/components/extras.scss
```diff
@@ -3,10 +3,10 @@ a.link-item {
   display: table;
 
   &.active {
-    background-color: rgba(0, 0, 0, .05);
+    background-color: rgba(0, 0, 0, .1);
   }
 
   &:hover {
-    background-color: rgba(0, 0, 0, .025);
+    background-color: rgba(0, 0, 0, .1);
```
iamareebjamal, line 10: Both opacity shouldn't be equal

app/templates/public.hbs
```diff
@@ -48,7 +48,7 @@
               <h4>
                 {{t 'Tracks'}}
                 <LinkTo @route="public.sessions" @models={{array this.model.id}} @query={{hash track=null}} data-tooltip="{{t 'Clear Track Filter'}}">
-                  <i class="minus circle icon" style={{css color='#f00'}}></i>
+                  <i class="minus circle icon" style={{css color='#3f3f40'}}></i>
```
iamareebjamal, line 51: #333 or #555

iamareebjamal (approved):

PR #4932, opened 2020-09-01
fix: increase page size to 21 for speakers

### LINKED ISSUE #4866: Public Speakers and Schedule Page: No Lazy Loading, it Slows Down Users Too Much
Public speakers page and schedule pages should have not lazy loading as it slows down the usage of the site too much, e.g. compare here https://eventyay.com/e/8fa7fd14/speakers

maze-runnar:
Fixes #4866
Making page size to 21 is giving better performance than 12 in terms of breaks while scrolling on a pageSize of 12 and total time also. And Also loads fast. So this is kind of trade of between total time and loading time I guess. So I found pageSize 21 as a best option.

#### Checklist

### DIFF
--- a/app/routes/public/speakers.js
+++ b/app/routes/public/speakers.js
@@ -47,7 +47,7 @@ export default class SpeakersRoute extends Route {
       event    : eventDetails,
       speakers : await this.infinity.model('speakers', {
         filter       : SPEAKERS_FILTER,
-        perPage      : 12,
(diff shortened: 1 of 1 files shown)


iamareebjamal:
I still think it doesn't trigger at the correct time but merging as a workaround for now. Let's see what else can be done later

iamareebjamal (approved):

PR #4964, opened 2020-09-03
fix: Revert LinkInput changes

### LINKED ISSUE #4960: Social System Links functionality URL and Form bugs after Recent Changes in Admin Section
Recently the social system links functionality was updated and now there are issues with each form field:
* Every field now requires a URL even though the URL field is already preceding
* blogs.fossasia.org should not be there as this is the admin should be able to add any domain here. So only show https:// or http:// as an option
* Please also add https://www.patreon.com as an option

Compare https://eventyay.com/admin/content
Screenshot from 2020-09-03 18-25-10

### LINKED ISSUE #4962: Admin: API URL hardcoded
The API URL is suddenly hardcoded in the admin section. Compare https://eventyay.com/admin/settings

Screenshot from 2020-09-03 19-45-33

iamareebjamal:
Fixes #4960 
Fixes #4962

### DIFF
--- a/app/utils/computed-helpers.js
+++ b/app/utils/computed-helpers.js
@@ -16,18 +16,17 @@ export const computedSegmentedLink = function(property) {
       const splitted = this.get(property) ? this.get(property).split('://') : [];
       if (!splitted || splitted.length === 0 || (splitted.length === 1 && splitted[0].includes('http'))) {
         return {
-          protocol : 'https://',
+          protocol : 'https',
           address  : ''
         };
       }
-      const socialUrl = splitted[1].split('/');
       return {
-        protocol : `${splitted[0]}://${socialUrl[0]}/`,
-        address  : socialUrl[1]
+        protocol : splitted[0],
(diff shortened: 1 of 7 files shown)


app/utils/computed-helpers.js
```diff
@@ -16,18 +16,17 @@ export const computedSegmentedLink = function(property) {
...
         return {
-          protocol : 'https://',
+          protocol : 'https',
           address  : ''
         };
       }
-      const socialUrl = splitted[1].split('/');
       return {
-        protocol : `${splitted[0]}://${socialUrl[0]}/`,
-        address  : socialUrl[1]
```
snitin315, line 26: I think the main bug was here, this should happen only for social links and not for other links like `https://api.eventyay.com/`.

Also adding an attribute `isSocialLink` for LinkInput widget.

If this sounds good to you I will create a PR.

iamareebjamal: Go ahead

PR #5000, opened 2020-09-07
feat: handle social links with LinkInput widget

### LINKED ISSUE #4609: Forms: Social Media - Only request to fill in short form, e.g. username for link specification
Make social media links on all forms in a way that users only need to add their username or other short forms.

Relevant Forms:
* Wizard step two social links
* Attendee Forms
* Speaker submission forms
* Speaker profile form

E.g. show as follows
Twitter https://twitter.com/___________
Github https://github.com/___________
Linkedin https://www.linkedin.com/____________
Instagram https://instagram.com/______________
Facebook https://facebook.com/_______________
YouTube https://youtube.com/________________

Screenshot from 2020-07-22 17-20-27

snitin315:
Fixes #4609 

captured-social


app/utils/computed-helpers.js
```diff
@@ -14,19 +14,33 @@ export const computedSegmentedLink = function(property) {
   return computed(property, {
     get() {
       const splitted = this.get(property) ? this.get(property).split('://') : [];
+      const socialPlatforms = ['twitter', 'facebook', 'instagram', 'linkedin', 'youtube', 'github', 'gitlab', 'vimeo', 'flicker'];
```
iamareebjamal, line 17: Move this up and export this and import it in the component above to remove duplication

snitin315: Done

iamareebjamal:
Have you checked that it doesn't interfere with any other place it's being used?

snitin315:
Let me implement https://github.com/fossasia/open-event-frontend/issues/4609 too in this PR, to make sure we aren't breaking anything.

mariobehling:
Is it possible to have a custom field also. Right now it is still possible to define the description yourself, e.g. I just added "Add to calendar" on the OSI event. 

Screenshot from 2020-09-08 12-21-47
Compare https://eventyay.com/e/8fa7fd14

snitin315:
> Is it possible to have a custom field also? Right now it is still possible to define the description yourself, e.g. I just added "Add to calendar" on the OSI event.

Yes, the possible solution I can think of this right now is that we can add one more `+` button for custom field.
`+ Social Link` `+ Custom Link` `- Delete`

mariobehling:
There is a way to have a dropdown menu with predefined fields and additionally you can enter a custom name. Google has something like that in their contact manager. Please see below. We need this as well.

Screenshot from 2020-09-08 15-37-42

snitin315:
We will probably need to write our custom component for this behavior. Can we do it with a separate Issue/PR?

iamareebjamal:
It will take more time @mariobehling as there is no such component already in semantic UI.

mariobehling:
Ok, thanks for the info.

> Yes, the possible solution I can think of this right now is that we can add one more + button for the custom field.

It seems then a custom field would be an immediate option even though it would make the UI a bit more complicated. Could you add this here as well? Basically it would seem we keep the current way and separate it into two components a) social links as covered in the PR already and b) simply using the former way for custom links.

app/templates/components/forms/wizard/other-details-step.hbs
```diff
@@ -16,6 +16,7 @@
       <Widgets::Forms::LinkInput
         @hasLinkName={{true}}
         @linkName={{socialLink.name}}
+        @isSocialLink={{true}}
         @inputId={{socialLink.identifier}}
         @segmentedLink={{socialLink.segmentedLink}}
         @isChild={{true}}
```
snitin315, line 22: @iamareebjamal how can I change `isSocialLink` to `false`, while creating a new record?

iamareebjamal: I don't understand. While creating a new record? You mean while dynamically adding a new LinkINput?

snitin315: Yes

iamareebjamal: You can check `socialLink.id`

It'll be null for a new record. So, instead of true, you can use that

snitin315:
custom-link

mariobehling:
Are all the link displays working on the public event page?

snitin315:
Yes.

Screenshot at 2020-09-09 19-17-31

mariobehling (approved):
Working for tests I did.

mariobehling:
The problem is that these changes always have some influence on areas that are not mentioned in the PR. I am no longer able to nullify (delete) admin social links at https://eventyay.com/admin/content, e.g. cannot delete the facebook link.

Issue #4947, opened 2020-09-03
LinkInput crashing after recent changes


iamareebjamal:
1. Open extra details in Event Dashboard
2. Social Link component won't be rendered
3. See errors in console

iamareebjamal:
The component should be refactored to computed properties instead of observers in future

snitin315:
> Open extra details in Event Dashboard

I can't find the `extra details` section in the Event Dashboard, can you please share the URL?

iamareebjamal:
Second step in event wizard

PR #5038, opened 2020-09-10
fix: ability to remove admin area social links

### LINKED ISSUE #4960: Social System Links functionality URL and Form bugs after Recent Changes in Admin Section

snitin315:
Fixes #4960 

#### Checklist

- [ ] The acceptance, integration, unit tests, and linter pass locally with my changes 

### DIFF
--- a/app/utils/computed-helpers.js
+++ b/app/utils/computed-helpers.js
@@ -5,7 +5,7 @@ import { isValidUrl } from 'open-event-frontend/utils/validators';
 import { FORM_DATE_FORMAT, FORM_TIME_FORMAT } from 'open-event-frontend/utils/dictionary/date-time';
 
 // Social Platforms
-export const socialPlatforms = ['twitter', 'facebook', 'instagram', 'linkedin', 'youtube', 'github', 'gitlab', 'vimeo', 'flicker', 'groups.google'];
+export const socialPlatforms = ['twitter', 'facebook', 'instagram', 'linkedin', 'youtube', 'github', 'gitlab', 'patreon', 'vimeo', 'flicker', 'groups.google'];
 
 /**
  * Get/set a splitted URL from/to a string URL field
@@ -41,8 +41,8 @@ export const computedSegmentedLink = function(property) {
     },
     set(key, value) {
       const finalLink = values(value).join('');
-
-      if (finalLink && isValidUrl(finalLink.trim())) {
+      const isUserNamePresent = finalLink.split('.com/')[1].length > 1;
+      if (finalLink && isValidUrl(finalLink.trim()) && isUserNamePresent) {
         this.set(property, finalLink.trim());
       } else {
         this.set(property, null);
--- a/app/templates/components/forms/admin/content/social-links-form.hbs
+++ b/app/templates/components/forms/admin/content/social-links-form.hbs
@@ -9,7 +9,7 @@
     <label>{{t 'Twitter'}}</label>
     <Widgets::Forms::LinkInput
       @segmentedLink={{this.socials.segmentedTwitterUrl}}
-      @linkname="twitter"
+      @linkName="twitter"
       @isSocialLink={{true}}
       @inputId="twitter" />
   </div>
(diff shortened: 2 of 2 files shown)


iamareebjamal:
Did this happen due to a typo or something else?

iamareebjamal:
I'm still not able to delete the links. Please check again. I need to be able to delete all the characters from the input field and it should nullify the bounded property.

So, for example if from `[https://facebook.com/]eventyay` `eventyay` is removed, meaning now the field is empty, then `facebookUrl` should be set to null

app/components/widgets/forms/link-input.js
```diff
@@ -31,7 +31,6 @@ export default class LinkInput extends Component {
 
   @observes('protocol', 'address')
   protocolAddressObserver() {
-    if (!this.address) {return}
     const link = this.linkName?.toLowerCase();
```
snitin315, line 34: This was not letting us delete the whole `eventyay`, even in UI we removed `eventyay` but `e` was still not removed behind the scenes.

app/components/widgets/forms/link-input.js
```diff
@@ -31,7 +31,6 @@ export default class LinkInput extends Component {
 
   @observes('protocol', 'address')
   protocolAddressObserver() {
-    if (!this.address) {return}
```
iamareebjamal, line 34: This was put here for a reason

app/utils/computed-helpers.js
```diff
@@ -41,8 +41,8 @@ export const computedSegmentedLink = function(property) {
     },
     set(key, value) {
       const finalLink = values(value).join('');
-
-      if (finalLink && isValidUrl(finalLink.trim())) {
+      const isUserNamePresent = finalLink.split('.com/')[1];
+      if (finalLink && isValidUrl(finalLink.trim()) && isUserNamePresent) {
```
snitin315, line 45: this will ensure we are saving only if we have something after `htps://facebook.com/`

app/utils/computed-helpers.js
```diff
@@ -41,8 +41,8 @@ export const computedSegmentedLink = function(property) {
     },
     set(key, value) {
       const finalLink = values(value).join('');
-
-      if (finalLink && isValidUrl(finalLink.trim())) {
+      const isUserNamePresent = finalLink.split('.com/')[1];
```
iamareebjamal, line 44: I don't understand. Why need to add more parsing? Don't we already have a separation between prefix and user-entered value? This is more error-prone. What if the prefix is `google.in`? Or `instagram.com/@`?

snitin315:
admin

app/components/widgets/forms/link-input.js
iamareebjamal: Then please ensure this isn't happening - https://github.com/fossasia/open-event-frontend/issues/4947

app/utils/computed-helpers.js
snitin315: Else it will save `https://facebook.com` in DB.

snitin315: right now all platforms we have defined have `.com`, We do need a better social link validator here though. I will search if I can find something useful.

or an alternative is set the prefix to 'https://' if `this.adress` is `null`

iamareebjamal: I mean don't we have a separate variable for only the text which user has input? I believe that is how we join the prefix and user input

app/components/widgets/forms/link-input.js
snitin315: Looks good to me.

Screenshot at 2020-09-10 19-13-08

app/utils/computed-helpers.js
snitin315: updated 👍🏻

snitin315:
@iamareebjamal please take a look

iamareebjamal:
Looks good but there is one issue. The prefix remains to be `https://` for a predefined link until I start typing in the field. For example, for instagram it will be `https://` but when I start typing it becomes `https://instagram.com/`

It should be `https://instagram.com/` since the beginning. But since we are already behind schedule, this can be added later

iamareebjamal (approved):

PR #5105, opened 2020-09-21
fix: Restore ticket clause for saving event

iamareebjamal:


mariobehling:
Thank you! Worked for me on a test event. I added a follow up issue here: https://github.com/fossasia/open-event-frontend/issues/5110

mariobehling (approved):

maze-runnar:
Save an event as a draft with tickets -> edit event -> delete tickets , I think this is only condition where it is not working. Please try this one.

iamareebjamal:
Draft can exist with no ticket. But should not be able to publish event without tickets

PR #5125, opened 2020-09-23
fix: scroll in one click

### LINKED ISSUE #5123: Public Event Page Menu: Moving to Sponsors, Organizers etc. requires two clicks
To get to sections on the same page that are listed on the menu requires two clicks, e.g. sponsors or organizers.

Compare https://eventyay.com/e/16fa59c7

Screenshot from 2020-09-23 00-38-17

maze-runnar:
Fixes #5123 

#### Checklist

### DIFF
--- a/app/templates/components/public/side-menu.hbs
+++ b/app/templates/components/public/side-menu.hbs
@@ -47,8 +47,8 @@
     {{#if this.event.isSponsorsEnabled}}
       {{#if (eq this.session.currentRouteName 'public.index')}}
         {{!-- template-lint-disable no-nested-interactive  --}}
-        <a href='#sponsor' onClick={{action "scrollToTarget"}} class='item scroll'>
-          <span role="button" onClick={{action 'scrollToTarget'}}>{{t 'Sponsors'}}</span>
+        <a href='#sponsor' {{action "scrollToTarget"}} class='item scroll' >
+          <span role="button" onClick={{action 'scrollToTarget'}} style="display:block;width:100%">{{t 'Sponsors'}}</span>
         </a>
       {{else}}
         <a class="item" href="{{href-to 'public.index'}}#sponsor">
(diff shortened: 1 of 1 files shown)


mariobehling (changes requested):
It does not scroll it jumps. How can we keep the scrolling effect?

maze-runnar:
It jumps only first time, them it scrolls smooth.

mariobehling:
> It jumps only first time, them it scrolls smooth.

So, for the first time the feature works in a different way to following times?

Features should always work in the same way.

maze-runnar:
> > It jumps only first time, them it scrolls smooth.
> 
> So, for the first time the feature works in a different way to following times?

I have tried several ways since two hours, but it is not triggering in first time without `OnClick` and on first click it's not giving smooth scroll, after that it's working.

mariobehling:
We had issues in other projects in the past, where it was a conflict between two JS files. If you cannot find a solution, you could also close the PR and move on to something else.

maze-runnar:
```
<a href='#owner' {{action "scrollToTarget"}} class='item scroll'>
   <button onClick={{action "scrollToTarget"}}>
        {{t 'Organizer'}}
  </button>
 </a>
```
this is working as expected

iamareebjamal:
Please add to the PR so we can test

maze-runnar:
ok

maze-runnar:
this is giving no-nested-interactive and no-invalid-interactive error. One alternate is to use 
`<a href='#sponsor' {{action "scrollToTarget" preventDefault=false}} class='item scroll'>` from the docs. But in fist it it will jump to scroll. After that it working correctly.

iamareebjamal:
Add `role="button"`

maze-runnar:
> Add `role="button"` to anchor tag

it jumps for first time again

iamareebjamal:
Disable lint by adding a comment on that line

iamareebjamal:
It is still jumping for me

maze-runnar:
Actually it will scroll on clicking the text, as span is put on text only.

iamareebjamal:
Add the click listener on the menu item instead

maze-runnar:
> Add the click listener on the menu item instead

not working, jumping from there too

PR #5144, opened 2020-09-25
fix: adding extra social link

### LINKED ISSUE #5142: Wizard Step 2: Add YouTube, VK, Weibo etc. as options for social links
A few options for social links are missing. Please add them to the wizard step 2 and show them also on the live event page, if used by the organizer.

Please add the following and include the logos on the public page where needed:
* YouTube https://youtube.com/
* Weibo https://www.weibo.com/u/
* VK https://vk.com/
* Xing https://www.xing.com

Screenshot from 2020-09-24 20-42-06

maze-runnar:
Fixes #5142 

scrnli_9_25_2020_10-12-57 AM

#### Checklist

### DIFF
--- a/app/components/widgets/forms/link-input.js
+++ b/app/components/widgets/forms/link-input.js
@@ -31,7 +31,7 @@ export default class LinkInput extends Component {
 
   @observes('protocol', 'address')
   protocolAddressObserver() {
-    const link = this.linkName;
+    const link = this.linkName?.toLowerCase();
 
     let add = this.address;
     let proto = this.protocol;
--- a/app/components/widgets/forms/link-input.js
+++ b/app/components/widgets/forms/link-input.js
@@ -56,10 +56,17 @@ export default class LinkInput extends Component {
   linkNameObserver() {
     const link = this.linkName;
     if (socialPlatforms.includes(link)) {
-      this.set('segmentedLink', {
-        protocol : `https://${link}.com/`,
-        address  : ''
-      });
+      if (link==='weibo') {
+        this.set('segmentedLink', {
+          protocol : `https://${link}.com/u/`,
+          address  : ''
+        });
+      } else {
+        this.set('segmentedLink', {
+          protocol : `https://${link}.com/`,
+          address  : ''
+        });
+      }
     } else {
       this.set('segmentedLink', {
         protocol : 'https://',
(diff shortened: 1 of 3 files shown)


mariobehling:
Well, when adding new social links it would be good to have them here the way it is needed. We dont need a function to make all small case or caps. We just need the social links to be written in the way that the brands do it, e.g. YouTube, VK.

I am aware that this is another issue for existing links, but you are introducing new social links that do not appear as expected.

app/components/widgets/forms/link-input.js
```diff
@@ -56,10 +56,17 @@ export default class LinkInput extends Component {
...
     if (socialPlatforms.includes(link)) {
-      this.set('segmentedLink', {
-        protocol : `https://${link}.com/`,
-        address  : ''
-      });
+      if (link === 'weibo') {
+        this.set('segmentedLink', {
+          protocol : `https://${link}.com/u/`,
+          address  : ''
+        });
```
iamareebjamal, line 63: This won't do. It is already highly complex, this should go as a prefix

maze-runnar: should just left as https://${link}.com/ for all

iamareebjamal:
@mariobehling Those are internal identifiers to be saved in DB, thus need to be kept in consistent lower casing for matching purposes, there should be another mapping of the identifiers to proper cased variants

app/components/widgets/forms/link-input.js
iamareebjamal: No, this should not happen here at all

mariobehling:
> @mariobehling Those are internal identifiers to be saved in DB, thus need to be kept in consistent lower casing for matching purposes, there should be another mapping of the identifiers to proper cased variants

So, that should be done in a separate issue? Please proceed as you see fit.

iamareebjamal:
No it will be done in the same PR

iamareebjamal:
Oh sorry, I thought this PR is about changing `twitter` to `Twitter`

Yes, the mapping changes will be done in that PR, not this one

app/components/widgets/forms/link-input.js
```diff
@@ -31,7 +31,7 @@ export default class LinkInput extends Component {
 
   @observes('protocol', 'address')
   protocolAddressObserver() {
-    const link = this.linkName?.toLowerCase();
+    const link = this.linkName;
```
iamareebjamal, line 34: Restore the change please

iamareebjamal (approved):

Issue #5127, opened 2020-09-23
Invoicing Process Implementation


mariobehling:
Implement the invoice process as follows:
1. On the first of the month an invoice gets generated as a PDF in the invoice area of event organizers account.  https://github.com/fossasia/open-event-frontend/issues/5120
a) This only includes organizers who run an active event (not other users or organizers who do not have an event "live").
b) And this includes organizers whose event is finished in the billing cycle.
2. The organizers with live events or events that finished in the billing cycle get an email with a link to the invoice page https://github.com/fossasia/open-event-frontend/issues/5147.
3. On the invoice page is a button "pay now with paypal or stripe"
4. Organizers should click there and pay. The invoice tag on the table changes to "Paid" after the payment.
5. 14 days after the invoice organizers who have not paid yet receive a "Friendly Reminder: Please pay within 15 days" https://github.com/fossasia/open-event-frontend/issues/5147
6. 27 days after the invoice organizers who have not paid yet receive a "Reminder: Please pay within 2 days." https://github.com/fossasia/open-event-frontend/issues/5147
7. After that add a past due email to be sent out after 30 days. https://github.com/fossasia/open-event-frontend/issues/5147
8. The admin account should receive invoice mails in BCC.

iamareebjamal:
All points completed, so closing

PR #5146, opened 2020-09-25
fix: Use correct billing info and PDF URL

iamareebjamal:
For #5127


app/templates/event-invoice/paid.hbs
```diff
@@ -18,11 +18,11 @@
   </div>
   <div class="row">
     <div class="column right aligned">
-      <button {{action 'downloadEventInvoice' this.model.event.name this.model.identifier }}
+      <a href="{{ this.model.invoicePdfUrl }}" target="_blank"
```
iamareebjamal, line 21:
```suggestion
      <a href="{{ this.model.invoicePdfUrl }}" target="_blank" rel="noopener"
```

PR #5302, opened 2020-10-13
fix: make CFS optional like adding org info wizard step 1

### LINKED ISSUE #5300: Wizard Step 5: Make Call for Speakers Optional
The Call for Speakers is currently required for events with sessions and speakers. However, there are events that have sessions and speakers in a schedule, but they do not make a public or private call for speakers. 

Therefore the Call for Speakers section should not be required. There should be a button "Enable Call for Speakers" similar to "Enable Organizer Info" on Wizard Step 1.

Please only provide the enable/disable feature for the "Call for Speakers" area itself. Leave the session form and speakers form area below intact.

Screenshot from 2020-10-13 11-07-50

maze-runnar:
Fixes #5300

#### Checklist

### DIFF
--- a/app/components/forms/wizard/sessions-speakers-step.js
+++ b/app/components/forms/wizard/sessions-speakers-step.js
@@ -194,6 +194,9 @@ export default Component.extend(EventWizardMixin, FormMixin, {
         isComplex : true
       }));
     },
+    resetCFS() {
+      this.set('data.speakersCall.announcement', '');
+    },
     onChange() {
       this.onValid(() => {});
     }
--- a/app/components/forms/wizard/sessions-speakers-step.js
+++ b/app/components/forms/wizard/sessions-speakers-step.js
@@ -126,6 +126,10 @@ export default Component.extend(EventWizardMixin, FormMixin, {
     return this.data.sessionTypes.filterBy('isDeleted', false);
   }),
 
+  hasCallForSpeaker: computed('data.speakersCall.announcement', function() {
+    return !!this.data.speakersCall.announcement;
+  }),
+
   customForm: computed('data.customForms.[]', function() {
     const grouped = groupBy(this.data.customForms.toArray(), customForm => customForm.get('form'));
 
--- a/app/templates/components/forms/wizard/sessions-speakers-step.hbs
+++ b/app/templates/components/forms/wizard/sessions-speakers-step.hbs
@@ -104,75 +104,86 @@
     </div>
 
     <br>
-    <h3 class="ui dividing header">
-      <i class="announcement icon"></i>
-      <div class="content">
-        {{t 'Call for speakers'}}
-      </div>
-    </h3>
+    <div class="ui section divider"></div>
     <div class="field">
-      <label>{{t 'Announcement'}}</label>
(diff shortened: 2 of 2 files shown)


mariobehling:
Was not able to fully test due to Stripe saving issue. Updated branch to see if it works then.

mariobehling (changes requested):
Ok, in regards to the wizard it is working fine. However the CfS is still visible on the public page:

https://open-event-frontend-git-wizard-grey.eventyay.vercel.app/e/5ac8e4d9/cfs

It shows "closed" instead:  
* The page should not be available if the CfS is deactivated.

Screenshot from 2020-10-14 10-54-36

iamareebjamal:
@maze-runnar Please delete CfS if the organizer disables CfS/

maze-runnar:
```
resetCFS() {
      this.data.speakersCall.deleteRecord();
 },
```

after applying this, when I am again trying to change it - 
:` Attempted to set 'startsAt' to '1602618900000' on the deleted record <speakers-call:8>`

mariobehling:
> @maze-runnar Please delete CfS if the organizer disables CfS

Does this mean the data will be deleted? This would be different to other fields where the data is still saved even if the field gets deactivated.

iamareebjamal:
Not deleting the announcement means that we'll have to add another field in DB and API and also add validation that enabling CFS actually contains CFS and disabling CFS actually has no CFS.

mariobehling:
> Not deleting the announcement means that we'll have to add another field in DB and API and also add validation that enabling CFS actually contains CFS and disabling CFS actually has no CFS.

It seems deleting the CfS is the quick solution, ok. But could you add an issue to follow up here, cause it would be better if the announcement would not be deleted when it gets deactivated. Reason is: Organizers play around with features. Also they could click a field by accident. There are many reasons organizers would not want data (like the announcement) to be deleted. So, which deleting the announcement solves the current issue quickly it should not be a final fix.

As a matter of fact if we delete the data we should actually have a pop up "You are deleting the announcement. Yes/No". This is all too much to go through. From the user perspective the best would be if the data does not get deleted simply.

So, I suggest:
1. Delete it for now.
2. Open an issue and change the functionality in a way that no data gets deleted when a field gets deactivated (just like most of the other fields).

Does this work?

iamareebjamal:
I have created the issue and we'll do that in future, but the problem I was talking about, desynchronization of the field and data already exists in the system for most fields. That's why there were so many errors in the past about:

- Stripe checkbox and stripe connection checked and not connected
- Sessions present but not showing on public
- Sessions not present but showing on public
- Schedule present but not showing on public
- Schedule not present but showing on public
- Speakers present but not showing on public
- Speakers not present but showing on public

So, adding a switch which is independent of data always creates edge cases which may eventually be fixed after many iterations, but are fragile. For example, above issues are now fixed on frontend, but there is no check on the backend, so it is entirely possible via API to say the event accepts stripe for payment but not add the connection. And adding such checks in server is very difficult because the change in stripe auth and connection may happen before or after each other and thus create race problems.

So, we'll add that feature in near future but it comes at development and bug cost.

mariobehling:
I understand. I wonder how others are doing it. An alternative would be to add a message "do you really want to do this?" every time a user switches off a feature. This was how Windows worked years ago and users hated it. So, it seems to me we need to go the way as described by you.

PR #5319, opened 2020-10-15
fix: add overflow to notify-modal

### LINKED ISSUE #5308: Speaker Notification: HTML code in message and pop up not entirely visible
The pop up notification box has some issues:
1. There is html code on the bottom of the message
2. The pop up is partly hidden
* behind an area that should be below on top and bottom
* buttons are hidden
* top area is hidden

Compare https://eventyay.com/events/5ac8e4d9/sessions/all
Screenshot from 2020-10-14 08-22-29

maze-runnar:
Fixes #5308 
Screenshot from 2020-10-15 16-51-34

#### Checklist


app/components/modals/session-notify-modal.hbs
```diff
@@ -2,7 +2,7 @@
 <div class="header">
   {{t 'Notify Speakers'}}
 </div>
-<div class="content">
+<div class="content" style="height:400px;overflow-y:scroll">
```
snitin315, line 5: How about `y-scrollable`?

```suggestion
<div class="content y-scrollable" style="height:400px">
```

mariobehling:
Please test this PR with different sessions in different locations. Move the page and click notify. You will see that the pop up is at different locations and when you choose the first item it is off screen on some resolutions.

Screenshot from 2020-10-15 13-50-45

mariobehling:
Do you see the html code below "Thank you"? Still there as well.

maze-runnar:
> Do you see the html code below "Thank you"? Still there as well.
regrading html code @iamareebjamal have it - 
https://github.com/fossasia/open-event-frontend/issues/5308#issuecomment-708275690

iamareebjamal:
That HTML code is link to eventyay. According to the format you sent about the session state change

iamareebjamal:
And about the modal position, this is the semantic UI issue I'm not sure fixable in this PR. This is general problem with any dialog as discussed before

maze-runnar:
there are a lot of issues regarding the single modal issue, that's confusing. I thing there should only be one issue opened not for each for each place.

mariobehling:
If the issue cannot be fixed in this PR then probably the PR can be closed. We need to find a solution for the pop up model as it impacts different areas of the system and renders part of it unusable.

mariobehling:
Or am I missing something. I am unclear what this PR fixes then. 

Regarding the html code: I am afraid an organizer would be confused and delete it. Instead of <a href='{frontend_link}'>{app_name}</a> Could we have something like {app_name_link} maybe?

iamareebjamal:
> I am unclear what this PR fixes then.

The PR fixes the modal part which was cut from up and down because of the message body. How there is a scrollbar so the modal won't be cut.

> Regarding the html code: I am afraid an organizer would be confused and delete it. Instead of <a href='{frontend_link}'>{app_name}</a> could we have something like {app_name_link} maybe?

That can be done. But then it'll be just `eventyay` and not [https://eventyay.com](eventyay)

mariobehling:
> That can be done. But then it'll be just `eventyay` in the email and not [eventyay](https://eventyay.com)

I understood {app_name} is just the app name but can we have {app_name_link} which includes the link instead of the html code?

iamareebjamal:
I think it won't be clickable. It'll most likely be like `https://eventyay.com`, but I'll try it

PR #5337, opened 2020-10-20
register to login

### LINKED ISSUE #5260: Organizer Invitations directs to "Register" instead of "Login" for Existing Accounts and then directs to Start Page
When an event owner invites another organizer and the organizer is not logged in the following happens: The organizer clicks the link and is directed to a sign up page. There is an option "Login instead". It this option is chosen and the user logs in, the user is directed to the start page.

Expected: The invited user (organizer) should be directed to the login page, if the account of the user with the invited email already exists. If the account does not exist the user should be directed to the registration page.

After logging in or registering the user should be directed to the event dashboard that he/she was invited to.

Note: If the user has to register, it should not be necessary to confirm the email after registration, cause the user has already received an email and clicked a link. So, we can immediately add the user to the group of "verified users" in this case.

Screenshot from 2020-10-07 12-48-09

aman-jaglan:
Fixes #5260 

"Login" instead of "Register"

-
-
-

#### Checklist


iamareebjamal:
Wrong

PR #5378, opened 2020-10-25
fix: Set valid CFS time range

### LINKED ISSUE #5340: Call for Speakers: Actions result in error messages at random
The final step of the wizard results in a number of errors at random:
* [ ] Error: When trying to save an event with the "Call for Speakers" switched off, the message "Field cannot be null" appears.
* [ ] Error: When switching on Call for speakers, 
   * [ ] the auto-inserted time is after an event time if the event is less than a month away resulting in the error message. Expected: Auto-inserted dates should be before the event. 
   * [ ] the auto-inserted start and end date show the same day. Expected add a larger date range as auto-inserted dates, e.g. 1 day.

maze-runnar:
Fixes #5340
As the event start date can't be before the current date, so adding cfs date before the current date will solve the issue.
#### Checklist

### DIFF
--- a/app/models/speakers-call.js
+++ b/app/models/speakers-call.js
@@ -10,8 +10,8 @@ const detectedTimezone = moment.tz.guess();
 
 export default ModelBase.extend({
   announcement : attr('string'),
-  startsAt     : attr('moment', { defaultValue: () => moment.tz(detectedTimezone).add(1, 'months').startOf('day') }),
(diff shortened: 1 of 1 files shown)


iamareebjamal:
What about the first point?

maze-runnar:
> What about the first point?

that is not due to CFS, that is due to any other field is empty. I have tested it. CFS is not causing such errors now.

iamareebjamal (approved):

PR #5441, opened 2020-11-01
fix: country dropdown menu should not show random info

### LINKED ISSUE #5438: Country Field in Billing Information: Only show country dropdown, do not automatically fill in random info - only country
* [ ] Fix Country dropdown. It automatically shows data saved in the browsers instead of using the dropdown country option

Expected: Only show country options. Do not show other possible data saved in the browser.

Screenshot from 2020-07-30 10-25-29

ankingcodes:
Fixes #5438 

Fixes the country dropdown which doesn't respond to `autocomplete="no"` in the nested `<input>`

- Changed the `class` of the DropDown to `selection` since the issue is due to the lack of `autocomplete="no"` in the input tag of semantic UI library
- Removed the nested `<Input/>` since it was not used at all.

#### Checklist

### DIFF
--- a/app/templates/components/forms/user-payment-info-form.hbs
+++ b/app/templates/components/forms/user-payment-info-form.hbs
@@ -7,32 +7,20 @@
   </h5>
   <div class="field">
     <label for="company">{{t 'Company'}}</label>
-    <Input
-      @type="text"
-      @id="company"
-      @value={{this.userBillingInfo.company}} />
+    <Input @type="text" @id="company" @value={{this.userBillingInfo.company}} />
   </div>
   <div class="field">
     <label for="contactName">{{t 'Contact Name'}}</label>
-    <Input
-      @type="text"
-      @id="contactName"
-      @value={{this.userBillingInfo.billingContactName}} />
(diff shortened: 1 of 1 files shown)


ankingcodes:
@mariobehling @iamareebjamal - This is ready for review 👍🏻

app/templates/components/forms/user-payment-info-form.hbs
```diff
@@ -57,15 +57,10 @@
   <div class="field">
     <label >{{t 'Country'}}</label>
     <UiDropdown
-      @class="search selection"
+      @class="selection"
```
iamareebjamal, line 60: We want it searchable

iamareebjamal (changes requested):
We want searchable field

app/templates/components/forms/user-payment-info-form.hbs
ankingcodes: Adding `search` would use a nested `<input>` tag which doesn't have `autocomplete="off"`. Is there a way to set that ?

iamareebjamal: Yes, see Tax Info box in Create Event page

Issue #5433, opened 2020-10-30
Account Registration: Get rid of intermediate step after account registration and login new user right away


mariobehling:
When a user registers an account, they have to enter email and password. Right after they register they receive a screen "You have been registered successfully, login...". Instead after a user registers they should be logged into their newly created account right away.

![Screenshot from 2020-10-30 19-43-28](https://user-images.githubusercontent.com/1583873/97745026-86215b80-1ae8-11eb-8b05-ecd9f834c549.png)

suryasiriki4:
Hi @mariobehling, I want to do this, please assign me the task. thank you.

maze-runnar:
@iamareebjamal is this need to be solve on server too?

iamareebjamal:
No

Yash-R:
done please check the PR

PR #5445, opened 2020-11-01
Account Registration: Get rid of intermediate step after account registration and login new user right away #5433

Yash-R:
Done!, Now User would not fill the form again.  @mariobehling

### DIFF
--- a/app/controllers/register.js
+++ b/app/controllers/register.js
@@ -25,7 +25,7 @@ export default class RegisterController extends Controller {
         if (this.inviteToken) {
           this.send('loginExistingUser', user.get('email'), password, this.inviteToken, this.event);
         } else {
-          this.transitionToRoute('login');
+          this.loginExistingUser(user.get('email'), password, this.inviteToken, this.event);
         }
       })
       .catch(reason => {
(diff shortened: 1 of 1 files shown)


app/controllers/register.js
```diff
@@ -25,7 +25,7 @@ export default class RegisterController extends Controller {
         if (this.inviteToken) {
           this.send('loginExistingUser', user.get('email'), password, this.inviteToken, this.event);
         } else {
-          this.transitionToRoute('login');
+          this.loginExistingUser(user.get('email'), password, this.inviteToken, this.event);
```
iamareebjamal, line 28: Both of statements are correct, what's the need of if statement then?

app/controllers/register.js
```diff
@@ -52,7 +52,7 @@ export default class RegisterController extends Controller {
     const authenticator = 'authenticator:jwt';
     this.session
       .authenticate(authenticator, credentials)
-      .then(async() => {
+      .then(async () => {
```
iamareebjamal, line 55: Why

iamareebjamal:
Issue has been fixed

Issue #4584, opened 2020-07-20
Session Form: There is no must for "Track" to be required (wizard step 4)


mariobehling:
The session forms always show "track" to be required, but some events dont even have tracks. Tracks could be required by the organizer, but they should not be required by the system.

![Screenshot from 2020-07-20 19-58-05](https://user-images.githubusercontent.com/1583873/87970094-8b234780-cac3-11ea-8416-d26d517663b6.png)

iamareebjamal:
This is strange because I couldn't find any event without tracks with sessions enabled. If this is indeed intended to be optional, it'll need changes in backend as well because currently it assumes session to always have a track and might break some things when making it optional

mariobehling:
Yes, this is the case because most of the events on the platform using the session form are run by us and these events are bigger. However, a lot of smaller events dont have tracks. Maybe if changes take a lot of time, do other issues first that have a larger impact to improve the system. But, ultimately "tracks" should not be required.

PR #5470, opened 2020-11-03
fix: making track optional

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/4584


iamareebjamal:
Is it working? I don't think so because it requires changes on the server as I said in the issue. Secondly, this needs to be editable in the session custom forms as written in the issue but I see no changes for that in this PR. So what issue is it solving?

maze-runnar:
> Is it working? I don't think so because it requires changes on the server as I said in the issue. Secondly, this needs to be editable in the session custom forms as written in the issue but I see no changes for that in this PR. So what issue is it solving?

it's working. I tried it. But not for the existing events, for existing events it is fixed and that part is needed to fix on the server. But for new events it's working fine as expexted. Please try with creating new Event .

maze-runnar:
Frontend requires this change to make `track field` optional.

iamareebjamal:
Yes but it is not even half of the issue on even the frontend. It does not make the track switchable in custom forms as well

maze-runnar:
for already created events
```
this.store.createRecord('custom-form', {
        fieldIdentifier : 'track',
        form            : 'session',
        type            : 'select',
        isRequired      : true,
        isIncluded      : true,
        isFixed         : true,
        event           : parent
      }),
```
So the events that are already created for them `is_fixed` is true in server-side,  so while editing them `is_fixed` is `true`.
But for new events it's switchable.

maze-runnar:
captured (2)

iamareebjamal:
Can you create a session without a track in a new event? Please show an example

maze-runnar:
> Can you create a session without a track in a new event? Please show an example

no

`{"errors": [{"status": 422, "source": {"pointer": "/data/relationships/track"}, "title": "Unprocessable Entity", "detail": "A valid relationship with track resource is required"}], "jsonapi": {"version": "1.0"}}`

maze-runnar:
What to do for it in server side?

iamareebjamal:
Check the custom form and require track conditionally. It is a bit complicated, so focus on other issues

maze-runnar:
ok

maze-runnar:
Now session can be create  without track.

PR #5472, opened 2020-11-03
Login Box: Show info message when page directs to login page

### LINKED ISSUE #5456: Login Box: Show info message when page directs to login page
If a user wants to perform different actions, they need to login. However, there is no message usually like "You need to login to perform this action." 

Please identify different cases when the login page shows up and add a message for the user, e.g.
* User verified by clicking verification link show "Thank you for verifying your email address. Please login to continue."
* User clicks "Create Event", but is not logged in show "Please login to create an event. If you have not registered yet, please create an account first. Thank you!"
* What other actions require a user to login? Please add cases and messages.

Screenshot from 2020-11-02 14-36-43

Draco9421:
Screenshot from 2020-11-03 19-08-25

Fixes #5456

- If clicked on create event htrough main page or any event page it shows login alert message if not logged in.

#### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - "12"
+  - "14"
 
 env:
   - CXX=g++-4.8 JOBS=1 PATH=$PATH:${HOME}/google-cloud-sdk/bin CLOUDSDK_CORE_DISABLE_PROMPTS=1
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12-alpine as builder
+FROM node:14-alpine as builder
 
 WORKDIR /app
 
@@ -20,7 +20,7 @@ RUN yarn l10n:generate && \
 ##
 ##
 
-FROM node:12-alpine
+FROM node:14-alpine
 
 WORKDIR /fastboot
 
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,7 +1,7 @@
 [build]
   command = "node scripts/l10n.js generate && ember build -prod"
 [build.environment]
-  NODE_VERSION = "12"
+  NODE_VERSION = "14"
 [context.production]
   [context.production.environment]
     API_HOST = "https://open-event-api-dev.herokuapp.com"
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "xgettext-template": "^4.1.0"
   },
   "engines": {
-    "node": "12.x",
+    "node": ">= 12.x",
     "yarn": ">= 1.2.0"
   },
   "private": true,


iamareebjamal:
> Please identify different cases when the login page shows up and add a message for the user,

This is just 1 case. 2 less than what's even mentioned in the issue as examples.

And will you manually add each case here like you have done for the event?

app/templates/login.hbs
```diff
@@ -1,5 +1,15 @@
-<div class="ui one column container stackable doubling centered grid">
+<div class="ui one column container stackable doubling centered grid text-center">
   <div class="column">
-    <Forms::LoginForm />
+    {{#if (not-eq this.session.Passage "/login")}}
+      {{#if (or (eq this.session.previousRouteName '/') (eq this.session.previousRouteName '/create') (eq this.session.Passage '1'))}}
```
iamareebjamal, line 4: Will you add hundreds of cases here like this?

app/routes/application.js
```diff
@@ -130,6 +130,18 @@ export default class ApplicationRoute extends Route.extend(ApplicationRouteMixin
...
+        {
+          this.set('session.Passage', '1')
+        } 
+        else 
+        {
+          this.set('session.Passage', '1')
+        }
+        
+      } else {
+        this.set('session.Passage', '/login');
```
iamareebjamal, line 144: What does this even do?

app/templates/login.hbs
Draco9421: No, but if we need different alert message for different case then I have to make cases for them.

app/routes/application.js
Draco9421: I used it to see that whether someone is trying to login through Login page or he/she is redirected to login page.

app/templates/login.hbs
Draco9421: I will surely try to generalize the cases.

Draco9421: can I use a action method and check every time that which button is pressed and show the message respectively.

iamareebjamal: No, that is too much work

Draco9421: when we verify our email address it automatically login ,So should I change the message which appears on the screen or should I prevent it from login automatically and show that message.

iamareebjamal: The issue is about messages shown on login screen when you are redirected to them

app/routes/application.js
Draco9421: Will this method work ?

iamareebjamal:
I don't think your changes are on track of what you're trying to solve. Or your approach is wrong. See how frameworks like flask and django do it. You are extremely complicating it

Draco9421:
Then what approach I should follow ? Any suggestions please.

iamareebjamal:
>   See how frameworks like flask and django do it.

app/routes/application.js
iamareebjamal: You have written this. How in the world would I know?

Issue #4626, opened 2020-07-23
Display of Error Message too Short and often not Easy to be Seen


mariobehling:
Error Message often appear as a pop up on the bottom of the page and they disappear to fast. The user is also not able to copy error messages.

Please suggest and implement a better way for error messages, that:
* Do not disappear until the user has resolved them (if the user is able to resolve them at all)
* Do not show up as a pop up, but instead show up on the page itself where the user sees them right away (or e.g. when the error appears auto-scroll the page automatically to the error message on the page itself)

![Screenshot from 2020-07-23 20-10-56](https://user-images.githubusercontent.com/1583873/88322204-a8088680-cd20-11ea-9f21-d8bdb5234434.png)

snitin315:
We can disable auto-hiding of these notifications by setting `closeAlert: null`, or we can increase the time duration by default it's 2.5 seconds`
https://github.com/adopted-ember-addons/ember-notify#examples

iamareebjamal:
This should follow the UX where if a user hovers over the message, the message timer stops and message does not dismiss until the user removes the mouse from the message. Secondly, message size and position should be changed more appropriately

ikartikgautam:
Hii, Can i take this up and fix this ? please assign it to me. Thank you

snitin315:
Go ahead

PR #5473, opened 2020-11-03
fix: Improving display of error message 

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/4626

- [x]   if a user hovers over the message, the message timer stops and message does not dismiss until the user removes the mouse from the message.
- [x]  message size and position should be changed more appropriately.

#### Checklist

### DIFF
--- a/app/styles/components/extras.scss
+++ b/app/styles/components/extras.scss
@@ -10,3 +10,14 @@ a.link-item {
     background-color: rgba(0, 0, 0, .05);
   }
 }
+
+.notification-container {
+  font-size: 15px;
+  left: 40%;
+  margin: auto;
+  max-width: 600px;
+  position: fixed;
+  right: 50%;
+  top: 5px;
+  z-index: 1060;
+}
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -26,4 +26,4 @@
   </div>
 </SideBar>
 <Modals::ConfirmModal @isOpen={{this.confirm.isOpen}} />
-<EmberNotify @messageStyle="semantic-ui" @defaultClass="notification-container" />
+<EmberNotify @messageStyle="semantic-ui" @defaultClass="notification-container" @closeAfter={{4000}}/>


iamareebjamal:
Before After screenshots?

maze-runnar:
captured (1)

iamareebjamal:
I think top right will be better place for it

mariobehling:
Well, this example is not an error message. Specifically at this function of switching of tabs I think we do not need an update message or if we need it, it should be similar to Eventbrite where there is a loading/saving icon while the next tab is loading.

mariobehling:
Basically, the question about error message is, if they should always be in the same space and it they should happen as a pop up.

How are other projects handle this?

mariobehling:
So, I think we have two kinds of errors.

1. Errors of form elements. They should work in the same way as e.g. Google forms.  If you have an error on a Google form and you click submit, it will show the error at the form element and jump to the area where there is an error.

2. There are error of the system. And here are two different kinds as far as I can think of:
a) An error that comes up on a newly loading page that loads because of some action and here an error appears. This error should appear on the page itself, not as a pop up. It should appear just below the top menu.
b) Secondly there could be errors that appear without a new page getting loaded, e.g. if the user clicks "Next" and there is a system error preventing it of happening. If that happens, the error should appear above the "Cancel, Save, Next" line.

iamareebjamal:
Those errors/messages are different. This PR doesn't change any of those. The issue you raised here https://github.com/fossasia/open-event-frontend/issues/4626 is about the error message popups "Display of Error Message too Short and often not Easy to be Seen"

About the second point in the issue:

> Do not show up as a pop up, but instead show up on the page itself where the user sees them right away (or e.g. when the error appears auto-scroll the page automatically to the error message on the page itself)

That is not possible for these messages as they are mainly unresolvable errors or from the server, the frontend doesn't know why they happened, so we cannot show them in any part of the UI.

> Secondly there could be errors that appear without a new page getting loaded, e.g. if the user clicks "Next" and there is a system error preventing it of happening. If that happens, the error should appear above the "Cancel, Save, Next" line.

Not every page has these buttons and it will be difficult to handle this in a generic way, but if it needed for specific pages, a dedicated issue should be opened for that. We cannot do away with popup messages, so what should be done about that. Is the current UI OK? Then we can close the PR but the issue will still remain "Display of Error Message too Short and often not Easy to be Seen". Every site shows unrecoverable errors in form of popups, so this PR handles the case for where and for how long the messages which are not tied to any UI have to be shown

mariobehling:
Ok, this issue is more complex than anticipated.

How about we start about this: There is still the point that actually this PR is actually changing system messages and not just taking care of error messages.

So, for system messages like "event has been saved", it is actually ok enough at the moment. Even though it would be nicer especially for the wizard to implement it like Eventbrite, but for error messages we need an improvement. However, we also achieved to reduce the error messages a lot. So the issue is not as high priority anymore. 

We need to think this through and I also think we can close this PR and label the issue "question" to consider. We need to look more like how others are doing it as well.

PR #5504, opened 2020-11-06
fix: remove intermidiate step of login while registering

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/5433

#### Checklist

### DIFF
--- a/app/controllers/register.js
+++ b/app/controllers/register.js
@@ -22,11 +22,8 @@ export default class RegisterController extends Controller {
     this.model.save()
       .then(user => {
         this.set('session.newUser', user.get('email'));
-        if (this.inviteToken) {
           this.send('loginExistingUser', user.get('email'), password, this.inviteToken, this.event);
-        } else {
-          this.transitionToRoute('login');
-        }
+          this.transitionToRoute('/');
       })
       .catch(reason => {
         if (reason && Object.prototype.hasOwnProperty.call(reason, 'errors') && reason.errors[0].status === 409) {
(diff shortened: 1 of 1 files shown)


app/controllers/register.js
```diff
@@ -22,11 +22,8 @@ export default class RegisterController extends Controller {
     this.model.save()
       .then(user => {
         this.set('session.newUser', user.get('email'));
-        if (this.inviteToken) {
-          this.send('loginExistingUser', user.get('email'), password, this.inviteToken, this.event);
-        } else {
-          this.transitionToRoute('login');
-        }
+        this.send('loginExistingUser', user.get('email'), password, this.inviteToken, this.event);
+        this.transitionToRoute('/');
```
iamareebjamal, line 26: Why is this needed? Shouldn't the above action be awaited?

app/controllers/register.js
```diff
@@ -60,7 +57,9 @@ export default class RegisterController extends Controller {
             await this.store.findRecord('user', tokenPayload.identity)
           );
         }
-        this.transitionToRoute('public.role-invites', eventId, { queryParams: { token } });
+        if (token != null) {
+          this.transitionToRoute('public.role-invites', eventId, { queryParams: { token } });
+        }
```
iamareebjamal, line 62: Shouldn't the `this.transitionToRoute('/');` call be placed here instead in else condition

app/controllers/register.js
maze-runnar: > Shouldn't the above action be awaited?

why? I am not getting it.

app/controllers/register.js
maze-runnar: ok

app/controllers/register.js
```diff
@@ -60,7 +56,11 @@ export default class RegisterController extends Controller {
             await this.store.findRecord('user', tokenPayload.identity)
           );
         }
-        this.transitionToRoute('public.role-invites', eventId, { queryParams: { token } });
+        if(token != null) {
```
iamareebjamal, line 59:
```suggestion
        if (token) {
```

PR #5532, opened 2020-11-09
fix: Display sponsor size according to their level

### LINKED ISSUE #5455: Sponsors: Organizers wants to have bigger logos for main sponsors
In our sponsors section we need an option to have a two column sponsor section with bigger logos. There could be a tick box "Large Image" to achieve this.

Question: Do we need an option in the sponsor form to show that large logos are in a two column area? Or can we do it in a way that if there are two large logos a third logo simply moves to the next line?

Compare: https://eventyay.com/e/1f720482

Screenshot from 2020-11-02 14-26-10

Large tickbox could be added

Screenshot from 2020-11-02 14-17-13

iamareebjamal:
Fixes #5455

image
image


mariobehling:
Yes, I see this, but it is not better than before.

The problem is some logos are too close together. On the hand some logos are too small and the organizer is either too big or too small depending on where in the list they are. We need a different approach, where we can give the organizer more control about the size of the logos in the user interface.

Let me formulate an approach in detail. I am afraid it is not working so easily.

Issue #5578, opened 2020-11-11
Call for Speakers: Event Showing Up on Frontpage even though CfS is deactivated


mariobehling:
An event is listed in the section "Call for Speakers" on the front page even though the CfS of that event is deactivated. However, the event has speakers and sessions.

![Screenshot from 2020-11-11 11-05-06](https://user-images.githubusercontent.com/1583873/98798307-08c1e900-240e-11eb-93f8-83a49422b7c0.png)

maze-runnar:
on it

PR #5583, opened 2020-11-11
fix: don't cfs filter when cfs is disabled

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/5578

#### Checklist

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -215,7 +215,16 @@ export default class IndexRoute extends Route {
           op   : 'ge',
           val  : params.start_date
         }
-      }
+      },
+      {
+        name : 'speakers-call',
+        op   : 'has',
+        val  : {
+          name : 'announcement',
+          op   : 'ne',
+          val  : ''
+        }
+      },
     ];
 
     return hash({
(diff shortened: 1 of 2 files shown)


app/routes/index.js
```diff
@@ -215,6 +215,15 @@ export default class IndexRoute extends Route {
...
           val  : params.start_date
         }
+      },
+      {
+        name : 'speakers-call',
+        op   : 'has',
+        val  : {
+          name : 'announcement',
+          op   : 'ne',
+          val  : ''
```
iamareebjamal, line 225: If speaker call is being disabled, announcement should be set to null, not empty string

maze-runnar: it's giving error `field may not be null`.

iamareebjamal: Make it nullable in server please

maze-runnar: `announcement = db.Column(db.Text, nullable=True)` do i have to migrate also after making it nullable.

iamareebjamal: Any change in schema means migration

iamareebjamal (approved):

PR #5584, opened 2020-11-11
 fix: Team Page Empty email fields

### LINKED ISSUE #5526: Team Page : Empty email fields
**Bug**
Whenever an organiser tries to add people with empty input fields, the frontend shows the empty fields after submission.

**To Reproduce**
1. Go to Events Section
2. Create an Event
3.  Go to teams section
4. Add people with empty fields
5) See the bug

**Expected behaviour**
When an owner tries to add another organizer without providing an email it should not be possible submit the form. The form should just show "field required" or similar error message.

**Screenshot**
Screenshot 2020-11-08 at 1 41 21 PM

daretobedifferent18:
Fixes #5526 

Fixed showing empty email fields in the team.

- when we click on add people and then click on side of the screen, the empty entry that was created for creating the people still appears which is fixed by adding an if statement

#### Checklist


daretobedifferent18:
@iamareebjamal Please see to the changes I have made.
These changes were made only in UI bcz the empty entries on reloading page is no longer shown but only shown when being on same page and clicking add people and then clicking on backdrop.

iamareebjamal:
That's not a fix. The entry should be deleted if someone clicks out of the window

daretobedifferent18:
@iamareebjamal Please see the changes I have made and let me know to change anything.

app/components/modals/modal-base.js
```diff
@@ -15,6 +15,9 @@ export default UiModal.extend({
     if (this.isOpen) {
       $element.modal(this.defaultOptions).modal('show');
     } else {
+      if(this.unloadRecord){
```
iamareebjamal: Please remove. You already have a callback for modal closing. Makes no sense to put logic of user role modal in modal base

daretobedifferent18: @iamareebjamal then the solution to this problem could be disabling clicking outside the window so that the modal won't unless user click on cancel button and clicking on cancel button is already deleting the empty record created.

iamareebjamal: Not really. There are tons of ways to solve it

daretobedifferent18: @iamareebjamal So in the previous solution where I was calling the function when the user click outside the box in the modal-base.Is this not be good enough sol? bcz what actually we want to do is that calling the function when the modal closes.

daretobedifferent18: > Please remove. You already have a callback for modal closing. Makes no sense to put logic of user role modal in modal base

@iamareebjamal and I have removed the function which was calling for modal closing and put it inside modal-base bcz we want same thing to happen even if user click on cancel button or outside the model

daretobedifferent18:
@iamareebjamal I have come with the changes. Please see to the changes I have made now and let me know to change anything

iamareebjamal:
Check every other modal that it's working correctly

daretobedifferent18:
> Check every other modal that it's working correctly

@iamareebjamal I have checked all the modals and every modal is working correctly.

app/components/modals/modal-base.js
```diff
@@ -15,6 +15,7 @@ export default UiModal.extend({
     if (this.isOpen) {
       $element.modal(this.defaultOptions).modal('show');
     } else {
+      this.close();
```
iamareebjamal, line 18: Won't this cause recursive error?

iamareebjamal (approved):

app/components/modals/modal-base.js
daretobedifferent18: No its no causing

PR #5585, opened 2020-11-11
Revert "fix: Display sponsor size according to their level"

iamareebjamal:
Reverts fossasia/open-event-frontend#5532


iamareebjamal:
Some required changes are in this PR. I'll raise another to show static volumes

PR #5597, opened 2020-11-13
fix: Handle www and value changes in link-field

iamareebjamal:


app/models/social-link.js
```diff
@@ -13,15 +13,19 @@ export default ModelBase.extend({
 
   event: belongsTo('event'),
 
-  normalizedName: computed('name', function() {
+  normalizedName: computed('site', function() {
+    return this.site === 'website' ? 'globa' : this.site;
```
iamareebjamal, line 17:
```suggestion
    return this.site === 'website' ? 'globe' : this.site;
```

Issue #5631, opened 2020-11-15
Organizer Ticket Sales Overview: Fix Sales Calculation


mariobehling:
The calculation of the sales on past events is wrong. The numbers do not add up correctly and the sales have many number after the dot.

Even though events are old I assume that the calculation needs to be fixed as the same error could appear on new events. 

Please compare (access for admins only) https://eventyay.com/events/3689bc98/tickets

![Screenshot from 2020-11-15 04-17-31](https://user-images.githubusercontent.com/1583873/99162096-f1c71380-26f9-11eb-9e9a-2102ef6766e3.png)

iamareebjamal:
We need to fix the formatting but the amount is correct

```
select sum(amount) from orders where event_id = 598 and status = 'completed';                                                                                                                                                     
+--------------------+
| sum                |
|--------------------|
| 14768.899999999998 |
+--------------------+
```

PR #5686, opened 2020-11-18
fix: ticket sales value 

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/5631

#### Checklist

### DIFF
--- a/app/helpers/round-up.js
+++ b/app/helpers/round-up.js
@@ -1,0 +1,7 @@
+import Helper from '@ember/component/helper';
+
+export function roundUp(params) {
+  return params[0].toFixed(2);
+}
+
+export default Helper.helper(roundUp);
--- a/app/helpers/round-up.js
+++ b/app/helpers/round-up.js
@@ -1,7 +1,8 @@
 import Helper from '@ember/component/helper';
+import { ceil } from 'lodash-es';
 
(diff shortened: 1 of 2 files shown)


app/helpers/round-up.js
```diff
@@ -0,0 +1,8 @@
+import Helper from '@ember/component/helper';
+import { ceil } from 'lodash-es';
+
+export function roundUp(params) {
+  return ceil(params[0], 2);
```
iamareebjamal, line 5: Use Math.ceil

iamareebjamal:
We already have ember math helpers, use round helper from that

maze-runnar:
using `ceil` helper.

iamareebjamal:
It's not displaying decimals now

maze-runnar:
> It's not displaying decimals now

round is not also displaying decimals. For that I have to use custom helper.

iamareebjamal:
No, you can pass `decimal=2` in the round helper. Check the docs

iamareebjamal:
I'm seeing only 1 digit after decimal

maze-runnar:
> I'm seeing only 1 digit after decimal

if there are more than two or two digits after decimal then it will show two digits after decimal. Otherwise wise original number

PR #5718, opened 2020-11-23
fix: Keep data saved when switching on/off Forms in Wizard steps

### LINKED ISSUE #4605: Forms: Keep data saved when switching on/off Forms in Wizard step 1, 2, and 3
Switching off the options in the wizard in step 1, 2 or 3 results in the data being erased.

Expected: Keep the data saved in the system, but don't show it publicly if the page/option is switched off in the wizard.

daretobedifferent18:
Fixes #4605 

Keep data saved when switching on/off Forms in Wizard step 1, 2, and 3

#### TODO 
- [x] Organizer Saved
- [x] Sponsor Step Saved
- [x] Attendee Form Saved
- [x] Session Speaker Form Saved

#### Checklist


iamareebjamal:
> but don't show it publicly if the page/option is switched off in the wizard.

Doesn't work

daretobedifferent18:
> > but don't show it publicly if the page/option is switched off in the wizard.
> 
> Doesn't work

for the meantime only sponsor switched on/off button is working , for rest I'm working

daretobedifferent18:
@iamareebjamal after not showing attendee fields when form is switched OFF and doing other work, the form was not being submitted, giving an error and I found this -

For Attendee form what I think that we have to do, to work it smoothly is that if we switched OFF attendee custom form we are saving data but with that some among them are required fields, which are also saved so now while filling ticket orderForm (attendee form OFF) , the form does not submit until and unless all the required fields that we have switched on in attendee form is filled and also now we don't have to show those fields when attendeeform is switched OFF so we don't have any option.

So what I think could be done is when we switched OFF the attendee form, the ticked fields will be saved but the required option will be disabled then for all those fields and as these fields are not even shown, so this is working good.

And now on switching back ON form, we have all those fields that we ticked earlier back, but not with the required option tick, so for that user can do that again but fields are ticked.

Pls let me know about this.

daretobedifferent18:
@iamareebjamal All the switched cases are working except for COC bcz switched ON/OFF depends on whether we have provided codeOfConduct text or not and if once provide and saved, even after switching Off, it returns back to ON which is already a bug.

daretobedifferent18:
> @iamareebjamal All the switched cases are working except for COC bcz switched ON/OFF depends on whether we have provided codeOfConduct text or not and if once provide and saved, even after switching Off, it returns back to ON which is already a bug.

what we can do for it is similar to the behaviors of what cfc announcement does when cfc is diabled.

daretobedifferent18:
@iamareebjamal Pls let me know

daretobedifferent18:
@iamareebjamal Pls review. I think this is Done and if anything to do then pls let me know

app/templates/events/view/index.hbs
```diff
@@ -2,15 +2,17 @@
...
-      <Events::View::Overview::EventSponsors
-        @data={{this.model.sponsors}}
-        @columns={{this.columns}}
-        @page={{this.page}}
-        @per_page={{this.per_page}}
-        @search={{this.search}}
-        @sort_by={{this.sort_by}}
-        @sort_dir={{this.sort_dir}}
-        @filterOptions={{this.filterOptions}} />
+      {{#if this.showSponsors}}
```
maze-runnar, line 5: why didn't used `{{if this.model.event.isSponsorsEnabled}}` directly here inside if, why to make a computed function?

maze-runnar (commented):
working fine

app/templates/events/view/index.hbs
daretobedifferent18: I haven't noticed that. Yup we can do that

daretobedifferent18:
@maze-runnar I have made the changes Pls have a look at it.

maze-runnar (approved):

app/templates/public/index.hbs
```diff
@@ -45,7 +45,7 @@
     </div>
     <div class="ui hidden divider"></div>
   {{/if}}
-  {{#if this.model.sponsors}}
+  {{#if this.model.event.isSponsorsEnabled}}
```
iamareebjamal, line 48: What if event sponsors is enabled and there are no sponsors?

daretobedifferent18:
@iamareebjamal Done the Changes working fine now

daretobedifferent18:
@iamareebjamal working fine even when sponsors is enabled with no sponsors. I think no more changes needed for this PR

Issue #5299, opened 2020-10-13
Video: Autogenerate underlying Jitsi Room and Phone numbers


mariobehling:
There is a way to autogenerate a Jitsi room with "additional information" such as phone numbers. An example of an implementation is available in the calendar chrome plugin at https://github.com/jitsi/jidesha.

Please implement this in a way that such "Additional Information" is added automatically into the field "Additional Information".

![Screenshot from 2020-10-13 10-49-11](https://user-images.githubusercontent.com/1583873/95838201-c9867680-0d41-11eb-95a9-fa3deca779d4.png)

iamareebjamal:
https://jitsi.org/api/

utkarsh147-del:
I want to work on this issue.Please give this work to me and a chance for contribution

Issue #5101, opened 2020-09-20
Video Tab: Implement Feature "Add Jitsi Room" to Microlocations


mariobehling:
Organizers should have two options to add a video room 
a) add a room manually anywhere on the Internet
b) add a Jitsi room with a button.

According to the Jitsi documentation here https://jitsi.github.io/handbook/docs/dev-guide/dev-guide-iframe the following should be possible.

Open question is: In order to restrict access to rooms we could also add rooms into an 100% iframe on eventyay.com and restrict access to account/ticket holders.

Please implement the option to automatically create a Jitsi room as follows on the video tab:
- Microlocation: add a button "Add Jitsi Room" and autofill other columns
- Video Room: After user has pushed "Add Jitsi Room" automatically create a virtual room at https://meet.jit.si/eventyay/[eventID-room-name]
- Pin: Automatically create a room password and add info here
- Additional Information: Automatically add Jitsi dial-in numbers etc. into this column
- Stream: Apart from the existing public stream URL, also show a field for the Streamkey here, that organizers can enter and pass on the stream key to Jitsi to enable the feature "Stream to Youtube" 
- In field "Additional Information" add: In order to stream a Jitsi Session the moderator in the session needs to start the stream.

Other features:
* Mute all participants in room when they join
* remove the "Embed Jitsi Meet Conference"
* Hide "Invite More People" using HIDE_INVITE_MORE_HEADER (Reference https://github.com/jitsi/jitsi-meet/pull/7021)
* Hide "not secure icon"

![Screenshot from 2020-09-19 14-16-04](https://user-images.githubusercontent.com/1583873/93666971-bbb33d80-fa82-11ea-816a-0306769e2648.png)

iamareebjamal:
Assuming this in the implementation: A room can have a single video stream linked. But a video stream can be linked to several rooms

PR #5720, opened 2020-11-23
feat: Add Jitsi Integration

iamareebjamal:
- [ ] Show password 
- [x] and additional info

#5299 #5101


mariobehling:
Very good. Works for me. We can iterate on this in follow-up issues including:
* Jitsi room has scroll bars in some resolutions
* Also need to add a text link "Join Video Channel" into the description above "Additional Info"
* Change Video Icon to a button "Join Video"

mariobehling (approved):

PR #5742, opened 2020-11-24
fix: Reimplement session route to not reload page on filter

### LINKED ISSUE #4882: Reimplement sessions route to not reload page on filter
I see whenever we change a filter or date the page goes up to the top of the page. How can we go to the top of the event page heading instead, e.g. here the heading "Sessions"?

_Originally posted by @mariobehling in https://github.com/fossasia/open-event-frontend/pull/4879#issuecomment-680685934_

nikolaytoplev:
Fixes #4882

Implement going to session heading after filtering

-Add observer to observe session data of model
-Whenever session is changed on filtering, animate body element to go to session heading. 
-Session template is modified for adding session heading id.

#### Checklist


mariobehling:
Thank you. Before going next steps, please resolve merge conflicts.

nikolaytoplev:
> Thank you. Before going next steps, please resolve merge conflicts.

Resolved.

app/routes/public/sessions.js
```diff
@@ -87,6 +87,8 @@ export default class SessionsRoute extends Route {
       });
     }
 
+    const isFiltering = (params.room === null && params.track === null && params.data === null) ? false : true;
```
iamareebjamal, line 90: First, shouldn't it be params.date and not params.data?

```suggestion
    const isFiltering = params.room || params.track || params.date;
```

And please check if this will work

nikolaytoplev: Yeah, I tried as you proposed at first, but if any one factor is 0(zero), isFiltering is evaluated to false.
So to get correct filtering info, I think my code should be used, but params.data is my mistake.
Is it okay to change only params.data to params.date?

iamareebjamal: Neither room, track or date can be 0, so my recommendation will make the code very small

nikolaytoplev: If you think it is better, let's go on that way.

PR #5776, opened 2020-11-26
Handled bug for unnecessary comma

### LINKED ISSUE #5774: Schedule: Not needed comma if no position filled in by speaker
On the schedule a comma and space show up even though they are not required. This happens if the speaker has not filled in a position. Expected no comma and space should show up if the speaker has not filled in a position. Equally if the speaker has not filled in a company no comma and space should show up.

Compare https://eventyay.com/e/3dbaaa50/schedule

Screenshot from 2020-11-26 12-51-11

Gautam-Arora24:
Fixes #5774

**Changes proposed in this pull request:**
1) No comma needed if there is no position filled

**Before**
100347884-634e6e00-2fe6-11eb-881d-1995adc460e5

**After**
Screenshot 2020-11-26 at 8 08 50 PM

### DIFF
--- a/app/components/public/session-item.hbs
+++ b/app/components/public/session-item.hbs
@@ -40,7 +40,15 @@
            </div>
            <div class="left floated nine wide column">
              {{#each @session.speakers as |speaker|}}
-                {{speaker.name}} ({{speaker.position}}, {{speaker.organisation}})
+                {{speaker.name}}(
+                  {{#if speaker.position}}
+                      {{speaker.position}}
+                      {{#if speaker.organisation}}               
+                      , {{speaker.organisation}}
+                      {{/if}}
+                    {{else}}
+                   {{speaker.organisation}}
+                  {{/if}})
                 <br>
               {{/each}}
             </div>


app/components/public/session-item.hbs
```diff
@@ -40,7 +40,15 @@
...
-                {{speaker.name}} ({{speaker.position}}, {{speaker.organisation}})
+                {{speaker.name}}(
+                  {{#if speaker.position}}
+                      {{speaker.position}}
+                      {{#if speaker.organisation}}               
+                      , {{speaker.organisation}}
+                      {{/if}}
+                    {{else}}
+                   {{speaker.organisation}}
+                  {{/if}})
```
iamareebjamal, line 51: Too complex

Gautam-Arora24: Just if else blocks

iamareebjamal: `[a, b, c].filter(Boolean).join(', ')`

maze-runnar: use inline if condition with a minimal set of condition and avoid nesting like above.

Gautam-Arora24: Can you explain what this piece of code does?

daretobedifferent18 (commented):
```handlebars
 {{speaker.name}}
 {{#if (or speaker.position speaker.organisation)}}
        ({{speaker.position}}{{if (and speaker.position speaker.organisation) ', '}}{{speaker.organisation}})
 {{/if}}
```
You can use like this as well

app/components/public/session-item.hbs
iamareebjamal: Joins multiple strings after removing null

iamareebjamal:
> ```
>  {{speaker.name}}
>  {{#if (or speaker.position speaker.organisation)}}
>         ({{speaker.position}}{{if (and speaker.position speaker.organisation) ', '}}{{speaker.organisation}})
>  {{/if}}
> ```
> 
> You can use like this as well

Not needed, `[speaker.position, speaker.organization].filter(Boolean).join(', ')` will work

daretobedifferent18 (commented):
Also need to care of the fact that when neither position is provided nor organization, brackets shouldn't be there as well.

Gautam-Arora24:
Sure, I am changing the PR as proposed

mariobehling:
Your example screenshot already shows the issue.

Correct would be, if it shows:

Ridhi (Student)

Gautam-Arora24:
@mariobehling Are you asking about this?


Screenshot 2020-11-26 at 8 08 50 PM

mariobehling:
Repeat

> Your example screenshot already shows the issue.
> 
> Correct would be, if it shows:
> 
> Ridhi (Student)

PR #5777, opened 2020-11-26
feat: Reduce front page queries and cache more

iamareebjamal:


iamareebjamal:
@daretobedifferent18 Please test if everything is working. Also try changing name of user and reloading

daretobedifferent18 (commented):
@iamareebjamal It is working fine and also I have checked after changing the user name and reloading the page. It is working fine.
But I found I bug maybe it was before also
Actual | Expected
--------- | ---------
Screenshot from 2020-12-07 22-45-25 | Screenshot from 2020-12-07 22-45-45

And this bug is removed only after reloading the page

iamareebjamal:
That should not be related to this PR. Is it?

iamareebjamal:
Not experiencing this myself, but if you find a way to reproduce, please raise an issue. Merging it and will revert if we find any bugs

daretobedifferent18:
> That should not be related to this PR. Is it?

shouldn't it be ?
I checked all the working and it is working fine and I just encountered with this problem.

iamareebjamal:
No, this PR doesn't touch schedule

Issue #5796, opened 2020-11-27
Change general-date helper format argument


iamareebjamal:
Most of the formats passed to general-date are now standardized. There is no need to pass big repeated formats, it there are 3-4 of those. So, format argument should accept certain keywords and resolve them to formats, like `{{general-date date 'dd MM YYYY HH:mm (z)'}}` should be treated same as ``{{general-date date 'date-time-long'}}``

This will also make changing specific formats easier in future.

If the passed in format is not a known keyword, it'll be treated as literal format. And format will be optional, with the most used format being the default

daretobedifferent18:
Working on it

daretobedifferent18:
working on it

PR #4794, opened 2020-08-18
fix: Session list date format with timezone

iamareebjamal:
Fixes #4238

PR #5798, opened 2020-11-27
fix: Timezone Display shows numbers instead of timezone name

### LINKED ISSUE #5749: Timezone Display shows numbers instead of timezone name
Some events display the timezone while others display e.g. (+08). Please change the implementation in a way that it always displays timezones correctly in all areas including on:
* Frontpage
* Events Dashboard
* Specific event dashboard
* Public event pages
* Admin pages

Screenshot from 2020-11-25 09-30-27l a
Screenshot from 2020-11-25 09-29-50

Screenshot from 2020-11-25 09-30-49

Screenshot from 2020-11-25 09-31-42

daretobedifferent18:
Fixes #5749

Some events display the timezone while others display e.g. (+08). Please change the implementation in a way that it always displays timezones correctly in all areas.

#### Checklist


daretobedifferent18:
@iamareebjamal Pls review this, its done

iamareebjamal:
Once we do #5796, it'll be much simpler to do

iamareebjamal:
I'll discuss the approach to be used in the meeting at 7 PM

daretobedifferent18:
@iamareebjamal I was searching for the approach u asked for and I found some useful information which might be useful. 
https://github.com/moment/moment-timezone/issues/477
https://github.com/moment/moment-timezone/issues/500

iamareebjamal:
But what's the solution?

daretobedifferent18:
> But what's the solution?

Solution could be defining all the country codes manually in an array and taking the country code from there bcz acc to the issue i linked in the above comment, they have removed country code bcz those country don't use them

iamareebjamal:
Did you try `moment.tz(params[0], timezone).zoneAbbr()`

See what it returns

iamareebjamal:
I changed it from that to (z) and it has been broken since https://github.com/fossasia/open-event-frontend/pull/4794/files

daretobedifferent18:
> I changed it from that to (z) and it has been broken since https://github.com/fossasia/open-event-frontend/pull/4794/files

ok I'll see to it

iamareebjamal:
Also check this if it doesn't work https://gist.github.com/redoPop/3915761

daretobedifferent18:
@iamareebjamal I have tried with both the approach, they both are giving numbers only and not tz abbr

iamareebjamal:
Then get abbr and if it doesn't give code back, use a dict of mapping tz codes. Dict should only contain those timezones which are not mapped in moment

daretobedifferent18:
> Then get abbr and if it doesn't give code back, use a dict of mapping tz codes. Dict should only contain those timezones which are not mapped in moment

Okay so what I should do now is create a dict for all those tz where moment is giving back number instead of tz abbr and then get abbr from moment and if it gives number then take tz abbr from dict I will be creating.

daretobedifferent18:
And what I found is that there are 226 tz whose tz Abbr is number.

daretobedifferent18:
@iamareebjamal 
I was thinking of the approach of getting country code from moment and then concating it with 'T' but the results are not coming correct for all. So I think the approach left is what we discussed above.

daretobedifferent18:
@iamareebjamal could you pls tell me from where I should reference timezone abbr for left out countries.

iamareebjamal:
https://www.timeanddate.com/time/zones/

app/helpers/general-date.js
```diff
@@ -25,12 +26,14 @@ export function generalDate(params, { tz }) {
...
-  if (reg.test(z)) {
-    dateTime = dateTime.replace(z, timezone);
+  const formatWithoutTz = format.replace(' (z)', '');
+
+  let dateTime = moment(params[0]).tz(timezone).format(formatWithoutTz);
+
+  const timezoneAbbr = tzAbbr[timezone] || moment(params[0]).tz(timezone).format('z');
+
+  if (format.includes('(z)')) {
+    dateTime = dateTime.concat(` (${timezoneAbbr})`);
```
iamareebjamal, line 36: Don't do this much of replacing and concat

Just remove (z) from the format literal, and append the abbreviation if the passed in code format was among the tz variants

daretobedifferent18: and then what about default case?

iamareebjamal: I thought we moved all timezone displaying strings in keywords. That was the goal of this PR

daretobedifferent18: > I thought we moved all timezone displaying strings in keywords. That was the goal for this PR

oh really, if that's the case then I can do that

iamareebjamal: Please check once as well

daretobedifferent18: goal for this or for the other timezone add argument tz one PR ?

daretobedifferent18: Done with all tz display string keywords

daretobedifferent18:
@iamareebjamal Its Done. Pls let me know to make any changes further

app/components/public/add-to-calender.hbs
```diff
@@ -3,11 +3,10 @@
     <i class="clock alternate outline icon pt-1"></i>
     <div class="content">
       {{#if this.isSingleDay}}
-        {{general-date @event.startsAt "dddd, D MMMM, YYYY" tz=@event.timezone}}<br>{{general-date @event.startsAt "h:mm A" tz=@event.timezone}} {{t 'to'}} {{general-date @event.endsAt "h:mm A" tz=@event.timezone}} 
+        {{general-date @event.startsAt "dddd, D MMMM, YYYY" tz=@event.timezone}}<br>{{general-date @event.startsAt 'time-tz-short' tz=@event.timezone}} {{t 'to'}} {{general-date @event.endsAt "h:mm A" tz=@event.timezone}}
```
iamareebjamal, line 6: Better is to create another keyword like `time-tz-short` -> `tz` and use it to replace `({{this.timezone}})`

app/helpers/general-date.js
```diff
@@ -17,14 +18,23 @@ export function generalDate(params, { tz }) {
...
     format = format.replace(' a', '');
   }
-  return moment(params[0]).tz(timezone).format(format);
+
+  let dateTime = moment(params[0]).tz(timezone).format(format);
+
+  const timezoneAbbr = tzAbbr[timezone] || moment(params[0]).tz(timezone).format('z');
+
+  if (!params[1] || params[1].includes('tz')) {
+    dateTime = dateTime.concat(` (${timezoneAbbr})`);
```
iamareebjamal, line 34: Why concat? Why not just ```return dateTime + ` (${timezoneAbbr})`;```

app/utils/dictionary/tzAbbr.ts
```diff
@@ -0,0 +1,228 @@
...
+  'America/Argentina/Cordoba'        : 'ART',
+  'America/Argentina/Jujuy'          : 'ART',
+  'America/Argentina/La_Rioja'       : 'ART',
+  'America/Argentina/Mendoza'        : 'ART',
+  'America/Argentina/Rio_Gallegos'   : 'ART',
+  'America/Argentina/Salta'          : 'ART',
+  'America/Argentina/San_Juan'       : 'ART',
+  'America/Argentina/San_Luis'       : 'ART',
+  'America/Argentina/Tucuman'        : 'ART',
+  'America/Argentina/Ushuaia'        : 'ART',
```
iamareebjamal, line 15: Please remove any ambiguous entry, i.e., any code which is repeated for 2 timezones

daretobedifferent18: but they all are diff timezones

iamareebjamal: Exactly, and they all show ART, which is ambiguous

daretobedifferent18: I understand all but like 
'America/Argentina/Buenos_Aires'   : 'ART',
'America/Argentina/Catamarca'      : 'ART',

both are unique so removing any on one results in not getting tz abbr

@iamareebjamal 
I want to know that is there a way we can create multiple keys with a single value object?

iamareebjamal: No, just remove them and let the offset show as the timezone like the current situation

When you see CRT, should it be considered China Standard Time or Cuba Standard Time. Hence, remove all entries which have ambiguous abbreviations between two timezones.

daretobedifferent18: ohk like that :+1:
but also I think I should not remove that abbr where they are duplicated but have same full forms like in above ART is used for Argentina Time which is same?

iamareebjamal: No, they can be confusing, better to remove them

daretobedifferent18: Last question like 'TRT' abbr is for 3 of them then which one to keep and which to remove?

iamareebjamal: Remove all which have more than one, so no TRT will be shown, because then how would a person know which of the 3 TRT represents?

daretobedifferent18:
@iamareebjamal Done

app/helpers/general-date.js
```diff
@@ -1,13 +1,15 @@
...
   'date-time-long'     : 'dddd, D MMMM, YYYY h:mm A',
-  'date-time-tz-long'  : 'dddd, D MMMM, YYYY h:mm A (z)',
+  'date-time-tz-long'  : 'dddd, D MMMM, YYYY h:mm A',
   'date-time-short'    : 'D MMM, YYYY h:mm A',
-  'date-time-tz-short' : 'D MMM, YYYY h:mm A (z)',
+  'date-time-tz-short' : 'D MMM, YYYY h:mm A',
   'date-short'         : 'D MMM, YYYY',
-  'time-tz-short'      : 'h:mm A (z)'
+  'time-tz-short'      : 'h:mm A',
+  'tz'                 : ' '
```
iamareebjamal, line 12: Shouldn't it be empty string?

iamareebjamal:
Run `yarn lint`

app/helpers/general-date.js
daretobedifferent18: No bcz then format will be 'tz' which after passing to moment will convert to for eg tz is Asia/Kolkata then will convert to tIST and then after appending timezoneabbr again will finally give ```tIST (IST)```

So just to get format correct when this line is executed 
```js
let format = (dateFormats[params[1]] || params[1]) || 'h:mm A, MMMM Do YYYY';
```

'tz' : ' '

daretobedifferent18: Conclution tz : '' returns false

PR #5799, opened 2020-11-27
chore: Change general-date helper timezone argument

### LINKED ISSUE #5795: Change general-date helper timezone argument
Current general date helper takes timezone as a positional argument, which was chosen to ensure devs pass event timezone always because they didn't in moment format and it displayed wrong time.

But as it is being used in other places where event timezone does not make much sense, like in event sales or order time, it requires empty string to be passed as timezone, which is ugly and unnecessary as it occurs before the format, it cannot be left as is.

Please change timezone to keyword argument, like `{{general-date date tz=timezone}}`

Which will be optional, so invocation without timezone will look like `{{general-date date}}`

daretobedifferent18:
Fixes #5795 

- change the timezone to keyword argument, like {{general-date date tz=timezone}}

#### Checklist


daretobedifferent18:
@iamareebjamal Its Done

iamareebjamal (approved):

PR #5806, opened 2020-11-28
fix: Download ticket and invoice made available to ticket purchaser only

### LINKED ISSUE #5623: Complete Order Page: Implement Per Ticket Download
The complete order page currently shows a button "Download Tickets" for all tickets in an order. This is useful for many cases. However it might be difficult if e.g. a company representative gets tickets for a number of people and then wants to send them their specific ticket or if we want to share rights to access (permissions to) a specific ticket to another account in future.

Therefore please implement a per ticket download additionally to the "Download Tickets" (which downloads all tickets of an order).

Implement this as follows:
* In the "Order Summary SUCCESS" box add a column "Download Tickets" on the very right of the table
* Show a button Download for each ticket (If there are several tickets sold, we need several download buttons)
* Each download button should consist of two areas
  * a) an icon "Ticket" and 
   * b) the name of the ticket holder. 
* If there is more than one ticket in the ticket type then show several buttons with a Ticket icon and the name of the ticket holder. Example: The quantity of tickets sold is - two "Community Tickets", then show two Download buttons with a `Ticket icon and the name of the ticket holder`.
* Finally change the existing "Download Tickets" button to "All Tickets"

Screenshot from 2020-11-13 22-29-04

codedsun:
Fixes #5623 
Fixes https://github.com/fossasia/open-event-server/issues/7397

-
-
-

#### Checklist

### DIFF
--- a/app/templates/orders/view.hbs
+++ b/app/templates/orders/view.hbs
@@ -30,19 +30,21 @@
       <Orders::EventInfo
         @data={{this.model.order}} />
       <br/>
-      <div class="d-flex space-between">
-        <button class="ui labeled icon {{if this.isLoadingTickets 'loading'}} {{if this.device.isMobile 'fluid'}} button"
+      {{#if (eq this.model.order.user.id  this.authManager.currentUser.id)}}
+        <div class="d-flex space-between">
+          <button class="ui labeled icon {{if this.isLoadingTickets 'loading'}} {{if this.device.isMobile 'fluid'}} button"
           {{action this.downloadTickets this.model.event.name this.model.order.identifier}}>
           <i class="ticket alternate icon"></i>
           {{t 'Download Tickets'}}
-        </button>
-        <button {{action 'downloadInvoice' this.model.event.name this.model.order.identifier }}
+          </button>
+          <button {{action 'downloadInvoice' this.model.event.name this.model.order.identifier }}
           class="ui labeled icon blue {{if this.isLoadingInvoice 'loading'}} {{if this.device.isMobile 'fluid'}} button">
           <i class="file alternate icon"></i>
           {{t 'Download Invoice'}}
-        </button>
-      </div>
-      <br/>
+          </button>
+        </div>
+        <br/>
+      {{/if}}
       <Orders::TicketHolder
         @data={{this.model.order}} />
     </div>


codedsun:
@iamareebjamal - I get this as issue when using `this.model.order.user.id`

```
Error: Assertion Failed: You attempted to access the `id` property (of <(unknown):ember679>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('id')` in this case.

If you encountered this error because of third-party code that you don't control,
there is more information at https://github.com/emberjs/ember.js/issues/16148, and
you can help us improve this error message by telling us more about what happened in
this situation.
```

codedsun:
image


image

mariobehling (changes requested):
Thanks, but this is a bit confusing for the user. First the user sees a button. Then the user clicks on the button. Then the use gets the info - "The information is not available". In that case if the user has not bought the tickets, the button should not be there at all. 

Therefore please change this as follows:
* Show the download link of the user ticket (only for that user) in the user area
* For users who have not bought the ticket (just received it from someone else), do not show "All Tickets" and "Invoice" button.

app/controllers/orders/view.js
```diff
@@ -13,6 +13,10 @@ export default class ViewController extends Controller {
 
   @action
   downloadInvoice(eventName, orderId) {
+    if (this.model.order.user.get('id') !== this.authManager.currentUser.id) {
+      confirm('Only the account that made the order is able to access this invoice.');
```
iamareebjamal, line 17: This is browser confirm, not our dialog

app/controllers/orders/view.js
```diff
@@ -44,6 +48,10 @@ export default class ViewController extends Controller {
   }
 
   downloadTickets(eventName, orderId) {
+    if (this.model.order.user.get('id') !== this.authManager.currentUser.id) {
```
iamareebjamal, line 51: This should download the attendee tickets instead

codedsun:
@iamareebjamal 

image

iamareebjamal:
order.user should not be available to the ticket holder

Please open a server issue for this

codedsun:
image

iamareebjamal (approved):

PR #5838, opened 2020-11-30
Attendee Form: Implement Checkboxes #5833

### LINKED ISSUE #5833: Attendee Form: Implement a Checkboxes
Currently the attendee form is missing the option "Checkboxes" for custom questions. Please implement it.

Screenshot from 2020-11-30 10-41-11

akash2408:
Fixes #5833

without
with

-
-
-

#### Checklist

### DIFF
--- a/app/components/forms/wizard/custom-form-input.hbs
+++ b/app/components/forms/wizard/custom-form-input.hbs
@@ -1,10 +1,11 @@
 <div class="ui hidden divider"></div>
 <div class="ui dividing">
-  <i class="checkmark box icon"></i>
-  <span class="content">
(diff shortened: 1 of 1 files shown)


iamareebjamal:
Read the issue again. That is not what the issues says at all. You should also realize the logic of why an issue is raised. Tell me why would we want a checkbox before `Add Custom Form Field` which does absolutely nothing?

maze-runnar:
@akash2408 add checkbox option where you are seeing options, number and text for any custom field.

iamareebjamal:
Just that is not enough. It has server component as well. But even if we only consider frontend component, then also it needs to handle custom form options rendering wherever they are used

PR #5933, opened 2020-12-06
fix: online event and enhance date/time in front page

### LINKED ISSUE #5926: Front Page: Add info about Online event and fix/enhance date/time
The front page nicely displays locations of events. At the same place there should be
* the info "Online Event" if the event is not location based and 
* "Online Event and at [address here]" if it is a mixed event

Furthermore the tile needs to be enhanced in regards to the date to
* show long date type for the event 
* show both start and end time of event (as many events go over several days now)

Screenshot from 2020-12-06 01-03-57

MeghalBisht:
Fixes #5926 

1. the info "Online Event" if the event is not location based and
2. "Online Event and at [address here]" if it is a mixed event
3. show long date type for the event
4. show both start and end time of event 

**Screenshots -**

#### Checklist

### DIFF
--- a/app/components/event-card.js
+++ b/app/components/event-card.js
@@ -3,6 +3,7 @@ import { classNames } from '@ember-decorators/component';
 import { action, computed } from '@ember/object';
 import Component from '@ember/component';
 import { forOwn } from 'lodash-es';
+import moment from 'moment';
 import { pascalCase } from 'open-event-frontend/utils/string';
 
 @classic
@@ -40,7 +41,8 @@ export default class EventCard extends Component {
     this.set('eventType', eventType === this.eventType ? null : eventType);
   }
 
-  isSingleDay() {
-    return this.startsAt.isSame(this.endsAt, 'day');
+  @computed('event.startsAt', 'event.endsAt')
+  get isSingleDay() {
+    return moment(this.event.startsAt).isSame(this.event.endsAt, 'day');
   }
 }
--- a/app/components/event-card.js
+++ b/app/components/event-card.js
@@ -39,4 +39,8 @@ export default class EventCard extends Component {
   selectEventType(eventType) {
     this.set('eventType', eventType === this.eventType ? null : eventType);
   }
+
+  isSingleDay() {
+    return this.startsAt.isSame(this.endsAt, 'day');
+  }
 }
--- a/app/styles/libs/_helpers.scss
+++ b/app/styles/libs/_helpers.scss
@@ -98,3 +98,7 @@ $spacer-heights: 50 100 200 300 400 500 600 700 800 900;
 .no-wrap {
   white-space: nowrap;
 }
+
+.card-height {
+  min-height: 165px;
+}
--- a/app/templates/components/event-card.hbs
+++ b/app/templates/components/event-card.hbs
@@ -17,11 +17,11 @@
           @src={{if this.event.thumbnailImageUrl this.event.thumbnailImageUrl this.event.originalImageUrl}} />
       </a>
     {{/unless}}
-    <a style="min-height:165px" class="main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
+    <a class="{{if this.device.isMobile '' 'card-height'}} main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
       <SmartOverflow @class="header">
         {{this.event.name}}
       </SmartOverflow>
-      <div class="meta">
+      <div class="meta" style="margin-bottom: 0.5em">
         <span class="date">
           {{#if this.isSingleDay}}
             {{general-date this.event.startsAt 'date-time-long' tz=this.event.timezone}}
@@ -41,12 +41,12 @@
         </p>
       {{/if}}
       {{#if this.event.shortLocationName}}
-        <p style="color: #000000AD">
-          <i class="map marker alternate icon"></i>
+          <span style="color: #000000AD">
+            <i class="map marker alternate icon"></i>
+          </span>
           <SmartOverflow @class="d-inline description">
             {{this.event.shortLocationName}}
           </SmartOverflow>
-        </p>
       {{/if}}
     </a>
     <div class="extra content small text">
--- a/app/templates/components/event-card.hbs
+++ b/app/templates/components/event-card.hbs
@@ -23,10 +23,18 @@
       </SmartOverflow>
       <div class="meta">
         <span class="date">
-          {{general-date this.event.startsAt tz=this.event.timezone}}
+          {{general-date this.event.startsAt 'date-time-long'}}
+          -
+          {{general-date this.event.endsAt 'date-time-long'}}
         </span>
       </div>
       <SmartOverflow @class="description">
+        {{#if this.event.online}}
+          Online Event
+        {{/if}}
+        {{#if (and this.event.online this.event.shortLocationName)}}
+          and at
+        {{/if}}
         {{this.event.shortLocationName}}
       </SmartOverflow>
     </a>
--- a/app/templates/components/event-card.hbs
+++ b/app/templates/components/event-card.hbs
@@ -21,7 +21,7 @@
       <SmartOverflow @class="header">
         {{this.event.name}}
       </SmartOverflow>
-      <div class="meta" style="margin-bottom: 0.5em">
+      <div class="meta mb-2">
         <span class="date">
           {{#if this.isSingleDay}}
             {{general-date this.event.startsAt 'date-time-long' tz=this.event.timezone}}
--- a/app/templates/components/event-card.hbs
+++ b/app/templates/components/event-card.hbs
@@ -17,26 +17,37 @@
           @src={{if this.event.thumbnailImageUrl this.event.thumbnailImageUrl this.event.originalImageUrl}} />
       </a>
     {{/unless}}
-    <a class="main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
+    <a style="min-height:165px" class="main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
       <SmartOverflow @class="header">
         {{this.event.name}}
       </SmartOverflow>
       <div class="meta">
         <span class="date">
-          {{general-date this.event.startsAt 'date-time-long'}}
-          -
-          {{general-date this.event.endsAt 'date-time-long'}}
+          {{#if this.isSingleDay}}
+            {{general-date this.event.startsAt 'date-time-long' tz=this.event.timezone}}
+            - 
+            {{general-date this.event.endsAt 'time-tz-short' tz=this.event.timezone}}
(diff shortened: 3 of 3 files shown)


app/templates/components/event-card.hbs
```diff
@@ -23,10 +23,18 @@
...
+          {{general-date this.event.startsAt 'date-time-long'}}
+          -
+          {{general-date this.event.endsAt 'date-time-long'}}
         </span>
       </div>
       <SmartOverflow @class="description">
+        {{#if this.event.online}}
+          Online Event
+        {{/if}}
+        {{#if (and this.event.online this.event.shortLocationName)}}
```
daretobedifferent18, line 35: use this if statement inside upper if statement and then no need to add ```and``` just use like 
```handlebars
{{#if this.event.shortLocationName}}
       ...
       ...
{{/if}}
```

MeghalBisht: if the event is not online then location will also not be displayed.

MeghalBisht: thanks 😄

daretobedifferent18 (approved):
For me it's working fine

app/templates/components/event-card.hbs
```diff
@@ -23,10 +23,18 @@
       </SmartOverflow>
       <div class="meta">
         <span class="date">
-          {{general-date this.event.startsAt tz=this.event.timezone}}
```
iamareebjamal, line 26: Why do you think we had `tz=this.event.timezone`. We had not written it for fun, there was a reason for it. Why did you remove it?

app/templates/components/event-card.hbs
```diff
@@ -23,10 +23,18 @@
...
+          {{general-date this.event.startsAt 'date-time-long'}}
+          -
+          {{general-date this.event.endsAt 'date-time-long'}}
         </span>
       </div>
       <SmartOverflow @class="description">
+        {{#if this.event.online}}
+          Online Event
```
iamareebjamal, line 33: Wrap in t

app/templates/components/event-card.hbs
```diff
@@ -23,10 +23,18 @@
...
+          {{#if this.event.shortLocationName}}
+            and at
```
iamareebjamal, line 35: wrap in t

app/templates/components/event-card.hbs
```diff
@@ -23,10 +23,18 @@
       </SmartOverflow>
       <div class="meta">
         <span class="date">
-          {{general-date this.event.startsAt tz=this.event.timezone}}
+          {{general-date this.event.startsAt 'date-time-long'}}
```
daretobedifferent18, line 26: also change date-time-long -> date-time-tz-long

daretobedifferent18: @iamareebjamal shouldn't we do this also?

iamareebjamal: yes

iamareebjamal:
@mariobehling I think it'll look better this way and separate information as well

image
image
image

mariobehling:
@iamareebjamal Nice! So, if you want us to choose a version, I would choose version 2. Better do not introduce another green color.

@MeghalBisht Could you please change the dates in a way: When the beginning and end of the event is on the same day just keep the date first and e.g. 2:00 PM - 5:00 PM (CET).

Thank you!

MeghalBisht:
**After update -**

app/templates/components/event-card.hbs
```diff
@@ -17,18 +17,37 @@
...
     {{/unless}}
-    <a class="main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
+    <a style="min-height:165px" class="main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
       <SmartOverflow @class="header">
         {{this.event.name}}
       </SmartOverflow>
       <div class="meta">
         <span class="date">
-          {{general-date this.event.startsAt tz=this.event.timezone}}
+          {{#if (eq (general-date this.event.startsAt 'date-short') (general-date this.event.endsAt 'date-short')) }}
```
daretobedifferent18, line 26: No, don't do it like this

daretobedifferent18: Instead check for whether it ```isSingleDay``` in event-card.js file and then use it here

app/templates/components/event-card.hbs
```diff
@@ -17,18 +17,37 @@
...
+            {{general-date this.event.startsAt 'date-time-long'}}
+            - 
+            {{general-date this.event.endsAt 'time-tz-short' tz=this.event.timezone}}
+          {{else}}
+            {{general-date this.event.startsAt 'date-time-tz-long' tz=this.event.timezone}}
```
daretobedifferent18, line 31: Also here show timezone only with endAt date

app/templates/components/event-card.hbs
```diff
@@ -17,18 +17,37 @@
...
-      <SmartOverflow @class="description">
-        {{this.event.shortLocationName}}
-      </SmartOverflow>
+      {{#if this.event.online}}
+        <p style="color: #000000AD">
+          <i class="wifi icon"></i>
+          {{t 'Online Event'}}
+        </p>
+      {{/if}}
+      <p style="color: #000000AD">
```
daretobedifferent18, line 43: put this entire thing inside 
```handlebars
{{#if}}
     ....
{{/if}}
```

app/templates/components/event-card.hbs
MeghalBisht: okay, just wanted to know is this wrong?

app/templates/components/event-card.hbs
```diff
@@ -23,12 +23,12 @@
...
       <div class="meta">
         <span class="date">
-          {{#if (eq (general-date this.event.startsAt 'date-short') (general-date this.event.endsAt 'date-short')) }}
+          {{#if this.isSingleDay}}
             {{general-date this.event.startsAt 'date-time-long'}}
             - 
             {{general-date this.event.endsAt 'time-tz-short' tz=this.event.timezone}}
           {{else}}
-            {{general-date this.event.startsAt 'date-time-tz-long' tz=this.event.timezone}}
+            {{general-date this.event.startsAt 'date-time-long'}}
```
daretobedifferent18, line 31: why removed tz=this.event.timezone

app/templates/components/event-card.hbs
```diff
@@ -23,12 +23,12 @@
       </SmartOverflow>
       <div class="meta">
         <span class="date">
-          {{#if (eq (general-date this.event.startsAt 'date-short') (general-date this.event.endsAt 'date-short')) }}
+          {{#if this.isSingleDay}}
             {{general-date this.event.startsAt 'date-time-long'}}
```
daretobedifferent18, line 27: pls understand passing tz is not to show tz, it is used in general-date to get time in particular tz

daretobedifferent18 (commented):
Screenshot from 2020-12-08 02-47-30
still showing tz on both dates

app/templates/components/event-card.hbs
MeghalBisht: Thanks for clarifying.

iamareebjamal:
Screenshot_20201208-035834~2
The card is elongated even though there's a lot of place below the address. Please ensure that there is no extra space below the address in case of online+address events so that the card occupies least height as possible

MeghalBisht:
> The card is elongated even though there's a lot of place below the address. Please ensure that there is no extra space below the address in case of online+address events so that the card occupies least height as possible

Please clarify, all cards need to be of the same height?

iamareebjamal:
Of course!

MeghalBisht:
> Of course!
So for that max-height should be hard coded(for mobile & desktop)? As currently height is variable depending on the contents inside the card

iamareebjamal:
Currently they are all of same height, this should remain same after changes. To achieve this, remove the empty space below address

MeghalBisht:
> Currently they are all of same height, this should remain same after changes. To achieve this, remove the empty space below address

This is not equal. Pls check

iamareebjamal:
After reducing the size, it is not equal, that is another issue but at least on max size it is equal and thus should remain after this PR
image

iamareebjamal:
image

Add margin top of 0.5em on online event and this event is not a single day event and yet it shows up as a single day event

MeghalBisht:
> Add margin top of 0.5em on online event and this event is not a single day event and yet it shows up as a single day event

 isSingleDay() {
    return this.startsAt.isSame(this.endsAt, 'day');
  }
This is used to check for same day, I cannot understand why the above case is happening. Am i missing something?

app/components/event-card.js
```diff
@@ -39,4 +39,8 @@ export default class EventCard extends Component {
   selectEventType(eventType) {
     this.set('eventType', eventType === this.eventType ? null : eventType);
   }
+
+  isSingleDay() {
+    return this.startsAt.isSame(this.endsAt, 'day');
```
daretobedifferent18, line 44: here it should be ```this.event.startsAt``` and ```this.event.endsAt```

MeghalBisht: same result

iamareebjamal: Maybe it is checking `day` as in Friday instead of date

app/templates/components/event-card.hbs
```diff
@@ -17,11 +17,11 @@
           @src={{if this.event.thumbnailImageUrl this.event.thumbnailImageUrl this.event.originalImageUrl}} />
       </a>
     {{/unless}}
-    <a style="min-height:165px" class="main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
+    <a class="{{if this.device.isMobile '' 'card-height'}} main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
       <SmartOverflow @class="header">
         {{this.event.name}}
       </SmartOverflow>
-      <div class="meta">
+      <div class="meta" style="margin-bottom: 0.5em">
```
iamareebjamal, line 24: Use mb- class

maze-runnar (approved):

app/styles/libs/_helpers.scss
```diff
@@ -98,3 +98,7 @@ $spacer-heights: 50 100 200 300 400 500 600 700 800 900;
 .no-wrap {
   white-space: nowrap;
 }
+
+.card-height {
+  min-height: 165px;
```
iamareebjamal, line 103: Is this needed?

MeghalBisht: Yes, else in dekstop view the cards will not be of equal size ( online + location events have greater height). I have removed this in mobile view so extra height is not added there

MeghalBisht:
@iamareebjamal  I have used isSame function in this PR from momentjs, just saw this warning message in their docs, should this be changed?
https://momentjscom.readthedocs.io/en/latest/moment/05-query/02-is-same/

iamareebjamal:
No

Issue #5051, opened 2020-09-11
Sessions: Do not allow "Withdraw" session after session was held or when session is locked


mariobehling:
Sessions can be withdrawn after session was held or when session is locked.

Compare https://eventyay.com/e/8fa7fd14/session/6262
![Screenshot from 2020-09-11 15-35-44](https://user-images.githubusercontent.com/1583873/92932039-a320a300-f444-11ea-86ea-1b960ed6b347.png)

iamareebjamal:
We discussed that we were still to allow edits after the session was held and to only prevent editing if session was locked?

anjali1102:
Claiming this issue

PR #5959, opened 2020-12-08
fix: Do not allow "Withdraw" session after session was held

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/5051

#### Checklist


mariobehling:
Does this also need a change in the backend?

maze-runnar:
> Does this also need a change in the backend?

i think, no.

iamareebjamal:
It does. This will only not show the button, but the API will allow it

maze-runnar:
> It does. This will only not show the button, but the API will allow it

ok, making changes on server side too.

Issue #5475, opened 2020-11-03
Integrate Project with Big Blue Button


mariobehling:
Offering integrated video features is a must for an event platform these days. There are two good options Jitsi and Big Blue Button. BBB has more moderation features which are important for conference events.

Please propose and implement ways to integrate BBB and open relevant sub issues on frontend and server project. Specifically we need the following:
* Integration with eventyay rooms = BBB rooms
* Integration with eventyay sessions = BBB session rooms (if chosen by organizer)
* Access Control/Permissions integration with event team, e.g. allow organizers to control moderation features
* Access Control/Permissions integration with ticket holders to join a BBB event
* Access to recorded sessions by BBB through eventyay UI 

There is a demo server on https://demo.bigbluebutton.org/
API documentation: https://docs.bigbluebutton.org/dev/api.html

anjali1102:
Is this issue beginner friendly?

iamareebjamal:
No

PR #5990, opened 2020-12-10
feat: Add video channel support

iamareebjamal:
For #5475


mariobehling:
Does not work for me yet. Also editing existing rooms results in error page, e.g. https://open-event-frontend-git-video-channel.eventyay.vercel.app/events/a315169f/videoroom/all

iamareebjamal:
BBB support is not added yet. The PR is WIP

iamareebjamal:
Server changes have not been deployed, hence the error

mariobehling:
Is it possible to keep the drop down button blue?

Currently the error message pops up also when the user clicks BBB or Jitsi.

Screenshot from 2020-12-16 12-25-39

Are other pop up messages in the system bold as well? If not please follow same style as others.

mariobehling:
Maybe the question should be something like "Selecting another video integration will reset the data in the form. Do you want to proceed?".

mariobehling:
Screenshot from 2020-12-16 12-31-30

In the overview if a room is using Big Blue Button the source URL field should show [auto-generated per user]

mariobehling:
* The width of the columns was changed in a PR just a short while ago, which made it more easily readable, but the width seems to be changed back. 
* Also there was a box for the URL which made the visible part of the URL shorter. A user does not always need to see so much of the URL, but in case they want we have the URL in the box.

Screenshot from 2020-12-16 12-43-02

mariobehling:
> Is it possible to keep the drop down button blue?

What I mean is the dropdown main button blue, but maybe the dropdown expanded items white? What is the best option? Hard to say without seeing it.

mariobehling:
Scrolling over Copy button on top table results in internal scroll bar.

Screenshot from 2020-12-16 12-51-40

mariobehling:
Is there already an option in the admin area for the BBB integration?

iamareebjamal:
> Are other pop up messages in the system bold as well? If not please follow same style as others.

Yes, all confirm dialogs are bold

> In the overview if a room is using Big Blue Button the source URL field should show [auto-generated per user]

That is not feasible as join URL is needed to be generated with an API call everytime so it'll make the loading of the page every expensive on server. I'll hide this URL for BBB as there is no difference in the generated BBB URL and the URL redirected to in the internal link

> URL shortening

Ironically, the box implementation takes much more space and introduces unnecessary line breaks in first column even when there is a lot of space. Most of the space is gone in the text `Copy` and `Open` which are redundant in each row and the link can be opened already on click, so no need of a button for it. 
You can check the results here:

Before
image

After
image

Also, we are still shortening the URL if it requires more space, this is not apparent because we have much more space now than before because of removing the text from buttons and the Open Button. 

image

The column width changed here is of the first column because it had forced line break without any reason due to the URL column being so long. I'll reduce the URL columns even more if need be, as we still truncate the URL if it exceeds the size

> Scrolling over Copy button on top table results in internal scroll bar.

I'll remove the tooltip which appears on hover

> What I mean is the dropdown main button blue

Will make it blue

image

> Is there already an option in the admin area for the BBB integration?

Not right now but it can be checked from here

https://api.eventyay.com/admin/videochannel/

iamareebjamal:
@mariobehling I have reduced the URL column width, please check

mariobehling:
Ok, thank you. We can iterate from here. I wonder what are your ideas to control the moderator rights for BBB.

iamareebjamal:
Current Implementation:

Speaker/Organizer/Admin - Moderator
Others - Attendee

mariobehling:
> Current Implementation:
> 
> Speaker/Organizer/Admin - Moderator
> Others - Attendee

We need to communicate this somewhere. Well, some helper instructions are needed in a number of places. Let's test this for a few days.

It seems my reasoning for the columns was not good. So, I tried to explain this in a follow up issue here:
https://github.com/fossasia/open-event-frontend/issues/6056

We can also talk about this in the next meeting if more discussion is needed.

mariobehling:
"Selecting another video integration will reset the data in the form.Do you want to proceed?"

Needs a space after the full stop.

iamareebjamal:
Wow, how did I miss that? Probably I was trying to separate it with a newline and forgot when I reverted the change. Fixed

PR #6006, opened 2020-12-11
fix: Explore Page: Vertical event card height fixed

### LINKED ISSUE #6003: Vertical event card UI distorted
image

MeghalBisht:
Fixes #6003

Height changes happened in #5933 distorted the vertical event card. Fixed the height by conditionally changing it for only index page

Screenshot -

#### Checklist

### DIFF
--- a/app/templates/components/event-card.hbs
+++ b/app/templates/components/event-card.hbs
@@ -1,4 +1,4 @@
-<div class="{{if this.isWide 'event wide ui grid row'}}">
+<div class="{{if this.isWide 'event wide ui grid row'}}" style="height: 100%;">
   {{#if this.isWide}}
     {{#unless this.device.isMobile}}
       <div class="ui card rounded-r-none three wide computer six wide tablet column">
@@ -9,7 +9,7 @@
       </div>
     {{/unless}}
   {{/if}}
-  <div
+  <div style="height: 100%;"
     class="ui card {{this.eventClass}}">
     {{#unless this.isWide}}
       <a class="image" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
@@ -17,7 +17,7 @@
           @src={{if this.event.thumbnailImageUrl this.event.thumbnailImageUrl this.event.originalImageUrl}} />
       </a>
     {{/unless}}
-    <a class="{{if this.device.isMobile '' 'card-height'}} main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
+    <a class="main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
       <SmartOverflow @class="header">
         {{this.event.name}}
       </SmartOverflow>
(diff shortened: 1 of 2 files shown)


app/templates/explore.hbs
```diff
@@ -49,7 +49,7 @@
       {{/if}}
     </div>
     {{#each this.filteredEvents as |event|}}
-      <EventCard @event={{event}} @isWide={{true}} @filterByTags={{true}} @category={{this.category}} @subCategory={{this.sub_category}} @eventType={{this.event_type}} @shareEvent={{action "shareEvent"}} />
+      <EventCard @minHeight={{false}} @event={{event}} @isWide={{true}} @filterByTags={{true}} @category={{this.category}} @subCategory={{this.sub_category}} @eventType={{this.event_type}} @shareEvent={{action "shareEvent"}} />
```
daretobedifferent18, line 52: change this to ```fullHeight={{true}}```

app/templates/index.hbs
```diff
@@ -12,18 +12,18 @@
     <h2 class="main-heading">{{t 'Featured Events'}}</h2>
     <div class="ui stackable three column grid">
       {{#each this.featuredEvents as |event|}}
-        <EventCard @event={{event}} @shareEvent={{action "shareEvent"}} />
+        <EventCard @minHeight={{true}} @event={{event}} @shareEvent={{action "shareEvent"}} />
```
daretobedifferent18, line 15: remove from everyplace

app/templates/components/event-card.hbs
```diff
@@ -17,7 +17,7 @@
           @src={{if this.event.thumbnailImageUrl this.event.thumbnailImageUrl this.event.originalImageUrl}} />
       </a>
     {{/unless}}
-    <a class="{{if this.device.isMobile '' 'card-height'}} main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
+    <a class="{{if (or (not this.minHeight) this.device.isMobile) '' 'card-height'}} main content" href="{{href-to (if this.isCFS 'public.cfs' 'public') this.event.identifier}}">
```
daretobedifferent18, line 20: use instead ```{{unless (or this.fullHeight this.device.isMobile) 'card-height'}}```

MeghalBisht: I was doing this but there was a lint error 
"Using {{unless}} in combination with other helpers should be avoided."

daretobedifferent18: so use ```{{if (not (or this.fullHeight this.device.isMobile)) 'card-height'}}```

daretobedifferent18 (dismissed):

iamareebjamal:
Not a robust solution. Think of a better one please. This requires manual configuration of full height cards which is error prone

MeghalBisht:
> Not a robust solution. Think of a better one please. This requires manual configuration of full height cards which is error prone

Both index and explore pages are using the same event card component, so height can only be changed conditionally. Please describe what could be a better way?

iamareebjamal:
Before min-height was not needed. Try something which prevents its usage. Maybe grid layout

MeghalBisht:
> Before min-height was not needed. Try something which prevents its usage. Maybe grid layout

Grid is already used, now cards are of same height for all the screen sizes. Please check

iamareebjamal:
Excellent, thanks

PR #6069, opened 2020-12-18
wip: Event Wizard video integration

iamareebjamal:


iamareebjamal:
@maze-runnar Please take up this as I have to move to another issue. Ask in the meeting if you are confused about something

maze-runnar:
> @maze-runnar Please take up this as I have to move to another issue. Ask in the meeting if you are confused about something

ok

maze-runnar:
I have tried this, and discussed it. It has much more issues than it seems. If any one finds a way continue in another PR.

Issue #6086, opened 2020-12-19
Captcha: Substitute Gcaptcha with Hcaptcha and implement it for sign ups, repetitive login attempts and password reset


mariobehling:
In order to secure the system and block hacking attempts please implement hcaptcha:
- [x] for sign ups
- [ ] if a user gets the login wrong more than 2 times
- [x] when the user requests a password reset

Please also ensure the captcha works smoothly if users sign up during a ticket order process.

Show the info about the captcha configuration in the respective admin area.

maze-runnar:
on it

PR #6087, opened 2020-12-19
feat: add hcaptcha to register and forgot password

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/6086

#### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,5 @@
 GOOGLE_API_KEY="Sample Key"
+HCAPTCHA_SITE_KEY=""
 API_HOST=https://open-event-api-dev.herokuapp.com
 MAP_DISPLAY=embed
 FASTBOOT_DISABLED=true
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
 GOOGLE_API_KEY="Sample Key"
-HCAPTCHA_SITE_KEY=""
 API_HOST=https://open-event-api-dev.herokuapp.com
 MAP_DISPLAY=embed
 FASTBOOT_DISABLED=true
+#HCAPTCHA_SITE_KEY=""
--- a/app/components/forms/register-form.js
+++ b/app/components/forms/register-form.js
@@ -56,7 +56,7 @@ export default Component.extend(FormMixin, {
     }
   },
 
-  showHcaptcha: computed(function(){
+  showHcaptcha: computed(function() {
     return ENV.hcaptchaKey === null || ENV.hcaptchaKey === undefined ? true : true;
   }),
 
--- a/app/components/forms/register-form.js
+++ b/app/components/forms/register-form.js
@@ -57,7 +57,7 @@ export default Component.extend(FormMixin, {
   },
 
   showHcaptcha: computed(function() {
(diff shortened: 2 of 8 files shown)


config/environment.js
```diff
@@ -79,6 +79,12 @@ module.exports = function(environment) {
     authorizer: 'authorizer:jwt'
   };
 
+  ENV['ember-h-captcha'] = {
+    jsUrl   : 'https://hcaptcha.com/1/api.js', // default
+    sitekey : '4d94db5b-d3fd-407b-a574-e6ba0af9b73c',
```
iamareebjamal, line 84: Hard-coded key

iamareebjamal (changes requested):

iamareebjamal:
If key is not present, captcha shouldn't be enabled.

maze-runnar:
so key would be added by .env or from admin section?

iamareebjamal:
.env

mariobehling:
We have an area for a Google captcha in the admin section. Ultimately it would be good to have it here.

Screenshot from 2020-12-19 20-43-54

iamareebjamal:
@mariobehling GSoC students just added it here for show without implementing it. We need the key at build time for the captcha service, just like the Google Maps API key, thus it is not possible to fetch it at runtime

.env.example
```diff
@@ -1,4 +1,5 @@
 GOOGLE_API_KEY="Sample Key"
+HCAPTCHA_SITE_KEY=""
```
iamareebjamal, line 2: Move it to last and comment out

iamareebjamal:
Captcha shouldn't be enabled if key isn't present, as reviewed previously

app/templates/components/forms/register-form.hbs
```diff
@@ -37,8 +37,9 @@
           </div>
         </div>
       </div>
+      <HCaptcha @onSuccess={{action 'captchaSuccess'}} @onExpired={{action 'onCaptchaExpired'}}/>
```
iamareebjamal, line 40: Simply use mut action instead of defining 2 actions in js

maze-runnar:
> Captcha shouldn't be enabled if key isn't present, as reviewed previously

ok

app/components/forms/register-form.js
```diff
@@ -53,6 +56,10 @@ export default Component.extend(FormMixin, {
     }
   },
 
+  showHcaptcha: computed(function() {
+    return ENV.hcaptchaKey === null || ENV.hcaptchaKey === undefined ? false : true;
```
iamareebjamal, line 60:
```suggestion
    return !!ENV.hcaptchaKey;
```

iamareebjamal: This doesn't change so no need of computed

app/components/forms/register-form.js
```diff
@@ -6,6 +6,7 @@ export default Component.extend(FormMixin, {
   email     : '',
   password  : '',
   isLoading : false,
+  captcha   : false,
```
Jyotsna1006: can i handle the issue?

maze-runnar: what...?

Issue #5100, opened 2020-09-20
Session/Speaker: Distinguish Private and Public Info Pages/Views and Separate Them


mariobehling:
In previous iterations we focused on the views of the public information of sessions and speakers, e.g. shown on the schedule and public session page. However, the speaker can enter much more information than should be publicly available in their call for speakers response. 

Right now we do not show to the speaker which information is public or privately shown to the review team. We also do not have a page to show the session and speaker information which includes the private info. It is only possible to see the private info
if the speaker or organizer/reviewer edits sessions/speakers info.

Therefore we need to change and tweak a number of things in different areas. (Please open sub-issues where necessary)
1. In the wizard step 5 session and speaker forms please add a column "Show Public" to display a field publicly. If this is chosen show the field on the public session or speaker page. In a first step we could simply decide which info is public or private by default show the info "public" or "private" instead of a switch button.
2. Session View page and Session Edit page are confusing - we added the same functionality under different URLs, e.g. compare https://eventyay.com/e/598/cfs/session/5057/edit and https://eventyay.com/events/3689bc98/sessions/5057/edit. Whereas we only have one (public) page to view the session page here https://eventyay.com/e/3689bc98/session/5057 without showing the private form fields. What to do?
* We need to distinguish public and private pages and views. 
  * Public page view is on a page for attendees of an event. It only includes essential information and fields marked as "public" in the wizard.
  * Private page view is a page for speakers and organizers/reviewers. It includes all submitted information.
* Where to put private vs. public view page and which one can be edited?
  * Public pages are part of an event and can continue to stay on pages such as this one here https://eventyay.com/e/3689bc98/session/5057. These pages should not have an edit or withdraw button. They are just there to show information to attendees.
  * Private pages as in the above case the private submission page should be here as in this example https://eventyay.com/events/3689bc98/sessions/5057/ They should include an edit and withdraw button 
  * In the above example the editable version would be available here https://eventyay.com/events/3689bc98/sessions/5057/edit
* When a user submits a session to the call for speakers, session page and speaker page should be created according to the event. They should not follow the cfs slug. Speaker and organizers can both edit the session. Once marked as "Locked" only the organizer can edit it.

Compare:
* Private session page - Access by speaker and organizers
/events/[event-number]/sessions/[session-number]/
/events/[event-number]/sessions/[session-number]/edit
* Public - Public Access
eventyay.com/e/[event-number]/session/[session-number]
(no edit page)

[...]

PR #6094, opened 2020-12-20
feat: introduce is_public in session-speaker wizard

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/5100
server : https://github.com/fossasia/open-event-server/pull/7524

#### Checklist

### DIFF
--- a/app/mixins/custom-form.js
+++ b/app/mixins/custom-form.js
@@ -47,7 +47,7 @@ export default Mixin.create(MutableArray, {
         form            : 'session',
         type            : 'text',
         isRequired      : false,
-        isPublic        : false,
+        isPublic        : true,
         isIncluded      : true,
         event           : parent
       }),
@@ -149,7 +149,7 @@ export default Mixin.create(MutableArray, {
         form            : 'speaker',
         type            : 'text',
         isRequired      : false,
-        isPublic        : false,
+        isPublic        : true,
         isIncluded      : true,
         event           : parent
       }),
--- a/app/mixins/custom-form.js
+++ b/app/mixins/custom-form.js
@@ -10,6 +10,7 @@ export default Mixin.create(MutableArray, {
         form            : 'session',
         type            : 'text',
         isRequired      : true,
+        isPublic        :true,
         isIncluded      : true,
         isFixed         : true,
         event           : parent
@@ -19,6 +20,7 @@ export default Mixin.create(MutableArray, {
         form            : 'session',
         type            : 'text',
         isRequired      : false,
+        isPublic        :false,
         isIncluded      : false,
         event           : parent
       }),
@@ -27,6 +29,7 @@ export default Mixin.create(MutableArray, {
         form            : 'session',
         type            : 'text',
         isRequired      : false,
+        isPublic        :true,
         isIncluded      : true,
         event           : parent
       }),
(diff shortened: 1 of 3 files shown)


iamareebjamal:
Before this, we need to create a migration to make previous fields as public. And decide which ones should be public by default

maze-runnar:
> we need to create a migration to make previous fields as public.

is it would be done like - like op.execute("update custom_forms set is_public = true where condition)

iamareebjamal:
Yes

maze-runnar:
how to add a migration? Means I have already added isPublic in schema previously so how to migrate now, to make previous fields as public.

iamareebjamal:
You wrote the answer yourself in the comment above

`op.execute("update custom_forms set is_public = true where condition")`

Find the field identifiers which should be public and then add them in the condition clause

maze-runnar:
I am asking for where to write op.execute("update custom_forms set is_public = true where condition"). Because I am not making any changes in schema at the time. So no migration file will be generated. 😕

iamareebjamal:
Whenever you need help with the command, you run it without arguments or with `--help`

```
python manage.py db
```

```
usage: Perform database migrations

Perform database migrations

positional arguments:
  {init,revision,migrate,edit,merge,upgrade,downgrade,show,history,heads,branches,current,stamp}
    init                Creates a new migration repository
    revision            Create a new revision file.
    migrate             Alias for 'revision --autogenerate'
    edit                Edit current revision.
    merge               Merge two revisions together. Creates a new migration file
    upgrade             Upgrade to a later version
    downgrade           Revert to a previous version
    show                Show the revision denoted by the given symbol.
    history             List changeset scripts in chronological order.
    heads               Show current available heads in the script directory
    branches            Show current branch points
    current             Display the current revision for each database.
    stamp               'stamp' the revision table with the given revision; don't run any migrations

optional arguments:
  -?, --help            show this help message and exit
```

Which shows it can be done using `revision` command

app/mixins/custom-form.js
```diff
@@ -19,6 +20,7 @@ export default Mixin.create(MutableArray, {
         form            : 'session',
         type            : 'text',
         isRequired      : false,
+        isPublic        : false,
```
iamareebjamal, line 23: Should be public

app/mixins/custom-form.js
```diff
@@ -35,6 +38,7 @@ export default Mixin.create(MutableArray, {
...
```
iamareebjamal, line 41: Should be public

iamareebjamal:
@maze-runnar Please complete this PR as instructed in the server PR. All the fields I have marked public in that PR should be public here

maze-runnar:
ok

app/mixins/custom-form.js
```diff
@@ -43,6 +47,7 @@ export default Mixin.create(MutableArray, {
...
```
iamareebjamal, line 50: This is also public IIRC

app/mixins/custom-form.js
```diff
@@ -134,6 +149,7 @@ export default Mixin.create(MutableArray, {
         form            : 'speaker',
         type            : 'text',
         isRequired      : false,
+        isPublic        : false,
```
iamareebjamal, line 152: This was public

mariobehling:
This is only implementing the public vs private information. However we still do not have a private or public page for the session or speaker info where we can clearly see the information. I am trying to make the issue clearer. As we come back to this over and over again, please help to resolve this until the end. So, we do not loose the train of thought and can resolve this ultimately: https://github.com/fossasia/open-event-frontend/issues/6222

PR #6116, opened 2020-12-23
feat: Add side panel to video room

### LINKED ISSUE #5842: Virtual Events: Provide sidebar menu that slides in with links to different video rooms
How could we enable participants to switch rooms directly at a virtual event? Other services offer direct link in a sidebar to other rooms or video channel overviews. We can do something similar. 

The idea is to implement a sidebar on the left of the screen. On the sidebar please display a list of rooms of the event. If an event runs over several days possibly show the rooms that are active on that day, but this kind of detail could also be implemented later.

Please implement the feature as follows:
* Add a top square button with a left < arrow or similar icon on the video screen
* Clicking the icon should result in a sidebar that moves in
* The sidebar should have the following content:
   * "eventyay" logo or text - link to eventyay.com
   * [Event Name] - link to event start page on eventyay
   * [Event Dates and Time]
   * [Schedule] - link
   * [Event Rooms] - list of rooms as buttons

To be implemented as possible or later
* Functionality to show or filter tracks
* Functionality to show or filter current sessions

Something like the following would be nice - with a right sidebar.
Screenshot from 2020-12-12 14-07-46

Gmail has a nice sidebar icon that would seem suitable
Screenshot from 2020-12-08 15-05-35

Compare sidebar menu of hopin:
sessions

Example of a sidebar that moves in
gallery-5852e16981493

iamareebjamal:
Fixes #5842 

image


mariobehling (approved):
Great! Let's iterate from here.

PR #6166, opened 2020-12-29
feat: Add stream side panel to public event page

### LINKED ISSUE #6135: Virtual Event: Streamline Participation and In-App Experience in Online Event
What is the easiest way to create a full online event experience with schedule, speaker and event information pages? We already have these pages, but they are not included in the "video app experience" yet. We probably do not want to start replicating the pages. So, what is the easiest way to include these pages without getting into loops and confusing users?

Here is an approach. For the online event experience with the iframe and videos, we can include the event pages such as the speaker page and make it accessible from the UI, but we need to limit the access to the eventyay area to make the experience consistent. 

So, basically when the user is within the iframe, the user is participating in the event. When the user is outside of the iframe he/she is just viewing the event (but not participating!). 

Therefore we need to make the following changes:
1. When the user clicks on a session "Join Video" change the implementation to open the link in the same tab and show a loading icon and the message _"Joining Event"_
2. After the user joined the event the user has access to the event information through the sidebar and can access different pages. The user is now participating in the event itself and accesses all info through the "app UI". This should work as follows:
  * Open all external links in a new tab (let's just implement this across eventyay by default)
  * Change the top bar as follows: When the user clicks the "eventyay" logo they leave the app and go to the frontpage
3. From the event name on the left side iframe sidebar direct to the info page of the event with the sidebar instead of opening a new tab.
4. When no schedule is available also do not show the button on the sidebar.

Screenshot from 2020-12-24 11-04-26

iamareebjamal:
Fixes #6135

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -31,3 +31,4 @@ rules:
     -
       allow-element-with-attribute: true
       allow-element-with-class: true
+  property-sort-order: 0
--- a/app/components/public/session-item.js
+++ b/app/components/public/session-item.js
@@ -56,6 +56,10 @@ export default class SessionItem extends Component {
   @action
   goToStream() {
     const url = this.router.urlFor('public.stream.view', this.args.event?.identifier ?? this.args.session.get('event.identifier'), this.args.session.get('microlocation.videoStream.slugName'), this.args.session.get('microlocation.videoStream.id'));
-    window.open(url, '_blank');
+    if (this.args.sameTab) {
+      location.href = url;
+    } else {
+      window.open(url, '_blank');
+    }
   }
 }
--- a/app/components/public/side-menu.js
+++ b/app/components/public/side-menu.js
@@ -3,18 +3,18 @@ import { action, computed } from '@ember/object';
 import Component from '@ember/component';
 import moment from 'moment';
 import { tagName } from '@ember-decorators/component';
-import { SPEAKERS_FILTER } from 'open-event-frontend/routes/public/speakers';
 import { tracked } from '@glimmer/tracking';
+import { hasSpeakers, hasSessions } from 'open-event-frontend/utils/event';
 
 @classic
 @tagName('')
 export default class SideMenu extends Component {
 
   @tracked
-  showSpeakers = false;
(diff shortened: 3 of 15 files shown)


mariobehling:
Thanks. The schedule button disappears when the user is on the schedule page itself.

Do you think we could also have "Speakers" button here? But only show buttons when there is a schedule and speakers defined. Could that be solved in this PR or a follow-up PR rather?

Screenshot from 2020-12-29 07-36-54

iamareebjamal:
I deliberately hid schedule button from here. Will remove the logic to hide it

Will add speaker button here as well soon

iamareebjamal:
However, I think there is no way an event can have a stream without schedule or sessions? Is it?

mariobehling:
> However, I think there is no way an event can have a stream without schedule or sessions? Is it?

Yes, there is - the event stream, that you are also implementing in wizard step 1.

mariobehling:
Screenshot from 2020-12-29 08-03-28

Sidepanel event link should now open in same tab.

mariobehling:
After last changes, Jitsi does not show up anymore.

Screenshot from 2020-12-29 08-14-16

The order of video rooms is different on the sidepanel of the schedule compared to the sidepanel when the user is on a video room page. Any chance to standardize this?

iamareebjamal:
Jitsi is showing up for me correctly. https://open-event-frontend-git-fork-iamareebjamal-side-pane.eventyay.now.sh/e/a315169f/video/codeheat-ask-me-anything-january-edition/65?side_panel=true

About order, the room you are in is always on top and then the event room. Then we can order the rooms by name so they will be in order

mariobehling:
As you adjusted the way it works I think the "Join Video" button can now open in the same window for our integrated Jitsi and BBB, but it should still open in an external window for other services, right?

iamareebjamal:
But it can't detect jitsi or BBB without loading an additional resource - video channel (which it loads on the dedicated video page). So, it is not that straightforward to implement. So, I think we follow up on it in another PR and explore what can be done otherwise it will result in code duplication or waited loading
_

PR #6178, opened 2020-12-30
fix: Change light gray font to dark gray

### LINKED ISSUE #5951: Fonts: Change light gray font to dark gray across app
There are a number of places with light gray color. This is often difficult to see, especially if it is not bold. However, there is no point of making more and more text bold. 

Please change light gray color to a dark gray that is easier to read.

sachinchauhan2889:
Fixes #5951 

__screenshots__
INVOICE
session card
ticket

#### Checklist

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/app/styles/partials/text.scss
+++ b/app/styles/partials/text.scss
@@ -11,7 +11,7 @@
 }
 
 .gray-text {
-  color: #555;
+  color: #555 !important;
 }
 
 .strike.text {
@@ -22,4 +22,8 @@
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
+}
+
+.weight-600 {
+  color: #555;
 }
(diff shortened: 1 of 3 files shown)


app/styles/partials/text.scss
```diff
@@ -23,3 +23,7 @@
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+
+.weight-600 {
```
iamareebjamal, line 27: That's not good, weight class should not dictate color of text

sachinchauhan2889: But sir, in divyam's PR you told to do this.

iamareebjamal: I told to change the color from whichever class adds it. I don't think weight class adds it

PR #6184, opened 2020-12-31
feat: Push and highlight active video stream

iamareebjamal:


mariobehling:
Very good. Could you make the active room a bit brighter, please? The difference is not that big that it is apparent right away.

Screenshot from 2020-12-31 08-54-42

iamareebjamal:
image

I can change the outline from white to blue for selected item

iamareebjamal:
Or should I make it brighter than the side panel?

image

mariobehling:
> Or should I make it brighter than the side panel?
> 
> image

Yes, I think this one would be good.

PR #6201, opened 2021-01-01
fix: Video Tab: Icons aligned when event or video room has a long name 

### LINKED ISSUE #6187: Video Tab: Icons disaligned when event or video room has a long name
In the video tab overview first column the icons are disaligned when an event or video room name has a longer name. Please ensure the icons stay aligned.

Screenshot from 2020-12-31 09-08-26

MeghalBisht:
Fixes #6187 

#### Checklist

### DIFF
--- a/app/styles/components/extras.scss
+++ b/app/styles/components/extras.scss
@@ -28,3 +28,9 @@ a.link-item {
     font-size: .9rem;
   }
 }
+
+@media screen and (max-width: 537px) {
+  .small-wrap {
+    flex-wrap: wrap !important;
+  }
+}
--- a/app/styles/components/extras.scss
+++ b/app/styles/components/extras.scss
@@ -28,9 +28,3 @@ a.link-item {
     font-size: .9rem;
   }
 }
-
-@media screen and (max-width: 537px) {
-  .small-wrap {
-    flex-wrap: wrap !important;
-  }
-}
--- a/app/styles/libs/_helpers.scss
+++ b/app/styles/libs/_helpers.scss
@@ -63,6 +63,7 @@ $spacer-heights: 50 100 200 300 400 500 600 700 800 900;
   display: flex;
 }
 
+.wrap,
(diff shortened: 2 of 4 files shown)


afzal442:
@MeghalBisht In mobile view I got something like this....
c2

MeghalBisht:
> @MeghalBisht In mobile view I got something like this....

Please provide a screenshot without these changes

mariobehling:
Ok, I was originally thinking about another solution. In any case the icons should be either next to each other or below to each other. It looks like you achieved this in this PR. We need to come up with a mobile view UI probably in another PR. So, looks good to me. Thanks.

iamareebjamal:
@MeghalBisht We don't need media query for this. Too much media queries complicate things. Just flexbox can solve the issue

MeghalBisht:
@iamareebjamal  But we don't want the wrap property to work for bigger screen resolution.
Also, mobile view starts at about 767px but at that resolution the original icon position works fine.

iamareebjamal:
It won't work on bigger screens, that's how flex works

MeghalBisht:
wrap for bigger screen :


no wrap for bigger screen :

MeghalBisht:
> It won't wrap on bigger screens, that's how flex works

I think it will wrap on bigger screens as once the text is long the icons have no other choice but to shift to the next line.

iamareebjamal:
There are other flex options like flex grow shrink. Try a few options

MeghalBisht:
Current commit changes work better than the last one.
Few relevant screenshots -

**NOW -**







**BEFORE -**  (some icons were horizontal, some were vertically aligned)

iamareebjamal:
You mean flex solution is better than media query or last commit is better than first flex solution?

MeghalBisht:
Both are flex wrap solutions only, in the previous commit wrap was done after a particular resolution, in the current commit wrap will happen for all resolutions whenever text is long

PR #6207, opened 2021-01-01
fix: scrollbar issues on the webpages

### LINKED ISSUE #6194: Mobile View: Avoid bottom scroll bar
In mobile view a numbers of pages have a bottom scroll bar. Please test this with an actual phone. It doesn't always show up on a browser.

Screenshot_20201231_210317_com android chrome

Creatoon:
Fixes #6194 

On some pages like ticket-buying page , a horizontal scrollbar on the bottom side of the page appears , now this resolves this issue.

WhatsApp Image 2021-01-01 at 10 18 18 PM

#### Checklist

### DIFF
--- a/app/styles/pages/events.scss
+++ b/app/styles/pages/events.scss
@@ -7,11 +7,6 @@
 }
 
 @media screen and (max-width: 450px) {
-
-  html {
-    overflow-x: hidden;
-  }
-
   .search-box {
     float: none;
     margin-top: 2px;
--- a/app/styles/pages/events.scss
+++ b/app/styles/pages/events.scss
@@ -7,6 +7,11 @@
 }
 
 @media screen and (max-width: 450px) {
+
+  html {
+    overflow-x: hidden;
+  }
+
   .search-box {
     float: none;
     margin-top: 2px;
--- a/app/styles/libs/_helpers.scss
+++ b/app/styles/libs/_helpers.scss
@@ -54,7 +54,7 @@ $spacer-heights: 50 100 200 300 400 500 600 700 800 900;
 }
 
 @media (max-width: 450px) {
-  .small-screen-margin {
+  .ui.relaxed.grid {
     margin-left: -1rem !important;
     margin-right: -1rem !important;
   }
--- a/app/styles/libs/_helpers.scss
+++ b/app/styles/libs/_helpers.scss
@@ -53,6 +53,13 @@ $spacer-heights: 50 100 200 300 400 500 600 700 800 900;
   }
 }
 
+@media (max-width:450px) {
+  .small-screen-margin {
+    margin-left: -1rem !important;
+    margin-right: -1rem !important;
+  }
+}
+
 [data-ember-action] {
   &:not(:disabled) {
     cursor: pointer;
--- a/app/styles/libs/_helpers.scss
+++ b/app/styles/libs/_helpers.scss
@@ -53,7 +53,7 @@ $spacer-heights: 50 100 200 300 400 500 600 700 800 900;
   }
 }
 
-@media (max-width:450px) {
+@media (max-width: 450px) {
   .small-screen-margin {
     margin-left: -1rem !important;
     margin-right: -1rem !important;
--- a/app/styles/libs/_helpers.scss
+++ b/app/styles/libs/_helpers.scss
@@ -53,13 +53,6 @@ $spacer-heights: 50 100 200 300 400 500 600 700 800 900;
   }
 }
 
-@media (max-width: 450px) {
-  .ui.relaxed.grid {
-    margin-left: -1rem !important;
-    margin-right: -1rem !important;
-  }
-}
-
 [data-ember-action] {
   &:not(:disabled) {
     cursor: pointer;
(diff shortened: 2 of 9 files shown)


MeghalBisht (commented):
This will affect other stylings as well, for example the search term now overflows on every page.

Creatoon:
@MeghalBisht can you share the screenshot of overflow , i am not able to reproduce the behaviour because from my side it is just looking fine.

app/styles/pages/events.scss
```diff
@@ -7,6 +7,11 @@
 }
 
 @media screen and (max-width: 450px) {
+
+  html {
+    overflow-x: hidden;
```
MeghalBisht, line 12: Setting overflow: hidden for the entire html is not a good way to fix this.

103444520-e66e7e00-4c8e-11eb-841f-67f87549cf02_LI

Creatoon: setting overflow-x: hidden to the "body" element fixes this problem, please test it once on your side , then i will commit the changes.

MeghalBisht: Adding overflow as hidden to html, body or * is not a good option. Please try to find the particular element which is overflowing and change it

Creatoon: ok mam I am working on this

MeghalBisht: Extra margin causes the overflow. 
2021-01-02 (2)_LI

Creatoon: After overwriting it to 0 , the margin is being created on the right side.

Screenshot from 2021-01-02 23-51-00


overwriting it to -1 would be the best case

Screenshot from 2021-01-02 23-50-48

MeghalBisht: why did you change margin for .ui.grid? only change for .ui.relaxed.grid

Creatoon: I have only changed margin for .ui.relaxed.grid , please look at the green lines, it shows changes happned only on the .ui.relaxed.grid

MeghalBisht: Okay then -1 is good

Creatoon: @MeghalBisht actually mam, after putiing the margin-right: -1rem , the ui started looking incosistent (see the margin on the left and right)

Screenshot from 2021-01-03 03-24-56

That's why i made both margin-right and margin-left to -1rem , and they looks perfect now

Screenshot from 2021-01-03 03-25-25

MeghalBisht: Works now, but this class is used in various other places also, so now everywhere margin will be altered hence adding the same as inline CSS would be better.

Creatoon: @MeghalBisht by inline css, do you mean i have to add styles directly in the tag.

Creatoon: Or should I add an id and override these properties.

MeghalBisht: you can add directly in the tag.

Creatoon: @MeghalBisht mam the problem is we can not add media queries in inline css, media query is required because on larger screens, the margin : -1 is not required.

Creatoon: and i found some more pages with this same problem , so i think i should make an separate css in helper.scss , so that i can add this class to every page where it is needed, should i proceed

MeghalBisht: As .ui.relaxed.grid was used only in these 5 files, so I think there is no need to add a new class. Instead please add the styling to this class(.ui.relaxed.grid) itself.

Creatoon: @MeghalBisht  ok mam , working

Creatoon:
@MeghalBisht please review mam.

MeghalBisht:
Unfortunately, the scrollbar still appears for larger screen sizes.

Creatoon:
@MeghalBisht this issue is coming on the touch screen devices , you can turn on the touch screen feature and please look at the nav bar from the right side on the top , you will see that the element below the nav bar ("pusher > footer-pusher") which is siblim element to the nav-bar itself.
you will find footer-pusher's content is overflowing from the container , and that's the issue , if will apply overflow-x: hidden to this element, then every problem will be solved .
right now we are changing the margins which can solve the issue below 767 px but above this width , we will face the same issue of overflow only on "TOUCH DEVICES"
so the crux is we should add overflow-x hidden to the footer-pusher element, please try this once.

Screenshot from 2021-01-03 22-06-14

Soumi150 (commented):
Please check if adding x-scrollable class fixes this .

Creatoon:
> Please check if adding x-scrollable class fixes this .

@Soumi150  no making overflox-x : auto , will only remove the scollbar , but not fix the scrolling nature on touch screens, we have to make it "hidden".

MeghalBisht:
>  overflow-x hidden to the footer-pusher element, please try this once.

Works for me so far, please check from your end that other element styles are not disturbed.

Creatoon:
> Works for me so far, please check from your end that other element styles are not disturbed.

yes mam checked the other pages, everything is working fine.

maze-runnar (approved):

Issue #3693, opened 2019-12-09
Pagination is broken in Message list of Admin panel


codedsun:
**Describe the bug**
The pagination is currently broken for the messages section in admin panel. The website says it's showing all the items but the API returns only 20 items and per single hit, if no filters(page[size], page[number]) are given.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Messages section in the admin panel

**Expected behavior**
It should work properly

**Screenshots**

![](https://user-images.githubusercontent.com/20026440/70394816-891de400-1a1e-11ea-8e75-4134d0faf1b8.png)

**Additional context**

kushthedude:
Please provide Screenshots while reporting the issues.

codedsun:
@kushthedude Ok, but you can reproduce the behavior by going to that section.

kushthedude:
I am Afk, thats why asked for a gif or image illustration.

codedsun:
@kushthedude have added when you commented to do so. Thanks

maze-runnar:
fixed https://github.com/fossasia/open-event-frontend/pull/6254

PR #6254, opened 2021-01-06
fix: pagination page size in message list

maze-runnar:
related https://github.com/fossasia/open-event-frontend/issues/3693

#### Checklist


app/routes/admin/messages/list.js
```diff
@@ -7,9 +7,9 @@ export default class extends Route.extend(EmberTableRouteMixin) {
     const searchField = 'recipient';
     const filterOptions = this.applySearchFilters([], params, searchField);
     let queryString = {
-      filter        : filterOptions,
-      'page[size]'  : params.per_page || 100,
-      'page[number' : params.page || 1
+      filter         : filterOptions,
+      'page[size]'   : params.per_page || 100,
+      'page[number]' : params.page || 1
```
iamareebjamal, line 12: What does this change?

maze-runnar: I was looking into https://github.com/fossasia/open-event-frontend/issues/3693 then I found the wrong syntax `page[number`

maze-runnar: the problem with #3693 is due to wrong passing of querystring. I have some doubts regarding that to discuss in technical meeting.

app/routes/admin/messages/list.js
```diff
@@ -7,9 +7,9 @@ export default class extends Route.extend(EmberTableRouteMixin) {
...
     };
     queryString = this.applySortFilters(queryString, params);
     return this.asArray(this.modelFor('admin.messages', queryString));
```
maze-runnar, line 15: Actually the queryString should be in 'admin.messeges' with page[size] and page[number]. But how to pass it there?

maze-runnar:
I am getting this error after moving mesages/list to mesages and deleting it.
Screenshot from 2021-01-08 09-24-13

maze-runnar:
the above is fixed, but sorting and searching is giving error.

maze-runnar:
`{"errors": [{"status": 400, "source": {"parameter": "filters"}, "title": "Invalid filters querystring parameter.", "detail": "MessageSettings has no attribute recipient"}], "jsonapi": {"version": "1.0"}}`

iamareebjamal:
Push the changes, or else how will I debug?

iamareebjamal:
See Message Setting model in server and remove sortable from all fields which are not in model, like recipient

maze-runnar:
I have removed recipient and now ony sent_at is sortable. And there is no other field that can be make searchable. So I hid the searchbox.

iamareebjamal:
Is the save button necessary?

maze-runnar:
> Is the save button necessary?

I think no, it's not working right now or it shouldn't be there.

iamareebjamal:
It should not be removed if it was saving the settings, please restore but put it in a proper place

maze-runnar:
> It should not be removed if it was saving the settings, please restore but put it in a proper place

but it's not saving settings actually. they are kind of fixed. On server side the setting is implemented as  

```
def create_admin_message_settings():
    default_mails = [
        "Next Event",
        "Session Schedule Change",
        "User email",
        "Invitation For Papers",
        "After Event",
        "Ticket(s) Purchased",
        "Session State Change",
        "Event Published",
        "Event Export Failed",
        "Event Exported",
        "Event Role Invitation",
        "New Session Proposal",
    ]
    for mail in MAILS:
        if mail in default_mails:
            get_or_create(
                MessageSettings,
                action=mail,
                mail_status=True,
                notification_status=True,
                user_control_status=True,
            )
        else:
            get_or_create(
                MessageSettings,
                action=mail,
                mail_status=False,
                notification_status=False,
                user_control_status=False,
            )
```

maze-runnar:
so on refreshing page the settings are set to previous(the above values), no matter they are saved whatever

iamareebjamal:
But notification and mail status are editable

maze-runnar:
reverted

iamareebjamal:
Move save to right and add padding top

maze-runnar:
is it ok?
Screenshot from 2021-01-08 17-05-59

PR #6275, opened 2021-01-08
fix: Ticket Sales: Sold out tickets are not bookable

### LINKED ISSUE #6269: Ticket Sales: Confusing Sold Out Message Instead of Showing "Sold Out"
There is a sold out message when people try to buy a ticket that is sold it. Expected is, that the ticket would not be "bookable" instead there should be a message "Sold Out" in the area drop-down of the ticket number.

For the sold out message itself, if it is displayed anywhere it should show "Ticket [ticket name] sold out." instead of ticket 1865 sold out.

Screenshot from 2021-01-07 10-16-42

MeghalBisht:
Fixes #6269

Sold out tickets can no longer be booked.

Screenshot -
Capturesw

#### Checklist


MeghalBisht:
Few more ways to style the Sold Out text. 
Captureww1
Captureww2
Capturewww

iamareebjamal:
If non-organizers are able to see order statistics, it is an issue, please open a bug for this in server

For showing sold out, there needs to be a new route which returns which ticket is available and which is sold out

MeghalBisht:
By order stats do you mean ticket stats?

app/templates/components/public/ticket-list.hbs
```diff
@@ -80,25 +80,29 @@
...
-                  <div class="default text">{{if (eq this.tickets.length 1) (max ticket.minOrder 1) 0}}</div>
-                  <div class="menu">
-                    <div class="item" data-value="{{map-value mapper 0}}">0</div>
-                    {{#each (range (max ticket.minOrder 1) (add ticket.maxOrder 1)) as |count|}}
-                      <div class="item" data-value="{{map-value mapper count}}">{{count}}</div>
-                    {{/each}}
-                  </div>
-                </UiDropdown>
-              </div>
+              {{#if (sub ticket.quantity (add ticket.orderStatistics.tickets.placed ticket.orderStatistics.tickets.completed))}}
```
iamareebjamal, line 83: .

MeghalBisht: Ok understood, waiting for @SARTHAK1SINGH to follow up else I'll complete it

MeghalBisht:
> If non-organizers are able to see order statistics, it is an issue, please open a bug for this in server
Negative

> For showing sold out, there needs to be a new route which returns which ticket is available and which is sold out
I guess, this needs server modification as well ?

iamareebjamal:
Yes

MeghalBisht:
Working on the server part, will reopen this after that

MeghalBisht:
@iamareebjamal  For the server part, if I add a sold-out column for each ticket will that be a good solution? If not please provide an overview of the approach

MeghalBisht:
As what I think is if I add a column then I need to update its state at multiple places which does not seem to be an optimal solution

iamareebjamal:
/events/ID/tickets/availability route should be created. Which will return each ticket's remaining count

MeghalBisht:
Okay, it should return a boolean as if I return the count then order stats are provided to the purchaser, right?

iamareebjamal:
I thought of boolean but then what if only 4 tickets are left and the user orders 5?

Complete order stats should still not be shown

MeghalBisht:
@iamareebjamal I am sure this is not a good solution, but it works! Please suggest a better way

iamareebjamal:
Move this to a function and call it in init, don't await. Also, set maxQuantity of tickets to min(maxQuantity, availability). and add ticket.remaining, not ticket.available

iamareebjamal:
Also, as the data is discarded after use, there is no use of setting it in the component using this.set

MeghalBisht:
Okay and what about the n² time complexity?

MeghalBisht:
> Also, as the data is discarded after use, there is no use of setting it in the component using this.set

But then how will I get it in the template?

iamareebjamal:
Where are you using it in the template?

app/templates/components/public/ticket-list.hbs
```diff
@@ -80,25 +80,29 @@
...
-                  <div class="default text">{{if (eq this.tickets.length 1) (max ticket.minOrder 1) 0}}</div>
-                  <div class="menu">
-                    <div class="item" data-value="{{map-value mapper 0}}">0</div>
-                    {{#each (range (max ticket.minOrder 1) (add ticket.maxOrder 1)) as |count|}}
-                      <div class="item" data-value="{{map-value mapper count}}">{{count}}</div>
-                    {{/each}}
-                  </div>
-                </UiDropdown>
-              </div>
+              {{#if (lte ticket.available 0)}}
```
MeghalBisht, line 83: for checking

iamareebjamal:
> Okay and what about the n² time complexity?

Here, n is not 1000000, not even 100.

Besides, doing even 2^n operations on even n=1000 is faster than making an API call to the server which takes 700ms, and is done frequently, so clearly, this is not bottleneck.

Besides, complexity matters when there is a better solution, nested looping over 2 arrays is the only solution to join data, even databases do it, although on more efficient data structures

Other way O(n) will be to loop over availability array and use this.store.peekRecord('ticket', id) to get the ticket ID, but I am not sure if it will return the same object or a different one and maybe template won't reactively change in that case.

Your choice

app/templates/components/public/ticket-list.hbs
iamareebjamal: That's not `ticketAvailabilities` which you are setting after awaiting

MeghalBisht:
> Other way O(n) will be to loop over availability array and use this.store.peekRecord('ticket', id) to get the ticket ID, but I am not sure if it will return the same object or a different one and maybe template won't reactively change in that case.

I tried this, but when I reload the page multiple times, sometimes it shows SOLD OUT and sometimes it simply shows every ticket available event when they are SOLD OUT. So going with the nested loop.

app/components/public/ticket-list.js
```diff
@@ -14,6 +14,21 @@ export default Component.extend(FormMixin, {
   orderAmount    : null,
   amountOverride : null,
 
+  init() {
```
iamareebjamal, line 17: Wasn't there already a didInsertElement here?

MeghalBisht: Yes, its still there, we can use both init and didInsertElement right?

iamareebjamal: Better to move to already existing method

MeghalBisht: you mean to shift code back to didInsertElement? 😕

iamareebjamal: Yes, the problem was with await. My comment said move the code to another function and call without await.

MeghalBisht: I am confused now, is the current commit okay?

iamareebjamal: I have fixed it

MeghalBisht: Okay now I understood, was confused with your above comments.
Thanks 👍🏻

MeghalBisht:
Please don't merge

MeghalBisht:
I am getting this error after the last commit
zyx

iamareebjamal:
Which event?

MeghalBisht:
I tested with 2 events I created

iamareebjamal:
Fixed, thank you :+1:

mariobehling:
Thanks!

PR #6282, opened 2021-01-08
feat: Add Join Event button

### LINKED ISSUE #6170: Virtual Events: Add "Join Event" button for Online and Mixed Events
Please implement the following
* Add a "Join Event" button next to the "Join Video" button on the right side bar and on the bottom area of public event pages. 
* Direct the ticket holder to the public event page with the event side panel expanded.
* Also show the button for not-logged in users or users without a ticket, but for them when they click on it display a pop-up "You need to login or get a ticket for this event before you can join it online. [Ok]"
* Instead of the "Join Event" button show a button "Active" when the user is on a page with the side panel active already. When the user clicks the button "Active" expand the side panel if it is not expanded.

Requirement for the Join Event button to appear:
* There must be at least one video room defined

Screenshot from 2020-12-30 08-22-48

iamareebjamal:
Fixes #6170

### DIFF
--- a/app/components/public/stream/join-video.hbs
+++ b/app/components/public/stream/join-video.hbs
@@ -1,8 +1,20 @@
- <a class="ui basic {{text-color @color 'basic' 'inverted'}} button no-wrap"
-  style={{css color=(text-color @color 'grey' 'lightgrey')}}
-  href="{{href-to 'public.stream.view' @event @videoStream.slugName @videoStream}}"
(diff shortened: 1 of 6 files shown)


mariobehling:
Thanks. Works well. Please make the following changes as well:
* Show "Join Event" button first and use the blue button color to make it stand out more
* Add some spacing in between the buttons if the second button moves to another line as in the screenshot below

Screenshot from 2021-01-08 22-41-48

mariobehling:
Merging it as some additional changes will be proposed in follow up issues.

PR #6315, opened 2021-01-11
fix: Allow speaker to be created without session

### LINKED ISSUE #6287: Not able to create a speaker due to Error "Field may not be null"
Trying to create a speaker on next months Codeheat event shows an error message "Field may not be null" even though I filled in all fields.

Compare
https://eventyay.com/events/f4fc71b8/speakers/create

Screenshot from 2021-01-10 05-38-24

iamareebjamal:
Fixes #6287


iamareebjamal:
Because of the structure of the speaker session form, it is not easy to check for new session enabled or not without major refactor, hence using session.title to check if the session form was filled, without which the form cannot be saved

PR #6389, opened 2021-01-15
fix: same dialog design for join event and join video

### LINKED ISSUE #6364: Virtual Events Sidebar: Only provide option to view sidebar for attendees with tickets
When an anonymous users accesses a "Join Video" link they will get a pop up to login or to get a ticket. If the user presses "Cancel" the event page and the side panel  will show up even for anonymous users. Users using the "Join Event" link see another message and do not have the option "Cancel".

Expected is: 
* That only the event page would be visible without the side panel. 
* The pop up message, buttons and design should be same for both "Join Event" and "Join Video" link.

Example https://eventyay.com/e/f4fc71b8?side_panel=true&video_dialog=true

Screenshot from 2021-01-14 14-35-43

Screenshot from 2021-01-14 14-36-00

ranjitkshah:
Fixes #6364 

- [ ] That only the event page would be visible without the side panel.
- [x] The pop up message, buttons and design should be same for both "Join Event" and "Join Video" link.

#### Screenshot:
https://www.loom.com/share/dffc067050e54998b4b17d38589717c7

-
-
-

#### Checklist

### DIFF
--- a/app/components/modals/simple-modal.hbs
+++ b/app/components/modals/simple-modal.hbs
@@ -13,7 +13,7 @@
       {{t-var (or @actionName 'OK')}}
     </button>
   {{/if}}
-  <button type="button" class="ui button" {{action 'close'}}>
+  <button type="button" class="ui button" {{action this.performCancel}}>
     {{t-var (or @cancelName 'Cancel')}}
   </button>
 </div>
--- a/app/components/modals/simple-modal.hbs
+++ b/app/components/modals/simple-modal.hbs
@@ -13,7 +13,7 @@
       {{t-var (or @actionName 'OK')}}
     </button>
   {{/if}}
-    <button type="button" class="ui button" {{action 'close'}} {{action @cancel}}>
+    <button type="button" class="ui button" {{action this.performCancel}}>
     {{t-var (or @cancelName 'Cancel')}}
   </button>
 </div>
--- a/app/components/modals/simple-modal.hbs
+++ b/app/components/modals/simple-modal.hbs
@@ -14,7 +14,7 @@
     </button>
   {{/if}}
   {{#if @cancel}}
-    <button type="button" class="ui button" {{action this.performCancel}}>
+    <button type="button" class="ui button" {{action 'close'}} {{action @cancel}}>
     {{t-var (or @cancelName 'Cancel')}}
   </button>
   {{/if}}
--- a/app/components/modals/simple-modal.hbs
+++ b/app/components/modals/simple-modal.hbs
@@ -19,7 +19,7 @@
   </button>
   {{/if}}
   {{#unless @cancel}}
-  <button type="button" class="ui button" {{action 'cancel'}}>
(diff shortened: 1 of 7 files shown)


app/components/modals/simple-modal.js
```diff
@@ -9,4 +9,10 @@ export default class SimpleModal extends ModalBase {
     this.close();
     this.action();
   }
+
+  @action
+  performCancel() {
+    this.close();
+    this.router.transitionTo('public', { queryParams: { video_dialog: null } });
```
iamareebjamal, line 16: Seriously? This component is not just there for video dialog, it is a generic component

app/components/modals/simple-modal.hbs
```diff
@@ -13,7 +13,14 @@
       {{t-var (or @actionName 'OK')}}
     </button>
   {{/if}}
+  {{#if @cancel}}
+    <button type="button" class="ui button" {{action 'close'}} {{action @cancel}}>
+    {{t-var (or @cancelName 'Cancel')}}
+  </button>
+  {{/if}}
```
iamareebjamal, line 20: Duplication

ranjitkshah: please suggest better approach

iamareebjamal: or

app/components/modals/simple-modal.hbs
```diff
@@ -13,7 +13,7 @@
       {{t-var (or @actionName 'OK')}}
     </button>
   {{/if}}
-  <button type="button" class="ui button" {{action 'close'}}>
+    <button type="button" class="ui button" {{action 'close'}} {{action @cancel}}>
```
ranjitkshah, line 16: @iamareebjamal confuse here, how to use OR condition between actions please suggest

afzal442:
plz also remove close icon from both. not required as cancel button does the job.

iamareebjamal:
No

ranjitkshah:
@iamareebjamal  pls check now

mariobehling (approved):
Works as expected for me. Thanks!

app/components/modals/simple-modal.hbs
```diff
@@ -13,7 +13,7 @@
       {{t-var (or @actionName 'OK')}}
     </button>
   {{/if}}
-  <button type="button" class="ui button" {{action 'close'}}>
+    <button type="button" class="ui button" {{action this.performCancel}}>
```
iamareebjamal, line 16:
```suggestion
  <button type="button" class="ui button" {{action this.performCancel}}>
```

app/controllers/public.js
```diff
@@ -49,4 +49,9 @@ export default class PublicController extends Controller {
       this.transitionToRoute('login');
     }
   }
+
+  @action
+  cancel() {
```
iamareebjamal, line 54: Rename to closeVideoDialog

Issue #5041, opened 2020-09-10
Organization Group and Follow Feature


mariobehling:
- An organization can be one or several accounts 
- The account in charge of an event is the “owner” of the organization
- Only owners are able to delete other accounts from an organization team
- Organizers can be contacted through the contact form in CC the sender should receive the same email
- All accounts that belong to an organizer will receive the email
- Replies will only be possible by email (eventyay should not function as a messaging system)
- Organizers should have a page where
  - The name of the organizer is listed
  - The link to the contact form
  - A button to “become a member” in the organizer group (which results in receiving email updates when the organizer publishes a new event)
  - All public events of the organizer (on top new ones, below previous ones)
- “My Organisations” should be accessible through the dropdown menu below “My Sessions” and above “Manage Events”
- If a user is part of an organization 
  - the events that are linked to the organization are accessible and editable by all other members of that organization
  - therefore such an event would show up on the event dashboard of each member
  - Only the “owner” of an organization would be able to change the organization of an event
  - If an owner wants to change the organization affiliation in the wizard of an event there should be a pop-up “are you sure, you want to change the affiliation of this event?” (Yes/Cancel)

Jahnavi-Jonnalagadda:
@mariobehling, an end-to-end organization web page should be created right?

PR #6442, opened 2021-01-20
feat: group feature implementation

maze-runnar:
Related to https://github.com/fossasia/open-event-frontend/issues/5041

#### Checklist


maze-runnar:
Screenshot from 2021-01-20 11-23-31

app/controllers/groups/create.js
```diff
@@ -0,0 +1,38 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class extends Controller {
+
+  @service errorHandler;
+
+  GroupName ='';
```
iamareebjamal, line 9: Sundaram, have you ever seen a JS variable, at least in this project starting with a capital letter?

app/controllers/groups/create.js
```diff
@@ -0,0 +1,38 @@
...
+      });
+      await myGroup.save();
+      this.notify.success(this.l10n.t('Your group has been saved'),
+        {
+          id: 'group_save'
+        });
+      this.router.transitionTo('groups.list');
+    } catch (e) {
+      console.error('Error while saving group', e);
+      const message = e.errors?.[0]?.detail ?? this.l10n.t('Oops something went wrong. Please try again');
```
iamareebjamal, line 27: You created error handler service if I remember correctly, right?

iamareebjamal:
Check out My Tickets and My Events section and tell me is it similar?

This is not admin section, this is user section feature

app/controllers/groups/create.js
maze-runnar: never 🤦🏻‍♂️

maze-runnar:
> Check out `My Tickets` and `Manage Events` section and tell me is it similar?
> 
> This is not admin section, this is user section feature

Similar in what terms like code or functionality?   Some feature in user section like access-code, videoroom, discount-code are like this.

maze-runnar:
instead of tab I should use a button for group creation?

iamareebjamal:
See the UI

iamareebjamal:
They are not tables

maze-runnar:
Screenshot from 2021-01-22 18-03-05
Screenshot from 2021-01-22 17-58-16
Screenshot from 2021-01-22 17-58-09

app/controllers/groups/create.js
```diff
@@ -0,0 +1,48 @@
...
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class extends Controller {
+
+  @service errorHandler;
+
+  groupName ='';
+
+  groupEvents = [];
```
iamareebjamal, line 11: Why storing it here? You already have group model, Use `group.events`

That's why it is not switching correctly

app/controllers/groups/create.js
```diff
@@ -0,0 +1,48 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class extends Controller {
+
+  @service errorHandler;
+
+  groupName ='';
```
iamareebjamal, line 9: Use `group.name`

app/controllers/groups/create.js
```diff
@@ -0,0 +1,48 @@
...
+
+  @action
+  async submit() {
+    try {
+      this.loading = true;
+      const myGroup = this.store.createRecord('group', {
+        user   : this.authManager.currentUser,
+        name   : this.groupName,
+        events : this.groupEvents
+      });
```
iamareebjamal, line 31: Create the group in route itself. See how it is done in video stream route for create and edit

app/controllers/groups/create.js
maze-runnar: but i am creating it in controller, so on first time there is no model.group.name?

app/controllers/groups/create.js
maze-runnar: ok

maze-runnar: is it wrong to create it in controller?

iamareebjamal: Then you have to do things this way and it won't automatically do things we discussed in call

maze-runnar:
Screenshot from 2021-01-24 22-15-13

app/controllers/groups/create.js
```diff
@@ -6,15 +6,16 @@ export default class extends Controller {
...
-  groupName ='';
-
-  groupEvents = [];
-
   @action
   addEvent(event) {
-    if (!this.groupEvents.includes(event)) {
-      this.groupEvents.push(event);
-    }
+    this.model.group.set('events', [event].concat(this.model.group.events.toArray()));
```
iamareebjamal, line 11: Use pushObject

maze-runnar: ok

app/controllers/groups/create.js
```diff
@@ -6,15 +6,16 @@ export default class extends Controller {
...
+  }
+
+  @action
+  removeEvent(event) {
+    let groupEvents = this.model.group.events.toArray();
+    groupEvents = groupEvents.filter(x => x !== event);
+    this.model.group.set('events', groupEvents);
```
iamareebjamal, line 18: Use removeObject

iamareebjamal:
As edit and create controller have same code, move the logic in GroupForm like in Video Stream route/controller

maze-runnar:
> As edit and create controller have same code, move the logic in GroupForm like in Video Stream route/controller

ok

iamareebjamal:
Cool, last thing would be to use glimmer component instead of classic

maze-runnar:
> Cool, last thing would be to use glimmer component instead of classic

I tried that it is giving some error like `you are passing undefined in action without quotes`

mariobehling:
Thanks. Please make the following changes

* On Event Pages, that are part of a group add a follow button below the calendar.
Screenshot from 2021-01-25 23-20-50

* On the Group Overview Page
  * Make a blue "Create Group" Button to the right of the screen (similar to other Create/Add buttons)
  * Change the following
     * Heading My Groups -> "Groups"
     * Tab Groups -> "My Groups" and show all groups of the account here
     * Tab Create -> "Following" and show all groups that the user follows
  * Move "Save" button above "All events"
  * The Group image has the default image "Event". Please change this to a default image "Group"

mariobehling:
Please fix this:
On the "Create Group" or "Edit Group" page, when a user adds an event to the group from the section "All Events" the event gets added to the area "Group Events". Expected is: 
* That at the same time the event should not show up anymore in the section "All Events". 
* When a user removes and event from the group the event should then show up in the list under "All events" again though.



Screenshot from 2021-01-25 23-30-04

iamareebjamal:
Follow button cannot be added right now because it is not implemented on the server

iamareebjamal:
Since we don't have the source of the events placeholder image, we cannot change it to groups, so I have shown the landing image as banner

iamareebjamal:
Other changes will be followed up in a PR after following feature is implemented

PR #6467, opened 2021-01-23
fix: Sorting logic

iamareebjamal:


mariobehling:
It seems like this PR changed the logic on the https://eventyay.com/events/live page. The standard logic on this specific page was correct: https://eventyay.com/events/live?sort_dir=ASC

But the logic on the "past" page on https://eventyay.com/events/past should be the other way around by default. So before the live page was correct and the past page was wrong. Now the live page is wrong and the past page is correct.

Actually there was not an issue open though as the live page was working fine. However, I think there was an issue open for the admin section with the same feature which is buggy and at certain actions results in a 505 or the sorting does not work in some areas.

iamareebjamal:
The logic here was wrong because if sort dir was ASC, it sorted in descending order. This fixes the internal logic.

The crash of 500 was happening because an event in 2016 had malformed data which was crashing the server, but why was an event from 2016 was being shown in the admin section in events tab? Even when sorting order was set to starts_at desc, because the logic of handling sorting was wrong, it was sorting it in ascending order, so I removed DSC from it and manually fixed the 2016 event, but that removed any sorting from that page, so I created this PR to finally fix the sort logic ASC for ascending and DSC for descending. 

If we need ascending sort in manage events section, I'd raise a PR to do so, will see if it is possible to change sort dir in different tabs according to the past/upcoming, not sure.

PR #6485, opened 2021-01-25
fix: Attendee Form: Simplify form and fix order of fields

### LINKED ISSUE #6454: Attendee Form: Simplify form, fix order, show helper text
The attendee form in the wizard and public form are not ordered in the same way. Instead of alphabetical order they should be ordered according to content logic. Furthermore the implementation of address fields in Home Address, Billing Address, Shipping Address, Work Address is not complete as it is missing city, country etc. and it would be helpful if the helper text was visible in the wizard itself.

Therefore please change the following:
1. Hide the following address options but keep the standard form field "Address". For now we will not use the address options such as:
* Home Address
* Billing Address
* Shipping Address
* Work Address
2. Change the field "Address" to "Address (Street, Building, Number etc.)"
3. Change the order of the fields to the following:
- Job Title
- Phone
- Work Phone
- Age Group
- Gender
- Organisation
- Tax Business Info
- Address (Street, Building, Number etc.)
- City
- State
- Country
- Website
- Blog
- LinkedIn
- Twitter
- GitHub
- Facebook
- Instagram
- Email consent
- Partner contact consent
- Photo & video & text consent
4. Increase the width of the table and include the "Helper Text" below each option (in the same cell)
5. Ensure the order on the public form page of attendees is the same as in the wizard
6. Fix the display of the form type column - show "switch" for switches instead of "checkbox"

Screenshot from 2021-01-21 12-01-19

MeghalBisht:
Fixes #6454

- Hides the following address options but keep the standard form field "Address".
    1. Home Address
    2. Billing Address
    3. Shipping Address
    4. Work Address
- Changes the field "Address" to "Address (Street, Building, Number etc.)"
- Change the order of the fields as described
- Ensure the order on the public form page of attendees is the same as in the wizard

Screenshot -
Screenshot from 2021-01-26 01-47-51

#### Checklist

### DIFF
--- a/app/controllers/public/cfs/new-session.js
+++ b/app/controllers/public/cfs/new-session.js
@@ -7,6 +7,7 @@ export default class NewSessionController extends Controller {
   async save(speakerDetails) {
     this.set('isLoading', true);
     try {
+      await this.model.session.save();
       speakerDetails.sessions.pushObject(this.model.session);
       await this.model.session.save();
       this.notify.success(this.l10n.t('Your session has been saved'),
--- a/app/routes/public/cfs/new-session.js
+++ b/app/routes/public/cfs/new-session.js
@@ -9,30 +9,27 @@ export default class NewSessionRoute extends Route {
 
   async model() {
     const eventDetails = this.modelFor('public');
-    const speakers = await eventDetails.query('speakers', {
-      filter: [
-        {
-          name : 'email',
-          op   : 'eq',
-          val  : this.authManager.currentUser.email
-        }
-      ]
-    });
-    const sessionDetails = await this.store.createRecord('session', {
-      event       : eventDetails,
-      creator     : this.authManager.currentUser,
-      track       : null,
-      sessionType : null,
-      speakers
-    });
     return {
       event : eventDetails,
       forms : await eventDetails.query('customForms', {
         sort         : 'id',
         'page[size]' : 0
       }),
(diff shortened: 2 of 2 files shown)


iamareebjamal:
Previous requirement was to show custom form fields as last

MeghalBisht:
So custom forms should be at last or consent options?

mariobehling:
> Previous requirement was to show custom form fields as last

Good point. I did not consider this in depth before, but now I think it would be good to have the "Consent Options" in a separate box below custom elements.

@MeghalBisht Would be great if you could also do this in this PR. Thanks!

mariobehling:
Still need to change the address field in the wizard.

> Changes the field "Address" to "Address (Street, Building, Number etc.)"

MeghalBisht:
**Current ss -**
Screenshot from 2021-01-27 00-10-25

iamareebjamal:
@MeghalBisht The logic is getting too complicated and changing the logic further in future is only going to complicate things. There is no indication/difference in consent fields on backend to make things easy to sort them. It'll be a maintainence hell moving forward

The easiest thing is to add `position` in custom form model and allow users to reorder the fields in the UI

MeghalBisht:
OK and in the frontend arrows would be added to move fields up and down, right?

iamareebjamal:
No, a drag handle

Issue #6546, opened 2021-01-30
Organizer Area: Speaker View Page shows links even if field was left empty


mariobehling:
There are social media links showing up on the speaker view pages accessible by organizers. If speakers leave the field empty these links show up even though they should not show up.

![Screenshot from 2021-01-30 17-23-54](https://user-images.githubusercontent.com/1583873/106361871-10fc3700-6320-11eb-9f3b-991747708bdf.png)

Creatoon:
@mariobehling Sir not reproducable

![Screenshot_2021-01-30 Aayush Dubey Brom Bones Events Open Event](https://user-images.githubusercontent.com/65965202/106364477-43745700-6355-11eb-830c-f988fbfc154c.png)

PR #6556, opened 2021-01-31
fix: don't show links if field was empty

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/6546

#### Checklist

### DIFF
--- a/app/templates/components/events/view/overview/speaker-view.hbs
+++ b/app/templates/components/events/view/overview/speaker-view.hbs
@@ -21,9 +21,7 @@
                 {{else if (eq field.fieldIdentifier 'gender')}}
                     {{@speaker.gender}}
                 {{else if (or (eq field.fieldIdentifier 'website') (eq field.fieldIdentifier 'github') (eq field.fieldIdentifier 'facebook') (eq field.fieldIdentifier 'twitter') (eq field.fieldIdentifier 'instagram') (eq field.fieldIdentifier 'linkedin'))}}
-                    {{#if (and (not-eq (get @speaker field.fieldIdentifier) 'https://github.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://facebook.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://instagram.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://twitter.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://linkedin.com/'))}}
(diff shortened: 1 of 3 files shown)


mariobehling (approved):
The strange thing is, I cannot see this error appearing anymore on the live site. If this PR ensures the error does not re-appear even better.

app/templates/components/events/view/overview/speaker-view.hbs
```diff
@@ -21,7 +21,9 @@
                 {{else if (eq field.fieldIdentifier 'gender')}}
                     {{@speaker.gender}}
                 {{else if (or (eq field.fieldIdentifier 'website') (eq field.fieldIdentifier 'github') (eq field.fieldIdentifier 'facebook') (eq field.fieldIdentifier 'twitter') (eq field.fieldIdentifier 'instagram') (eq field.fieldIdentifier 'linkedin'))}}
-                    <a class="truncate" href={{get @speaker field.fieldIdentifier}} target="_blank" rel="noopener">{{get @speaker field.fieldIdentifier}}</a>
+                    {{#if (and (not-eq (get @speaker field.fieldIdentifier) 'https://github.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://facebook.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://instagram.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://twitter.com/') (not-eq (get @speaker field.fieldIdentifier) 'https://linkedin.com/'))}}
```
iamareebjamal, line 24: Not good, why the links were being saved in the first place

iamareebjamal (changes requested):

maze-runnar:
now links will save correctly. And keeping `if` for previosly saved links.

iamareebjamal:
No, remove them. That's faulty logic. I'll manually make previous links null

Also, I have to think about why I added this && condition and the test so that it doesn't break anything

maze-runnar:
> No, remove them. That's faulty logic. I'll manually make previous links null

ok

mariobehling:
Yes, the issue is actually that these links exist in the first place. And they also show up on the public profile actually.

Screenshot from 2021-02-01 12-08-04

maze-runnar:
> Yes, the issue is actually that these links exist in the first place. And they also show up on the public profile actually.
> 
> Screenshot from 2021-02-01 12-08-04

after this PR it would be resolved. as links would be save as null if they are not entered

PR #6583, opened 2021-02-02
feat: Reorder Speakers

### LINKED ISSUE #5443: Public Speaker Page/Organizer Speaker Page: Make Speakers Sortable
Currently there is no option to control the order of speakers on the public page. Sometimes important speakers are listed at the end of the speakers list. This is not good. We need a way to sort speakers. Therefore please implement the following:
* On the organizer speakers page provide the option to change the order of speakers. Use the same up and down arrow feature that is available on the ticket list.
* On the feature speaker list and public speaker page of an event show the speakers in the same orders as on the organizer speakers page.

Speakers on the organizer speakers page
Screenshot from 2020-11-01 23-33-48

Public speakers view
Screenshot from 2020-11-01 23-25-59

iamareebjamal:
Fixes #5443

### DIFF
--- a/app/controllers/events/view/speakers/reorder.ts
+++ b/app/controllers/events/view/speakers/reorder.ts
@@ -1,14 +1,44 @@
 import Controller from '@ember/controller';
 import { action } from '@ember/object';
+import { getOwner } from '@ember/application';
+import { inject as service } from '@ember/service';
 import Speaker from 'open-event-frontend/models/speaker';
+import Loader from 'open-event-frontend/services/loader';
 
 export default class SpeakersReorderController extends Controller {
+  @service declare loader: Loader;
+  @service notify: any;
+  @service l10n: any;
+  @service router: any;
+
   useSwap = false;
 
   @action sortEndAction(): void {
-    this.model.forEach((speaker: Speaker, index: number) => {
-      console.log(speaker.name, speaker.order, index)
+    this.model.speakers.forEach((speaker: Speaker, index: number) => {
+      speaker.order = index;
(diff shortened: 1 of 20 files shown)


mariobehling:
Excellent!

Just one small thing. Can the width of the buttons be expanded a bit so they only cover one line like most other buttons. Thanks!

Screenshot from 2021-02-03 10-34-02

mariobehling:
Does not seem to be 100% reliably yet. If I change a lot of speakers it does not seem to save all changes. Maybe the API does not catch everything so fast? Also some speaker appear twice.

Screenshot from 2021-02-03 10-46-11

mariobehling:
When I save and then refresh changes are gone.

iamareebjamal:
Because of the cache, it'll take 10 minutes. I'm thinking of a solution

mariobehling:
> Because of the cache, it'll take 10 minutes. I'm thinking of a solution

Ok, I understand. If the public page is cached that is ok, but the order is also not correct if I save the re-order page and then reload it here https://open-event-frontend-git-fork-iamareebjamal-speaker-order.eventyay.vercel.app/events/fa96ae2c/speakers/reorder

Maybe it works a bit later. Need to test in more detail.

iamareebjamal:
That page is cached as well

iamareebjamal:
Now the reorder page is not cached, so changes should be visible after reload. Public page is cached so changes will take 10 minutes to show and because of this, some speakers may be visible twice

mariobehling (approved):

Issue #6428, opened 2021-01-19
Public Event Page: Large space above ticket box


mariobehling:
The public ticket page of an event shows a large space above the ticket box. The space does not exist in the rich text editor of the description. Possibly it is caused by invisible tags?

![Screenshot from 2021-01-19 17-52-11](https://user-images.githubusercontent.com/1583873/105066698-58143d80-5a7f-11eb-81e0-06dbee260b01.png)

himanshujaidka:
Looking into it

himanshujaidka:
![Screenshot (147)](https://user-images.githubusercontent.com/58654018/105073288-dfc47100-5aac-11eb-9386-5941c01c3b38.png)
@mariobehling It seems to work fine

iamareebjamal:
@mariobehling Event Link?

iamareebjamal:
Not reproducible for me

iamareebjamal:
We need to truncate leading newlines/breaks in the rich text editor

PR #6619, opened 2021-02-05
fix: removing extra line breaks at the end of para on public page

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/6428

#### Checklist

### DIFF
--- a/app/components/widgets/forms/rich-text-editor.js
+++ b/app/components/widgets/forms/rich-text-editor.js
@@ -63,12 +63,11 @@ export default Component.extend({
       const updateValue = () => {
         debounce(this, () => {
           const value = this.editor.getValue();
-          const description = String(value);
-          let x = description.length;
-          while (description.substr(0, x).endsWith('<br>')) {
-            x = x - 4;
+          let description = String(value);
+          while (description.endsWith('<br>')) {
+            description = description.replace(/<br>$/, '');
           }
-          this.setProperties({ _value: description.substr(0, x), value: description.substr(0, x) });
+          this.setProperties({ _value: description, value: description });
         }, 200);
       };
 
--- a/app/components/widgets/forms/rich-text-editor.js
+++ b/app/components/widgets/forms/rich-text-editor.js
@@ -63,7 +63,12 @@ export default Component.extend({
       const updateValue = () => {
         debounce(this, () => {
           const value = this.editor.getValue();
-          this.setProperties({ _value: value, value });
+          const description = String(value);
+          let x = description.length;
+          while (description.substr(0, x).endsWith('<br>')) {
+            x = x - 4;
+          }
+          this.setProperties({ _value: description.substr(0, x), value:description.substr(0, x) });
         }, 200);
       };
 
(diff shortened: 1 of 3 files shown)


app/controllers/public/index.js
```diff
@@ -13,7 +13,12 @@ export default class IndexController extends Controller {
 
   @computed('model.event.description')
   get htmlSafeDescription() {
-    return htmlSafe(this.model.event.description);
+    const description = String(this.model.event.description);
+    let x = description.length;
+    while (description.substr(0, x).endsWith('<br>')) {
+      x = x - 4;
+    }
+    return htmlSafe(description.substr(0, x));
```
iamareebjamal, line 21: :eyes: ??

iamareebjamal:
Read the last comment on the issue

maze-runnar:
> Read the last comment on the issue

I had read that. Why add truncate, why can't just remove line breaks from end string?

iamareebjamal:
> why can't just remove line breaks from end string?

That's the meaning of truncate. But before saving in the server, not after

maze-runnar:
> But before saving in the server, not after

oh..ok

iamareebjamal:
Use regex to do so

app/components/widgets/forms/rich-text-editor.js
```diff
@@ -63,12 +63,11 @@ export default Component.extend({
       const updateValue = () => {
         debounce(this, () => {
           const value = this.editor.getValue();
-          const description = String(value);
-          let x = description.length;
-          while (description.substr(0, x).endsWith('<br>')) {
-            x = x - 4;
+          let description = String(value);
+          while (description.endsWith('<br>')) {
+            description = description.replace(/<br>$/, '');
```
iamareebjamal: That's not what I meant. There should be no loop

maze-runnar: removed that in the final commit already.

app/components/widgets/forms/rich-text-editor.js
```diff
@@ -63,7 +63,9 @@ export default Component.extend({
       const updateValue = () => {
         debounce(this, () => {
           const value = this.editor.getValue();
```
iamareebjamal, line 65:
```suggestion
          let value = this.editor.getValue();
```

app/components/widgets/forms/rich-text-editor.js
```diff
@@ -63,7 +63,9 @@ export default Component.extend({
...
-          this.setProperties({ _value: value, value });
+          let description = String(value);
+          description = description.replace(/(<br>)*$/g, '');
+          this.setProperties({ _value: description, value: description });
```
iamareebjamal, line 68:
```suggestion
          value = value.replace(/(<br>)*$/g, '');
          this.setProperties({ _value: value, value });
```

iamareebjamal:
Please check now on null and empty descriptions as well

maze-runnar:
yes it's working.

PR #6620, opened 2021-02-05
feat: Add "Reply-to" to organizers in session notify

### LINKED ISSUE #6544: Notify Speakers: Change "Reply-to" to actual organizers instead of eventyay
The first iteration of the submission system used eventyay.com emails. This was good to get it up and running. 

Moving to the next iteration we now need this feature to specifically support communication with organizers. Speakers who receive an email from the submission system should no longer reply to eventyay but instead directly to the organizers of the event.

Therefore please change the following:

* Change the email from/reply as to the organizers of the event. We can show their email addresses here as the organizers should know their own email. Also to enable communication of speakers with organizers their email addresses can be shared here. Please implement this as follows
a) show the info on the notify speakers pop up
b) enter this following as default for sending the email in the code

From: eventyay submission system <submissions@eventyay.com>
Reply-to: Event Organizer 1 <organizer1email@orgdomain.com>, Event Organizer 2 <organizer2email@orgdomain.com> ...

* Below text box add text "A copy of this email will be send to all organizers of this event."

Screenshot from 2021-01-30 13-34-30
Screenshot from 2021-01-30 13-34-04

sachinchauhan2889:
Fixes #6544 

new pr 4
new pr

### DIFF
--- a/app/components/modals/session-notify-modal.hbs
+++ b/app/components/modals/session-notify-modal.hbs
@@ -11,11 +11,27 @@
   {{#if this.mail}}
     <form class="ui form">
       <div class="field">
-        <label>Name and Email</label>
+        <label>{{t 'From'}}</label>
         <Input
           type="text"
-          name="name and email"
+          name="from"
+          @value={{t 'eventyay submission system <submissions@eventyay.com>'}} 
+          disabled />
+      </div>
+      <div class="field">
+        <label>{{t 'To'}}</label>
+        <Input
+          type="text"
+          name="to"
           @value={{this.speakerEmails}} 
+          disabled />
+      </div>
+      <div class="field">
+        <label>{{t 'Reply-to'}}</label>
+        <Input
+          type="text"
+          name="reply to"
+          @value={{this.organizersEmails}} 
           disabled />
       </div>
       <div class="field">
@@ -41,7 +57,7 @@
       </div>
     </form>
     <p class="m-2">
-      You can use these identifiers as placeholders for actual details sent in the email:
+      {{t 'You can use these identifiers as placeholders for actual details sent in the email'}}:
       <div class="ui bulleted list">
         <div class="item"><code>{session_name}</code> - Name of the Session</div>
         <div class="item"><code>{session_link}</code> - Link of the Session</div>
@@ -51,6 +67,10 @@
         <div class="item"><code>{app_name}</code> - Name of the App (Site)</div>
         <div class="item"><code>{frontend_link}</code> - Link of the Frontend Site</div>
       </div>
+    </p>
+    <br/>
+    <p>
+      {{t 'A copy of this email will be send to all organizers of this event'}}.
     </p>
   {{/if}}
 </div>
--- a/app/components/modals/session-notify-modal.hbs
+++ b/app/components/modals/session-notify-modal.hbs
@@ -15,7 +15,7 @@
         <Input
           type="text"
           name="from"
-          @value={{t 'eventyay submission system <submissions@eventyay.com>'}} 
+          @value={{this.mailFrom}} 
           disabled />
       </div>
       <div class="field">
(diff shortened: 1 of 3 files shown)


app/components/modals/session-notify-modal.hbs
```diff
@@ -11,13 +11,29 @@
   {{#if this.mail}}
     <form class="ui form">
       <div class="field">
-        <label>Name and Email</label>
+        <label>{{t 'From'}}</label>
         <Input
           type="text"
-          name="name and email"
+          name="from"
+          @value={{t 'eventyay submission system <submissions@eventyay.com>'}}
```
iamareebjamal, line 18: What is eventyay? This project is open event

iamareebjamal:
This is just the UI, where is the logic for reply to?

sachinchauhan2889:
https://github.com/fossasia/open-event-server/pull/7714

iamareebjamal:
OK

app/components/modals/session-notify-modal.js
```diff
@@ -42,6 +42,14 @@ export default class SessionNotifyModal extends ModalBase {
     return session.speakers.map(speaker => `${speaker.name} <${speaker.email}>`).join(', ');
   }
 
+  @computed('sessionId')
+  get organizersEmails() {
+    const session = this.store.peekRecord('session', this.sessionId);
+    const owner = session.event.get('owner');
+    const ownerEmail = owner.get('fullName') + ' <' + owner.get('email') + '>';
+    return session.event.get('organizers').map(organizer => `${organizer.fullName} <${organizer.email}>`).concat(ownerEmail).join(', ');
```
iamareebjamal, line 50: Is this public info? People can see organizers and their email?

iamareebjamal: If so, it is a bug, please open an issue on the server

app/components/modals/session-notify-modal.js
```diff
@@ -14,6 +14,7 @@ export default class SessionNotifyModal extends ModalBase {
   @tracked subject = '';
   @tracked message = '';
   @tracked bccString = '';
+  @tracked settings = this.modelFor('events.view.sessions.list').settings;
```
iamareebjamal, line 17: No need for tracked. Use settings service

app/routes/events/view/sessions/list.js
```diff
@@ -75,14 +75,17 @@ export default class extends Route.extend(EmberTableRouteMixin) {
 
     const sessionStatesMapPromise = sessionStateMapCached || this.loader.load('sessions/states');
 
-    const [feedbacks, sessions, sessionStateMap] = await Promise.all([feedbacksPromise, sessionsPromise, sessionStatesMapPromise]);
+    const settingsPromise = this.store.queryRecord('setting', {});
```
iamareebjamal, line 78: Why? You need to check your PRs before requesting review if they are ready or contain unneeded changes

mariobehling:
Well-done. 

There is are two issues though: 
1. If there are more organizers, it does not show all reply-to addresses as the field does not change to several lines.
2. If an organizer gets deleted from an event, the organizer still shows up as a reply-to address even though the organizer was already deleted.

mariobehling:
So, there is another issue happening after this PR.

Before the PR a copy of the email was sent to the organizer email. After the update the copy of the email does not go to the organizer email anymore.

sachinchauhan2889:
@iamareebjamal  @mariobehling sir, i have checked out my PR at server side. My PR only adds reply-to field in mail. 
I removed changes(done by my merged PR) in my local setup and checked email still not send to all organizers. 
The only difference I am able to see is reply-to field. There is no CC containing all organizers in mail, neither mail send by looping all organizers. 

My observation :- 
## before

before

## after
after

iamareebjamal:
Yes, this email was never sent to organizers. Even before this PR

The organizer could manually add bcc if they wanted, but it was not automatically done

mariobehling:
Ok, thanks for the info. So, basically before a copy of the email was received by office@fo... and there is no more copy arriving. Let's discuss how we want this to work later. Good job looking into detail! Thanks.

Issue #5763, opened 2020-11-25
Schedule: Implement Feature "My Schedule"


mariobehling:
Please implement the feature "my schedule". Add the option to tick a session and by ticking it adding it to a my schedule filter or page. A my schedule page could be a subpage of the schedule page e.g. https://eventyay.com/e/8fa7fd14/schedule and with a my schedule at https://eventyay.com/e/8fa7fd14/schedule/my/. 

Please add a white form area that the user can tick and by ticking it the user add the session to "my schedule" similar to below.

![Screenshot from 2020-11-25 21-06-51](https://user-images.githubusercontent.com/1583873/100276706-440dfd00-2f62-11eb-9535-4bb870abdf27.png)

The user then has a dedicated my schedule page:
![Screenshot from 2020-11-25 20-58-40](https://user-images.githubusercontent.com/1583873/100276832-728bd800-2f62-11eb-8283-79470135b163.png)

PR #6664, opened 2021-02-10
feat: Add user favourite session

iamareebjamal:
Related to https://github.com/fossasia/open-event-frontend/issues/5763


mariobehling:
@iamareebjamal Ok, there is the option to favourite, but how can I see only favourite sessions? Not seeing the option.

iamareebjamal:
That's being implemented in another PR by Sundaram

mariobehling (dismissed):
Works from what I can see.

Issue #5764, opened 2020-11-25
Schedule: On my schedule page provide the option to share personal schedule with other attendees and show profile pic (or anonymous icon) in public schedule


mariobehling:
It is nice to know who else attends a session. Currently we cannot see who else signs up and plans to attend a session. Therefore we can implement a feature to allow people to share this information.

Please implement the following:
1. After implementing a my sessions page add a button, e.g. a drop-down button, to the my schedule page on the upper area with the following options:
   1. Show my session participation publicly.
   2. Show my session participation to other attendees.
   3. do not show the sessions I participate in.

According to the above choices show the profile image or an anonymous profile to the visitors of the schedule of an event on the bottom of a session info area
a) in the schedule with up to 12 people and behind that the link "View all" linking to the area on the full session page
b) on the session page itself with all attendees below the session info.

![Screenshot from 2020-11-25 21-02-29](https://user-images.githubusercontent.com/1583873/100277516-8e43ae00-2f63-11eb-9a13-88d3c78c194d.png)

PR #6692, opened 2021-02-14
feat: my schedule filter

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/5763

Screenshot from 2021-02-14 08-54-30
Screenshot from 2021-02-14 08-54-41
Screenshot from 2021-02-14 09-00-02

#### Checklist

### DIFF
--- a/app/templates/public.hbs
+++ b/app/templates/public.hbs
@@ -41,27 +41,6 @@
       </div>
       {{#unless this.device.isMobile}}
         <div class="ten wide column">     
-          {{#if (eq this.session.currentRouteName 'public.sessions')}}
-            <div class="d-flex" style="align-items: center;">
-              <h1 class="ui header" id="session-heading">{{t 'Schedule'}}</h1>
-                {{#if this.session.isAuthenticated}}
-                  <LinkTo
-                    class="rounded-default link-item ui right floated ml-auto mr-4 {{if (eq this.params.mySchedule null) 'active'}}"
-                    @route="public.sessions"
-                    @models={{array this.model.id}}
-                    @query={{hash mySchedule=null}}>
-                    <h4>{{t 'Public Schedule'}}</h4>
-                  </LinkTo>
-                  <LinkTo
-                    class="rounded-default link-item ui right floated {{if (eq this.params.mySchedule true) 'active'}}"
-                    @route="public.sessions"
-                    @models={{array this.model.id}}
-                    @query={{hash mySchedule=true}}>
-                    <h4>{{t 'My Schedule'}}</h4>
-                  </LinkTo>
-                {{/if}}
-            </div>
-          {{/if}}
           {{outlet}}
         </div>
       {{/unless}}
--- a/app/templates/public.hbs
+++ b/app/templates/public.hbs
@@ -79,7 +79,7 @@
                   <i class="red heart icon"></i>
                   {{t 'My Schedule'}}
                 </LinkTo>
-                <LinkTo @route="public.sessions" @models={{array this.model.id}} @query={{hash mySchedule=null}} data-tooltip="{{t 'Clear My Schedule Filter.'}}">
+                <LinkTo @route="public.sessions" @models={{array this.model.id}} @query={{hash mySchedule=null}} data-tooltip="{{t 'Clear My Schedule Filter'}}">
                   <i class="minus circle icon" style={{css color='#555'}}></i>
                 </LinkTo>      
               </h4>
--- a/app/templates/public.hbs
+++ b/app/templates/public.hbs
@@ -72,15 +72,11 @@
             <div class="ui basic segment m-0 pt-0">
               <h4 class="d-flex">
                 <LinkTo
-                  class="rounded-default link-item"
+                  class="p-1 mb-1 rounded-default link-item {{if (eq this.params.mySchedule true) 'active'}}"
                   @route="public.sessions"
                   @models={{array this.model.id}}
                   @query={{hash mySchedule=true}}>
-                  {{#if (eq this.params.mySchedule true)}}
-                    <i class="red heart icon"></i>
-                  {{else}}
-                    <i class="red outline heart icon"></i>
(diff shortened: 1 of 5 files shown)


mariobehling:
Which issue does it solve?

Where is the button to see "My schedule"?

maze-runnar:
> Where is the button to see "My schedule"?

In the rightside, in the screenshot, above tracks.

maze-runnar:
> Which issue does it solve?

Adding my Schedule filter to schedule.

mariobehling:
Please work according to best practices. The issue that I see comes closest to this one https://github.com/fossasia/open-event-frontend/issues/5763

Here we discussed making a dedicated page. If we change the requirement to filters on the same page you need to mention this and a decision has to be made. Follow-up PRs are open which enable attendees to share their schedule publicly. I am not sure if the current approach would enable this.

@iamareebjamal Please help to clarify.

mariobehling:
Please match a PR with an issue always. We are loosing track what is going on and what development decision has been made.

iamareebjamal:
@mariobehling A separate page will duplicate the code a lot and will be significantly difficult to implement than adding a filter. By all intents and purposes, the filtered page is a separate page in itself as it has a separate link that can be copied and pasted, and it also provides features to change the timezone and search and filter according to date and tracks

If let's say we only show favourite sessions in a separate page, without timezone switcher and any other filter feature, even without the date grouping, even then it would take significant effort and duplication to implement. What benefit does implementing a separate page give us?

iamareebjamal:
> Follow-up issues are open which enable attendees to share their schedule publicly. I am not sure if the current approach would enable this.

A different page would also not enable this as there is no feature currently which enables other users to see favourites of other users, this requirement was added after the API was finalized as far as I know. This feature needs more implementation discussion as I don't know how it would be possible in current API structure. Either we can make all user's favourite public or all user's favourite private (this is default for now), as there is no option to store different setting for user currently

mariobehling:
Thanks @iamareebjamal. Let's discuss this in the weekly meeting, but I see an increasing tendency that our issues and what is done in PRs divert. There are a number of issues regarding the my schedule feature. I have labeled the most important ones high priority: https://github.com/fossasia/open-event-frontend/issues?q=is%3Aopen+label%3A%22Priority%3A+High%22+my+schedule

Attendees of virtual events say they want to be able to see what others are doing, hence in November I opened an issue for the feature to enable attendees to share their schedule and attendance in other sessions. We are simply following existing solutions like sched.

iamareebjamal:
We'll be able to show which people will attend a session like here - https://github.com/fossasia/open-event-frontend/issues/5764
This API is implemented and ready to be integrated in UI (except the anonymous part)

The implementation problem is with personal link of a particular user showing which session a particular user has favourited. I did not see any issue for a user-specific link, and thus the API was not planned

For example, https://github.com/fossasia/open-event-frontend/issues/5763 this issue mentions `eventyay.com/e/8fa7fd14/schedule/my` as the URL, this is not user specific as it does not contain any user ID and thus, a user cannot share it with any other user to show which session they have favourited, and thus I did not create any public API for this

The above issue only mentions a separate section for viewing the sessions I liked, for which implementing a filter is the easiest way. 

All the issues mentioned related to `My Schedule` feature including https://github.com/fossasia/open-event-frontend/issues/5764 are still implementable. 

Sharing my schedule with someone else is not implementable as it was not mentioned in any of the above issues, and thus relevant APIs were not created.

So, right now we have:

- [x] Ability to add/remove sessions as favourite
- [x] Ability to see which sessions I favourited
- [x] Ability to see which sessions I'm a speaker in
- [ ] Ability to see who favourited a particular session (to be implemented)

I think all of the above cover what's mentioned and discussed in the My Schedule feature except marking a profile anonymous.

mariobehling:
> The implementation problem is with personal link of a particular user showing which session a particular user has favourited. I did not see any issue for a user-specific link, and thus the API was not planned

Ok, we should have talked about this more in detail. Sched provides an option where users can click on an image of a person and see their entire schedule if they gave the permission to do that. This is why I mentioned the permissions in https://github.com/fossasia/open-event-frontend/issues/5764.

The public "my schedule" page should work as a kind of hub to see 
* the activity of a person in an event
* the profile
* the speaker sessions
* a contact button to get in touch or chat directly with the user.

So, the private "my schedule" should be the first steps towards this goal. Not everything must be possible right now. We already know that direct chatting is not so easy to set up, but the architecture should be done in a way that we can move in this direction.

Regarding the URLs you are right and we would still need to define what would be a public URL.

mariobehling:
As we go with the filter for the start, please move the filter to the top right and of the central area and show two buttons a) Public Schedule and b) My Schedule. Make the active schedule darker.

Please also rename the heading "Sessions" to "Schedule".

mariobehling:
Screenshot from 2021-02-14 11-23-13

maze-runnar:
Screenshot from 2021-02-14 09-00-02

iamareebjamal:
Add more padding to `Public Schedule` and `My Schedule`

app/templates/public.hbs
```diff
@@ -40,9 +40,30 @@
         {{/if}}
       </div>
       {{#unless this.device.isMobile}}
-      <div class="ten wide column">
-        {{outlet}}
-      </div>
+        <div class="ten wide column">     
+          {{#if (eq this.session.currentRouteName 'public.sessions')}}
+            <div class="d-flex" style="align-items: center;">
+              <h1 class="ui header" id="session-heading">{{t 'Schedule'}}</h1>
```
iamareebjamal, line 46: Why adding this in public? Why not in sessions.hbs?

iamareebjamal:
Change `mySchedule` to `my_schedule`

And add the variable in sessions controller so `Public Schedule` is highlighted by default.

maze-runnar:
Screenshot from 2021-02-15 16-15-42
Screenshot from 2021-02-15 16-15-37

PR #6728, opened 2021-02-16
fix: Add top margin to footer

iamareebjamal:
image

As discussed in meeting

### DIFF
--- a/app/components/footer-main.js
+++ b/app/components/footer-main.js
@@ -8,7 +8,7 @@ import { sortBy } from 'lodash-es';
 
 @classic
 @tagName('footer')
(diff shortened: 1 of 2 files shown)


mariobehling:
Very good. It could even be a slight bit more.

Currently it does not look like it is applied everywhere, e.g. not here
https://open-event-frontend-git-fork-iamareebjamal-footer-margin.eventyay.vercel.app/e/fa96ae2c/cfs/session/6890
https://open-event-frontend-git-fork-iamareebjamal-footer-margin.eventyay.vercel.app/e/fa96ae2c/cfs

PR #6785, opened 2021-02-21
feat: Public Schedule: Add dropdown filter for session typ

### LINKED ISSUE #6752: Public Schedule: Add dropdown filter for session type
The public schedule view would benefit if users could choose session types, e.g. some users might be looking for workshops in a schedule. Therefore below the "Rooms" filter please provide a dropdown list with a tickbox for session types. So, the idea is that a user could tick the different workshop session types that are available at an event.

Screenshot from 2021-02-18 10-58-17

MeghalBisht:
Fixes #6752

Screenshot from 2021-02-21 21-03-24
Screenshot from 2021-02-21 21-03-14
Screenshot from 2021-02-21 21-03-08

#### Checklist


mariobehling:
Thanks!

Please allow to set several session types at once, e.g. the user should be able to filter on all workshops in the FOSSASIA Summit: https://open-event-frontend-git-fork-meghalbisht-session-type-eventyay.vercel.app/e/fa96ae2c/schedule

app/routes/public/sessions.js
```diff
@@ -84,6 +87,21 @@ export default class SessionsRoute extends Route {
       });
     }
 
+    if (params.sessionType) {
+      const sessions = params.sessionType.split(',');
```
MeghalBisht, line 91: @iamareebjamal is there a better way to retrieve session types? This fails when sesison types themseleves have a `,`

mariobehling:
Good. The feature is working now. First when I tested it, it showed an error page, but after a few minutes the error did not appear anymore.

We need to think now how we can have the view well for the user.

Right now after the user selected sessions, there is no visual indication that filters are selected. How should this be? Maybe we should have a heading "Session Types" and below that the selected filter show up in small font? Or do you have a better idea? We need to have a visual indication that a filter is selected here. Otherwise the user might get confused.

We also need a clear filter option like the other filters.

MeghalBisht:
> Right now after the user selected sessions, there is no visual indication that filters are selected. How should this be? Maybe we should have a heading "Session Types" and below that the selected filters show up in small font?

> We also need a clear filter option like the other filters.

Currently the logic needs to be reviewed, once done then I will do this in a follow up PR.

mariobehling:
Thanks. Sounds good to me. Logic is working from user perspective.

iamareebjamal:
Logic is correct, please go ahead

MeghalBisht:
**update**
Screenshot from 2021-02-27 17-10-58

mariobehling:
Good! Working as expected. Only need to finalize the following, please.

1. When the user selects a session type do not close the dropdown. Keep it open.
2. Please change the design as in the screenshot:
* Move "Session Types" to above the dropdown
* Active Session Types -> Session Types
* List active type in the same row and do not show a black dot (this is a design bug that needs to be change for "Rooms" as well)

109386146-c968cb00-791e-11eb-8905-e7d7463f3198

MeghalBisht:
> * List active type in the same row and do not show a black dot (this is a design bug that needs to be change for "Rooms" as well)

Should I change for rooms as well? What about tracks?

mariobehling:
Well, would be great if you could take out the big dots in front of rooms and part from that leave as is. Tracks have a color. So, that is perfect.

MeghalBisht:
Screenshot from 2021-02-27 19-26-12

mariobehling (approved):
Thank you!

PR #6822, opened 2021-02-24
feat: Add user schedule page

iamareebjamal:
image


mariobehling:
Thanks, but I guess there are still permission issues. I cannot access your schedule for example here: https://open-event-fro-git-fork-iamareebjamal-user-schedule-e-25b911.vercel.app/e/fa96ae2c/schedule/user/2770

Screenshot from 2021-02-25 09-54-59

iamareebjamal:
It is work in progress. Server PR was not deployed, try now

mariobehling:
Ok, cool it is working now. However, I am able to click the tick icon and take out sessions from the schedule.

Is it because I am an admin?

How should this feature work?

Should other users see the tick and simply not be able to change the tick on off? 
Or
Should other users see the session unticked and be able to add a tick and therefore add a session to their own "My Schedule" (which would be nice)?

iamareebjamal:
The tick show only show if you have added it in your schedule, which is broken currently and I'm trying to fix it

> Should other users see the session unticked and be able to add a tick and therefore add a session to their own "My Schedule" (which would be nice)?

This would be the final behavour

iamareebjamal:
@mariobehling This is as good as I could make it to work. Please check

mariobehling:
Thanks, I am not able to select the tick box, but this issue was solved already. So, I guess it will be working after I update the branch.

mariobehling:
All my scheduled session appear unselected.

Screenshot from 2021-02-27 18-14-54

iamareebjamal:
Yes, there's another bug. I recommend we hide the checkbox on detail and user schedule page otherwise it will be buggy and inconsistent. I have tried almost everything but something breaks anyway

mariobehling:
> Yes, there's another bug. I recommend we hide the checkbox on detail and user schedule page otherwise it will be buggy and inconsistent. I have tried almost everything but something breaks anyway

Ok, can we hide it now and someone can re-add it later and test in detail? Still, it should work on the general schedule page.

mariobehling (approved):
Ok, works that way. I regularly receive an error page, but at reload the error is gone. Difficult to replicate. I wonder if you see something in Sentry.

PR #6827, opened 2021-02-25
fix:  Cannot progress without adding form fields in "switched off" sponsor step "Field may not be null"

### LINKED ISSUE #6816: Wizard: Cannot progress without adding form fields in "switched off" sponsor step "Field may not be null"
When a user creates an event in the wizard and moves forward to the sponsor step, the user cannot progress without adding form fields in the "switched off" sponsor step. Error: "Field may not be null".

Expected: It should not be necessary to enter any sponsor.

Screenshot from 2021-02-24 11-13-57

sachinchauhan2889:
Fixes #6816 

When a user creates an event in the wizard and moves forward to the sponsor step, the user cannot progress without adding form fields in the "switched off" sponsor step. Error: "Field may not be null".

This PR fix this issue.

#### screenshots
Not required as it is not UI fix. You need to run deployment for reviewing.


app/templates/components/forms/wizard/sponsors-step.hbs
```diff
@@ -20,7 +21,7 @@
     </div>
   </div>
 
-  {{#if this.data.event.isSponsorsEnabled}}
```
sachinchauhan2889, line 23: this is done because when we toggle wizard pages, sometimes checkbox is pre-checked even if this.data.event.isSponsorsEnabled is false. i have used this.data.sponsors.length and it is working fine.

Issue #6823, opened 2021-02-25
Unexpected Error when joining a session


mishari:
**Describe the bug**
When joining a session, I get the error "Oops, the page you were trying to access has caused an error. ."

**To Reproduce**
Go to https://eventyay.com/e/fa96ae2c/session/6707
click on "Join Event"

**Screenshots**

<img width="873" alt="image" src="https://user-images.githubusercontent.com/503018/109084862-286dea80-773b-11eb-9158-6077c518e8b8.png">

**Desktop (please complete the following information):**
OSX and Android
Latest Firefox, Chrome, and Safari

PR #6795, opened 2021-02-22
fix: join event button should open side_panel on same page.

sachinchauhan2889:
#### Checklist

- [X] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [X] My branch is up-to-date with the Upstream `development` branch.
- [X] The acceptance, integration, unit tests and linter pass locally with my changes 
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)

PR #6851, opened 2021-02-27
fix: session view page error on clicking join event

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/6823

https://github.com/fossasia/open-event-frontend/pull/6795

https://user-images.githubusercontent.com/56407566/109376043-cb636780-78e7-11eb-9914-3c9dbae8964d.mp4

#### Checklist


app/components/public/stream/join-video.ts
```diff
@@ -36,11 +36,15 @@ export default class JoinVideo extends Component<Args> {
 
   @action
   openPanel(): void {
+    let currentRoute  = this.router.currentRoute.name;
+    if (currentRoute === 'public.session.view' || currentRoute === 'public.exhibition.view') {
+      currentRoute = 'public';
```
iamareebjamal, line 41: Why?

maze-runnar: on the session view page it needs session-Id also to transition.

iamareebjamal: There must be a better solution

maze-runnar:
https://user-images.githubusercontent.com/56407566/109525634-ba794880-7ad7-11eb-95ed-785876f9e672.mp4

mariobehling:
Does it work if you click on "Join Video" on the schedule and the session page itself - in the green area in the below screenshot?

Screenshot from 2021-03-01 19-41-08

iamareebjamal:
Instead, pass in a callback in args which supply their own route instead of parsing the current route. It doesn't belong in the component and is brittle

maze-runnar:
I couldn't find another way.

iamareebjamal:
@maze-runnar See the solution

PR #6854, opened 2021-02-27
chore: Make root font size 16px

iamareebjamal:


mariobehling (approved):
Nice. If you agree, I would go with this change. The main issues are with buttons, but many buttons had issues before and we had already discussed to change some spacing in buttons. Apart from that the social icons on the right-hand side need to be adapted again as they were just fitting nicely before this change here.

But generally I think this improves readability a lot.

mariobehling:
So, we want to keep the size of those buttons as is if possible.

Screenshot from 2021-02-27 21-04-32

PR #6970, opened 2021-03-09
feat: allow multiple room track filters, allow enable disable using clicks

### LINKED ISSUE #6930: Schedule: Make several filters active on click and activate filter on one click and deactivate the filter when user presses it again
Currently the user can only filter one date, track or room at the same time. 

1. Please allow the user to choose several dates, tracks or rooms.
2. When the user clicks on a tracks, rooms, or date filter the filter gets activated and stay activated, but the user should be able to select an additional date, track or room at the same time.
3. And it should work the other way around: When the user clicks on the filter that is already activated the same track, room or date again the filter option should get deactivated.

Screenshot from 2021-03-07 00-12-03

MeghalBisht:
Fixes #6930

https://user-images.githubusercontent.com/61330148/110539242-0026b880-814b-11eb-99a1-34e485a8f293.mp4

#### Checklist

### DIFF
--- a/app/controllers/public.js
+++ b/app/controllers/public.js
@@ -13,6 +13,10 @@ export default class PublicController extends Controller {
 
   @tracked activeSession = this.router.currentRoute.queryParams.sessionType ? this.router.currentRoute.queryParams.sessionType.split(',') : [];
 
+  @tracked activeRoom = this.router.currentRoute.queryParams.room ? this.router.currentRoute.queryParams.room.split(':') : [];
+
+  @tracked activeTrack = this.router.currentRoute.queryParams.track ? this.router.currentRoute.queryParams.track.split(':') : [];
+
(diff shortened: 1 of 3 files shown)


mariobehling:
Good.

If I click the filter several times it gets applied several times.

When the filter is activated it should become a gray kind of button (like before). Otherwise the user cannot see which filters are applied.

MeghalBisht:
> If I click the filter several times it gets applied several times.

This does not happen on my testing and the logic allows filter to toggle on multiple clicks.

> When the filter is activated it should become a gray kind of button (like before). Otherwise the user cannot see which filters are applied.

Will fix this.

MeghalBisht:
**Update** 
Screenshot from 2021-03-10 23-51-29

mariobehling (approved):
Works for me. Thank you!

PR #6980, opened 2021-03-11
fix: scroll video stream side panel

### LINKED ISSUE #6975: Virtual Events Side panel menu items moving upwards in smaller windows
If the browser window is shorter the side panel in virtual events moves menu items upwards that overlap above items. Expected is that the items stay as expected and that video rooms create an internal scroll bar as in the previous implementation.

Screenshot from 2021-03-10 10-16-34

sachinchauhan2889:
Fixes #6975

### DIFF
--- a/app/styles/components/stream.scss
+++ b/app/styles/components/stream.scss
@@ -59,7 +59,8 @@
   padding-top: 0 !important;
   background-color: #555;
   color: #fff;
-  overflow: auto;
+  overflow: hidden;
+  overflow-y: auto;
   z-index: 1;
 }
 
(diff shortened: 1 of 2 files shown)


app/styles/components/stream.scss
```diff
@@ -59,7 +59,7 @@
   padding-top: 0 !important;
   background-color: #555;
   color: #fff;
-  overflow: hidden;
+  overflow: auto;
```
iamareebjamal, line 62: Only overflow-y needed

iamareebjamal: And not in this CSS

PR #6981, opened 2021-03-11
feat: Add Confirmation Modal before Chat page

sachinchauhan2889:
This PR add Confirmation Modal before directing to Chat page.


iamareebjamal:
This should be in chat component, not side panel

iamareebjamal:
Now that I think of it, I'm not sure. Will think of what should happen. But this dialog should only open once for the user if they confirm it

mariobehling:
Here is an idea: 

> Please confirm you understand and agree to the conditions of using the chat!
> 
> If you join the event chat your profile name and image will be visible to other attendees. Other event attendees can also contact you directly.
> 
> You may change your chat name and chat profile picture by going to account settings on the chat page on the top left. You need to minimize the side panel to access it. The feature integration is still in Alpha stage and currently your profile on the eventyay account page and on the chat are not linked and can be independently edited. When you change the chat settings you may receive additional email confirmations.
> 
> Do you want to use the chat now?
> 
> OK | Cancel

PR #7027, opened 2021-03-18
feat: show BBB recordings

MeghalBisht:
reqd: fossasia/open-event-server#7838

Screenshot from 2021-03-19 05-01-59

#### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
 #MAPBOX_ACCESS_TOKEN = "Access Token"
-API_HOST=http://localhost:5000
+API_HOST=https://open-event.dokku.fossasia.org
 MAP_DISPLAY=embed
 FASTBOOT_DISABLED=true
 #HCAPTCHA_SITE_KEY=""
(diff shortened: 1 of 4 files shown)


app/components/forms/events/view/videoroom-form.hbs
```diff
@@ -134,6 +134,18 @@
       {{t 'Add'}}
     </button>
   </div>
+  {{#if (and this.recordings (eq this.data.stream.videoChannel.provider 'bbb'))}}
```
iamareebjamal, line 137: Instead, simply check this.videoRecordings

.env.example
```diff
@@ -1,5 +1,5 @@
 #MAPBOX_ACCESS_TOKEN = "Access Token"
-API_HOST=https://open-event.dokku.fossasia.org
+API_HOST=http://localhost:5000
```
iamareebjamal: Revert

MeghalBisht: yes, done already

mariobehling:
Thank you, it is good that the feature is there. Right now we do not have a place where we have an overview of recordings. We need to think this through, but for now it seems ok to have it here.

PR #7101, opened 2021-04-06
fix: only allow event owners to add event to group

maze-runnar:
Fixes #as per discussion

-  only allow event owners to add event to group

#### Checklist

### DIFF
--- a/app/components/forms/group/group-form.js
+++ b/app/components/forms/group/group-form.js
@@ -40,7 +40,7 @@ export default class GroupForm extends Component.extend(FormMixin) {
 
   @computed('events.[]', 'group.events.[]')
   get remainingEvents() {
-    return this.events.toArray().filter(event => !this.group.events.toArray().includes(event));
+    return this.events.toArray().filter(event => (!this.group.events.toArray().includes(event) && event.hasAccess(this.authManager.currentUser.email)));
   }
 
   @action


app/components/forms/group/group-form.js
```diff
@@ -40,7 +40,7 @@ export default class GroupForm extends Component.extend(FormMixin) {
 
   @computed('events.[]', 'group.events.[]')
   get remainingEvents() {
-    return this.events.toArray().filter(event => !this.group.events.toArray().includes(event));
+    return this.events.toArray().filter(event => (!this.group.events.toArray().includes(event) && event.hasAccess(this.authManager.currentUser.email)));
```
iamareebjamal, line 43: Won't hasAccess be true for organizers as well?

app/models/event.js
```diff
@@ -210,4 +210,7 @@ export default class Event extends ModelBase.extend(CustomPrimaryKeyMixin, {
         || this.coorganizers.includes(currentUser));
   }
 
+  hasOwnerAccess(currentUser) {
+    return currentUser && (currentUser.isAnAdmin || currentUser.email === this.owner.get('email'));
```
iamareebjamal, line 214:
```suggestion
    return currentUser && (currentUser.isAnAdmin || currentUser.id === this.owner.get('id'));
```

iamareebjamal:
Still showing events that I am coorganizer in, not owner

iamareebjamal:
OK, so it shows me those events because I am an admin, that shouldn't be possible

iamareebjamal:
Now it's not showing any event

klsingh016:
can i work on this?

PR #7147, opened 2021-04-16
feat: Add 3cx as video channel

MeghalBisht:
from meeting notes

**There is no prefix for 3cx web meetings so this is more like a custom option as I can even add any other url and it would work fine.**

Screenshot from 2021-04-17 00-09-19
Screenshot from 2021-04-17 00-10-24

#### Checklist


app/components/forms/events/view/videoroom-form.hbs
```diff
@@ -62,6 +62,11 @@
           @site="vimeo"
           @value={{this.data.stream.url}}
           @onChange={{action (mut this.data.stream.url)}} />
+      {{else if (eq this.data.stream.videoChannel.provider '3cx')}}
+        <Widgets::Forms::LinkField
+          @inputId="url"
+          @value={{this.data.stream.url}}
+          @onChange={{action (mut this.data.stream.url)}} />
```
iamareebjamal, line 69: Isn't it same as else?

MeghalBisht: Yes, will fix it. 
But is this correct? Why do we need another channel when it is similar to custom?

app/components/public/stream/video-stream.ts
```diff
@@ -50,6 +53,9 @@ export default class PublicStreamVideoStream extends Component<Args> {
 
     if (provider === 'jitsi') {
       this.loading = false;
+    } else if (provider === '3cx') {
+      this.cxUrl = stream.url;
```
iamareebjamal, line 57: Use iframeUrl

app/components/public/stream/video-stream.hbs
```diff
@@ -14,6 +14,9 @@
   {{#if (and (eq @videoStream.videoChannel.provider 'vimeo') this.vimeoId)}}
     <iframe src="https://player.vimeo.com/video/{{this.vimeoId}}?autoplay={{if @videoStream.extra.autoplay 1 0}}&loop={{if @videoStream.extra.loop 1 0}}" width="100%" height="100%" frameborder="0" title="Vimeo Live Stream" allow="autoplay" allowfullscreen="true"></iframe>
   {{/if}}
+  {{#if (and (eq @videoStream.videoChannel.provider '3cx') this.cxUrl)}}
```
iamareebjamal, line 17: iframeUrl case should be reused here

app/components/forms/events/view/videoroom-form.hbs
iamareebjamal: Custom won't show in iframe

Issue #6813, opened 2021-02-24
Profile Image: If user has no profile image use Gravatar


mariobehling:
Many users already have Gravatar. Please use a gravatar if the user has no profile image to display on attendee lists and other places.

Yousufkhan1:
if u want i can?

iamdevvalecha:
starting this issue.

tanay123456789:
working on this issue

tanay123456789:
work done.

PR #7318, opened 2021-05-22
feat: Using Gravatar if user has no profile image

sachinchauhan2889:
Fixes https://github.com/fossasia/open-event-frontend/issues/6813

Where gravatar will appear:-

- attendee list
- speaker-item
- session-item

### DIFF
--- a/app/components/image.hbs
+++ b/app/components/image.hbs
@@ -1,0 +1,1 @@
+<img alt={{this.alt}} class={{this.classes}} src={{this.link}}>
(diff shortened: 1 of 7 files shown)


ranjitkshah (approved):

MeghalBisht (commented):
Works! I tested the following -
1. Attendees of a session
2. Session speakers

@sachinchauhan2889 Can we also add it on the speaker tab of events?
Screenshot from 2021-05-24 20-55-06

maze-runnar:
> @sachinchauhan2889 Can we also add it on the speaker tab of events?

I think it was only for public pages.

MeghalBisht (approved):

mariobehling:
As discussed I am closing this issue. Best would be to have a solution without gravatar.com.

PR #7332, opened 2021-05-25
fix: Text changes from record to enable recording.

iamdevvalecha:
Fixes:- text changes of record to enable recording in video-room(bbb).


iamdevvalecha:
@MeghalBisht check!

ranjitkshah (approved):

PR #7559, opened 2021-07-14
feat: add isCfsEnabled field and save announcement

### LINKED ISSUE #5315: Add CfS enabled field in event
> Not deleting the announcement means that we'll have to add another field in DB and API and also add validation that enabling CFS actually contains CFS and disabling CFS actually has no CFS.

It would be better if the announcement would not be deleted when it gets deactivated. Reason is: Organizers play around with features. Also they could click a field by accident. There are many reasons organizers would not want data (like the announcement) to be deleted. So, while deleting the announcement solves the current issue quickly it should not be a final fix.

As a matter of fact if we delete the data we should actually have a pop up "You are deleting the announcement. Yes/No". This is all too much to go through. From the user perspective the best would be if the data does not get deleted simply.

So, I suggest:
1. Delete it for now.
2. Open an issue and change the functionality in a way that no data gets deleted when a field gets deactivated (just like most of the other fields).

Does this work?

_Originally posted by @mariobehling in https://github.com/fossasia/open-event-frontend/pull/5302#issuecomment-708691864_

sachinchauhan2889:
Fixes #5315 

Required https://github.com/fossasia/open-event-server/pull/8100


MeghalBisht (approved):

sachinchauhan2889:
Need to manual migration.
If speaker_call.announcement !=Null
then set is_cfs_enabled = true

sachinchauhan2889:
`UPDATE events SET is_cfs_enabled = 't' FROM speakers_calls WHERE speakers_calls.event_id = events.id AND speakers_calls.announcement IS NOT NULL;`

Issue #7601, opened 2021-07-23
Groups: Add "Followers" page for Owners/Organizers


mariobehling:
The group information is implemented and following is possible. Group organizers do not have the possibility to view the members yet though. Therefore please

Implement the following:
- [x] Add a tab "Followers" behind the "Events" tab on the groups edit page for organizers
- [x]  * On this "Followers" page (for organizers) show a table of followers with user image (if made public), public profile name, group join date, email address
- [ ] Also implement a CSV export for this table

- [x] Change the implementation when users click a "Follow" button of a group as follows:
 Implement a pop up with the message: "By following this group you accept that: 1. You will receive announcement emails for events of this group. 2. You share your email with the group owners and organizers. 3. All group members will be able to see your public profile image and name. Do you want to proceed? Yes, follow this group. / Cancel"

![Screenshot from 2021-07-23 17-13-30](https://user-images.githubusercontent.com/1583873/126803835-fbf01aa6-6a44-4fa9-846f-4ad918cb41b1.png)

PR #7620, opened 2021-07-27
feat: Groups: Add "Followers" page for Owners/Organizers

MeghalBisht:
Fixes parts of #7601 
To be followed up -  CSV export part

Screenshot from 2021-07-29 03-38-28
Screenshot from 2021-07-29 03-38-24

#### Checklist


maze-runnar (dismissed):

maze-runnar:
Screenshot from 2021-07-27 22-37-06

maze-runnar:
@MeghalBisht you will create new PR for CSV export, right?

app/templates/groups/edit/followers.hbs
```diff
@@ -0,0 +1,16 @@
+<GroupNav @group={{this.model.group}} />
```
MeghalBisht, line 1: @maze-runnar After adding the navigation, if you try to save the event settings or events to the group it throws an error.

maze-runnar: Oh, I will check and try to fix it. If not possible then will revert.

maze-runnar:
After visiting to followers route, whether the group-nav is added or not , saving event is giving error. And i think followers should not be in edit section.

mariobehling:
I think the slugs we are implementing do now always make sense, but I do not have the perfect solution right now. In any case the followers list will be editable in some ways by organizers, but these are future issues (e.g. organizers should be able to throw out a user). There could also be a public followers list, but again it is a future issue. So, this looks like a good approach to me in general.

mariobehling:
I missed a point in the issue description. We need a "followers icon" behind the "Events" icon on the groups overview page. The followers icon should be different to the team icon. Could you add it, please as well?

Screenshot from 2021-07-28 13-44-11

I see that the page is not working on the test deployment. I guess it is because we are using the production API server for testing. Have you been able to see how the PR frontend deployments on Vercel could be changed to use https://test-api.eventyay.com instead?

maze-runnar:
@MeghalBisht I have changed the path for the followers page and added follower icon. if it's ok, I can push changes?

Screenshot from 2021-07-28 22-25-09
Screenshot from 2021-07-28 22-25-05

MeghalBisht:
> @MeghalBisht I have changed the path for the followers page and added follower icon. if it's ok, I can push changes?

I tried this too. But after you visit follower's page and again try to go back to edit events or settings page, the patch request fails. Did you check that?

maze-runnar:
> > @MeghalBisht I have changed the path for the followers page and added follower icon. if it's ok, I can push changes?
> 
> I tried this too. But after you visit follower's page and again try to go back to edit events or settings page, the patch request fails. Did you check that?
> 
> 

Yeah, error is not solved 😅

maze-runnar (approved):

Issue #7573, opened 2021-07-16
Admin Sales Page: Export table as CSV


mariobehling:
On the tickets summary page on https://eventyay.com/admin/sales/live please add an option to export the table as a CSV.

Additionally include the following as columns in the CSV:
* Owner Names
* Names and Emails of other organizers 
* Event Creation Date

![Screenshot from 2021-07-16 10-16-42](https://user-images.githubusercontent.com/1583873/125916471-626007f5-198e-42d3-b587-b24eba3f22a6.png)

ranjitkshah:
Looking

Issue #7572, opened 2021-07-16
Admin Sales Page: Make Ticket number and Sales Number sortable through backened


mariobehling:
Tickets and Sales numbers on the admin sales page are computed properties and it is not easily possible to sort them on the frontend itself. Even if we could sort them here it could slow down the page due to potentially large lists.

The sorting is very important to the admin, because it enables the admin to identify top organizers.

Please implement a sorting / list order according to ticket sales and revenue through the backend.

![Screenshot from 2021-07-16 10-05-20](https://user-images.githubusercontent.com/1583873/125914128-b151dbba-c9eb-431b-859f-8072a0188c67.png)

PR #7629, opened 2021-07-29
fix: Admin Sales Page: Make Ticket number and Sales Number sortable

maze-runnar:
~~Fixes https://github.com/fossasia/open-event-frontend/issues/7573~~

Correct Issue:
https://github.com/fossasia/open-event-frontend/issues/7572

Screenshot from 2021-07-29 22-54-40

#### Checklist

### DIFF
--- a/app/controllers/admin/sales/index.js
+++ b/app/controllers/admin/sales/index.js
@@ -8,7 +8,7 @@ export default class IndexController extends Controller.extend(AdminSalesMixin, 
 @or('authManager.currentUser.isSuperAdmin', 'authManager.currentUser.isAdmin') hasRestorePrivileges;
 sort_by = 'starts-at';
 queryParams = ['sort'];
-  
+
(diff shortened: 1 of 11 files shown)


mariobehling:
1. It only sorts the tickets on the current page. It does not sort all ticket. 
2. It is not possible to go to the second and following pages of the list.

We need a list sorting that uses the backend and sorts all events in the system, e.g. past events should be sorted as well in the past tab and we have a large number of events here. Frontend sorting is not sufficient.

mariobehling:
You are referencing the wrong issue. The correct issue is https://github.com/fossasia/open-event-frontend/issues/7572

It already mentions that we need the backend.

maze-runnar:
So we can't achieve that using the backend, as we don't know how to sort using computed property, that's why at every place computed properties are left unsortable. So I am closing this PR.

PR #7693, opened 2021-08-12
fix issues on view page: user able to delete other followers

sachinchauhan2889:
- Now user can follow and unfollow. Only its data will be alter in DB table. Now user can't delete delete data from User_follow_groups Table.
- Count is updating when we alter follow.
- User logo is also updating when we alter follow.

### Before
Before

### After
after


maze-runnar:
@sachinchauhan2889 I am facing this issue on https://open-event-frontend-1k23a087d-eventyay.vercel.app/g/59
Screenshot from 2021-08-13 20-55-27

PR #7717, opened 2021-08-20
chore: gitpod config

### LINKED ISSUE #7716: add gitpod integration
https://gitpod.io

progmatic-99:
Fixes #7716


pc-beast (approved):
Tested. Works fine...But I am not sure that it is needed.

progmatic-99:
this really fastens up the developer feedback and saves time wasted during building of the app. Developer can focus more on the coding part and also this would be really useful for beginners wanting to contribute, as they would not be stuck setting up the project.

maze-runnar:
Merging it for now, if it doesn't found useful, will revert it.

maze-runnar (approved):

pc-beast:
If we are considering online editors for project,  I think we should also consider github.dev

progmatic-99:
@pc-beast github.dev has problems in building projects and installing dependencies. If you can check that, those problems are not there, then that would be awesome. 
Also, we also need to mention in the documentation about gitpod integration, so contributors know about this

maze-runnar:
> @pc-beast github.dev has problems in building projects and installing dependencies. If you can check that, those problems are not there, then that would be awesome. 
> Also, we also need to mention in the documentation about gitpod integration, so contributors know about this

Its not as tough as it sounds in your statement. Here we only have 3 commands. I am merging it but don't change anything in docs for now about it.

progmatic-99:
Gitpod is more feature rich than github, also it's open source. So, it makes sense for us to give preference to open source tools, that's why we use jitsi & bbb instead of google meet.

PR #7726, opened 2021-08-25
fix env variable for authentication

sachinchauhan2889:
fix env variable for authentication.

It fix the error while logging in from modal on event public page.


mariobehling:
Thanks. It is working. I am merging it.

PR #7770, opened 2021-09-09
fix:workaround for user not able to upload slides on old sessions

### LINKED ISSUE #7735: Users of existing events are unable to change sessions because of "session may not be null error" or Test Slides instead of Slides shows up without form field
1. Users are unable to change sessions. Error message `/data/attributes/slides/0/link: Field may not be null`. But: Slides are not enabled even in the Wizard. Therefore there should be no requirement for slides. This is the case for older existing sessions.
2. Newer events show at points the option "Test Slides" instead of "Slides". It seems this was wrongly implemented. The previous slides solution was implemented with a new solution and the update of the functionality was not done properly.

Please resolve this.

Compare: https://eventyay.com/events/bee2414e/session/7219/edit

Screenshot from 2021-08-30 05-33-30

progmatic-99:
Fixes #7735

This is a workaround solution for the issue.
 As mentioned [here](https://github.com/fossasia/open-event-frontend/issues/7735#issuecomment-916334528) 
After the migration is done and slides_url is removed, then this code should also be modified.


progmatic-99:
@mariobehling  Please check this workaround solution. If you have any old session, please check this on those sessions.

app/templates/components/forms/session-section.hbs
```diff
@@ -43,7 +43,8 @@
           @hint={{t "Select an image"}}
           @maxSizeInKb={{this.settings.logoSize}} />
       {{/if}}
-      {{#if (and (eq field.type 'file') (eq field.fieldIdentifier 'slides'))}}
+      {{#if (eq field.type 'file')}}
+        {{#if (or (eq field.fieldIdentifier 'slides') (eq field.fieldIdentifier 'slidesUrl'))}}
```
maze-runnar, line 47: don't add extra if, use it in me statement

progmatic-99: me??

maze-runnar: Sorry, means use in one line.

maze-runnar (approved):

PR #7776, opened 2021-09-11
fix: group thumbnail & header image resized

### LINKED ISSUE #7761: Groups: Thumbnail images and groups page header images are too big
The thumbnail images and groups page header images are huge. They are not downscaled like other images, e.g. on the front page. Images here are in the original size that people upload them, e.g. if anyone uploads a 6MB image that image will appear. In other areas of the system we use the thumbnail version of images. This should be used here as well and the top image should also be a downscaled image.

Compare: https://eventyay.com/my-groups/list and https://eventyay.com/g/40

Screenshot from 2021-09-06 11-53-37

progmatic-99:
Fixes #7761 
ss_1631353291

ss_1631353270

### DIFF
--- a/app/styles/vars.scss
+++ b/app/styles/vars.scss
@@ -1,5 +1,5 @@
 $page-top-margin: 14px;
-$event-lead-image-height: 400px;
+$event-lead-image-height: 350px;
 $event-lead-image-height-minimum: 300px;
 
 $ember-power-select-border-color: rgba(34, 36, 38, .15) !default;
--- a/app/styles/vars.scss
+++ b/app/styles/vars.scss
@@ -1,5 +1,5 @@
 $page-top-margin: 14px;
-$event-lead-image-height: 500px;
+$event-lead-image-height: 400px;
 $event-lead-image-height-minimum: 300px;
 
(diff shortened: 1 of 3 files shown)


app/styles/vars.scss
```diff
@@ -1,5 +1,5 @@
 $page-top-margin: 14px;
-$event-lead-image-height: 500px;
+$event-lead-image-height: 400px;
```
maze-runnar, line 2: this change will impact other areas too. please add the screenshot of those areas where it is used.

mariobehling:
There is no change. The image size is still big e.g. here https://open-event-frontend-git-fork-progmatic-99-fix-7761-eventyay.vercel.app/explore/groups and here https://open-event-frontend-git-fork-progmatic-99-fix-7761-eventyay.vercel.app/my-groups/list.

You changed the style of the image area instead of choosing thumbnail size for the image itself. These changes are not relevant to solve the issue.

progmatic-99:
ss_1632558163
ss_1632558200
ss_1632558233

mariobehling:
I think you did not understand the issue. You keep changing the width and height of the UI, whereas we need a small thumbail image as in other places of the system. Please bring this up in the next meeting.

PR #7938, opened 2021-10-14
fix: add search box to exhibitors page

progmatic-99:
Fixes #7935 
ss_1634206309

### DIFF
--- a/app/components/forms/wizard/basic-details-step.js
+++ b/app/components/forms/wizard/basic-details-step.js
@@ -522,10 +522,12 @@ export default Component.extend(FormMixin, EventWizardMixin, {
     },
 
     async openTaxModal() {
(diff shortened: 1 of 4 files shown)


app/routes/public/exhibition/index.js
```diff
@@ -20,6 +30,19 @@ export default class ExhibitionRoute extends Route {
         val  : 'accepted'
       }
     ];
+
+    if (params.search) {
+      filterOptions.push({
+        or: [
```
maze-runnar, line 36: no need to add or there is only one option for search (by name)

progmatic-99: Name would need to be added to the url search query. I tried without the or, it doesn't work.

PR #8092, opened 2021-11-16
Prevent yarn install on node 16 because it's not supported yet

squatica:
This should prevent the repeated questions of users who try installing the project on node 16. Currently only node <=14 is supported.


squatica:
That makes no sense. PR from my fork's `development` branch is no different than any other branch.

Take it, or leave it. I'm not redoing this.

Issue #8637, opened 2022-10-11
Time format mixed up


mariobehling:
The time format is sometimes following EU continental format mixed up with other formats in the same line, e.g. 16.30 - 4.50.

![Screenshot from 2022-10-11 13-23-35](https://user-images.githubusercontent.com/1583873/195077783-4512f5a6-e710-4415-a68e-8fdc2650d9f4.png)

subhradwip01:
I would like to work on that problem

Santoshkumbhkar1:
i want to work on this problem

harshitraisahab:
@mariobehling it would easily be fixed by JavaScript please assign to me

aniket2468:
I can fix this, please assign me this issue

maze-runnar:
@SARTHAK1SINGH please pick this. 
cc: @mariobehling

PR #8662, opened 2023-01-26
Time format mixed up fixed

SARTHAK1SINGH:
Fixes # https://github.com/fossasia/open-event-frontend/issues/8637

## Before:
Screenshot from 2023-01-26 13-22-42

## After:
Screenshot from 2023-01-26 13-05-21

-
-
-

#### Checklist


maze-runnar (approved):

maze-runnar:
LGTM!

SARTHAK1SINGH:
@mariobehling Tests are failing due to some other reason.
Tests failed before this PR also.
PR cannot be merged due to this.
Please create a new issue for this.
I will take a lot of time and effort for this.

maze-runnar:
tests are failed for master as well currently. 
Screenshot 2023-01-26 at 7 23 29 PM

mariobehling (approved):

mariobehling (approved):

rolandturner:
I have presumably missed a motivating discussion somewhere, but would suggest that there remains a problem:

- It's good that we no longer show sessions as finishing in the early hours of the morning more than 12 hours after they start!
- It's unfortunate that we're using two different time formats in the same display. At any one time we should either use either 12-hour or 24-hour, not mix the two.

We already have a means to select timezone. I'd suggest the same again immediately beside it to select 12/24 hour time. If there is information available in the browser to default to the user's current preference then of course it should default accordingly.

marcoag:
This PR lacks explanation on the problem and why this change fixes the issue. The template of the issue is ignored. Let's try to document PR/Issues properly.

I agree with @rolandturner in that we should not mix time formats in the same display but I would say not even across eventyay. We should further discuss this over issues and other discussion means 

Since this change is minimal and it seems to solve an issue I'll give the LGTM, however I don't think we should close the issue until the standarization of time formats is completed.

marcoag (approved):

cweitat (approved):

PR #8921, opened 2023-07-04
feat #8887: fixing question order on the attendee overview

### LINKED ISSUE #8887: Sort form fields on the overview page in the same way as on the questionnaire
When making orders the overview page shows all questions that have been filled in on the previous page. These questions are in a different order than the attendee form questions. 

Please ensure the overview pages shows the questions in the same order as the attendee form.

Steps:
1. Get a ticket
2. Fill in the form
3. See the overview page

Bekka592:
Fixes #8887 

The overview pages now shows the custom questions in the same order as the attendee form generator and the attendee form. This also works when the user edited the order of questions.

Here an example with some questions and the question "Github" pulled up to the very top, as well as the question "Work Phone" ordered just below "Phone" (else, all questions are ordered alphabetically - as per default):

Screenshot_20230704_161213

Screenshot_20230704_161234

Screenshot_20230704_161305

Screenshot_20230704_161318

Screenshot_20230704_161350

I also adjusted the order of questions in ```custom-form.js```, just so they are uniform with the question order in the server code. This might otherwise lead to weird bugs later on (however, right now this change makes no difference).

#### Checklist


mariobehling (approved):

mariobehling (dismissed):

cweitat:
@mariobehling first name and last name are separated. should we proceed with this for now?

app/components/forms/orders/attendee-list.js
```diff
@@ -74,8 +74,9 @@ export default class AttendeeList extends Component {
         field.transName = field.name;
       }
 
-      return field.form;
-    });
+      return true;
```
cweitat, line 77: i think this line of code is redundant. please check thanks!

cweitat (changes requested):

app/components/forms/orders/attendee-list.js
norbusan: But something needs to be returned. I am more concerned about why `true`? And how the sorting works.

Bekka592: The sorting process works in analogue to the sorting in the `allFields` method of `order-form.js`. That's also where I got that filtering from, even though the `return true` was pretty useless here. However, as Norbert mentioned, something needs to be retuned. I therefore now switched that `filer` to a `map` function. This returns the `field` objects that got modified in the function, in order to then have them accessible for the sorting. What do you think of that solution?

cweitat: This is GTG

Issue #8919, opened 2023-07-04
Access Code Ticket disappears when user clicks number from dropdown


mariobehling:
When an attendee tries to get a hidden ticket using a access code the ticket appears, but when the user clicks the number of a ticket, then the ticket itself disappears.

Test on: https://eventyay.com/e/c93ad54c?code=devtest
Click the ticket number to see the error.

Business Ticket disappears when user clicks ticket number.
![Screenshot from 2023-07-04 10-16-03](https://github.com/fossasia/open-event-frontend/assets/1583873/6a305eaa-062f-444c-b0a5-e4fd83074baa)

Bekka592:
Once the page reloads (that happens whenever just about any button is clicked), the ticket category, unlocked by the access code, hides. However, any selected tickets are added to the shopping cart. You can for example select one access code ticket, then one other ticket and go to checkout with both of them. I assume that the issue is caused by a missing permanent flag that would need to be set within ```applyPromotionalCode()``` in ```ticket-list.js```. However, I was unable to solve it myself.

norbusan:
I debugged it as follows:

I have added `console.log` messages to `this.tickets` (which is a `computed function`, whatever that might be).

When updating one ticket number (irrelevant which one), the function `updateOrder` is called. At the end of this function I log `this.tickets` to console and get three entries, correct. Interestingly, at that point NO logging from the `this.ticket` function is shown.

Then, after a short time, logging from `this.ticket` function appears. In this function, the tickets returned are based on `this.data` and in `this.data` there are only two tickets.

My guess (but I have no idea about JS and ember and whatever), that the `this.ticket` call that is logged comes from when the template is rerendered.

The bug seems to be in how `this.tickets` uses `this.data`

norbusan:
I think this is fixed by the above PR, see explanation in the PR.

PR #8923, opened 2023-07-05
Fix incorrect treatment of promotional codes (Fixes: #8919)

norbusan:
If a promotional code is passed in, the code checks both access code and discount code, and if one of those is not valid, lists the code as invalid. Since access codes are checked first, the later check for discount code will disable the promotional code.

Use two separate booleans to check.


norbusan:
That PR is strange and misses one commit

PR #8924, opened 2023-07-05
Fix access code ticket disappearing 

### LINKED ISSUE #8919: Access Code Ticket disappears when user clicks number from dropdown

norbusan:
Fixes #8919 

When a ticket is shown via access code, it disappears when a ticket number is selected

Fix for the original issue: `this.tickets` is a computed property that is initialized from `this.data` which
only contains the tickets that are shown by default. During the toggling of promo code, the access
code related ticket is added to `this.tickets` via `.addObject`, but when the cache is invalidated by
changing the property `orderAmount`, `this.tickets` is recomputed and falls back to the initial set
of tickets without the access code ticket.

Solution is to always add `this.accessCodeTickets` which contains the additional tickets.

Another fix proposed is to make the logic about promotional code handling correct.
As of now, if a promotional code is passed in, the code checks both access code and
discount code, and if one of those is not valid, lists the code as invalid.
Since access codes are checked first, the later check for discount code will disable the promotional code.

Use two separate booleans to check for these two cases.

#### Checklist

### DIFF
--- a/app/components/public/ticket-list.js
+++ b/app/components/public/ticket-list.js
@@ -45,7 +45,8 @@ export default Component.extend(FormMixin, {
   accessCodeTickets : A(),
   discountedTickets : A(),
 
-  invalidPromotionalCode: false,
+  invalidPromotionalAccessCode: false,
+  invalidPromotionalDiscountCode: false,
 
   tickets: computed('orderAmount', function() {
     const ticketMap = {};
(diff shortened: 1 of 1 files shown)


mariobehling (approved):
Wow, you even thought of a potential error scenario that I had not considered. Thank you

